PR #22, opened 2017-10-02
[FIX] report_id required on new KPI

### LINKED ISSUE #18: need to save report template before adding the first KPI
When creating a new report, it is necessary to save it before adding the first KPI, otherwise an error occurs when attempting to save the new KPI, because the report_id field is not initialized.

richard-willdooit:
When adding a new KPI to an unsaved report, the report_id is required and fails on the popup.

Fixes #18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,9 @@ coverage.xml
 .project
 .pydevproject
 
+# Eclipse
+.settings
+
 # Rope
 .ropeproject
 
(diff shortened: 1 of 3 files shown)


adrienpeiffer (approved):
:+1:

pedrobaeza:
Please change module version number

sbidoul:
Merging on 10.0 as it is a bugfix.

sbidoul:
Merged at https://github.com/OCA/mis-builder/commit/3887de819dccd6955c1406924b4d4ae790d7f70c

PR #23, opened 2017-10-02
Multi company support

### LINKED ISSUE #7: multi-company consolidation
_From @Martronic-SA on April 6, 2017 8:20_

As said in the Readme, we might add a list of companies (which might have their own currencies) and convert amounts for each company/currency to user company currency. That's what Odoo does in account_reports.
As we probably will have to implement this in the mis builder, could you give us advices or ideas to start the project?
Thanks!

_Copied from original issue: OCA/account-financial-reporting#295_

sbidoul:
Moved from https://github.com/acsone/account-financial-reporting/pull/16

fixes #7


mis_builder/tests/test_multi_company_aep.py
```diff
@@ -0,0 +1,275 @@
...
+        self.account_model = self.env['account.account']
+        self.move_model = self.env['account.move']
+        self.journal_model = self.env['account.journal']
+        self.currency_model = self.env['res.currency']
+        self.curr_year = datetime.date.today().year
+        self.prev_year = self.curr_year - 1
+        # create company A and B
+        self.companyA = self.res_company.create({
+            'name': 'AEP Company A',
+            'currency_id': self.currency_model.search([('code', '=', 'USD')])})
```
Martronic-SA: use "name" instead of "code" for searching

mis_builder/tests/test_multi_company_aep.py
```diff
@@ -0,0 +1,275 @@
...
+        self.companyB = self.res_company.create({
+            'name': 'AEP Company B',
+            'currency_id': self.currency_model.search([('code', '=', 'EUR')])})
```
Martronic-SA: same here

sbidoul:
Now working on the testsuite.

mis_builder/tests/test_multi_company_aep.py
```diff
@@ -0,0 +1,275 @@
...
+        self.account_model = self.env['account.account']
+        self.move_model = self.env['account.move']
+        self.journal_model = self.env['account.journal']
+        self.currency_model = self.env['res.currency']
+        self.curr_year = datetime.date.today().year
+        self.prev_year = self.curr_year - 1
+        # create company A and B
+        self.companyA = self.res_company.create({
+            'name': 'AEP Company A',
+            'currency_id': self.currency_model.search([('name', '=', 'USD')])})
```
karan-dreambits: Search on currency_model would return the model and not just id so changing this here and below would make travis go green

karan-dreambits (changes requested):

mis_builder/tests/test_multi_company_aep.py
sbidoul: @karan-dreambits a lot more than that is needed to make travis go green.

I started reworking that test a bit to make it shorter and focus on multi company stuff (we don't need to retest everything that is already covered in test_aep.py)

sbidoul:
Ok, multi-company should work for admin users (we still need to sort out record rules).

@Martronic-SA after some conversations here at the code sprint, we decided to use the date of the end of each period for the currency rate date. That should be a reasonable default for a first version, what do you think?

lepistone (approved):

Martronic-SA:
I think that the most used is todays'rate (most recent rate).
((That's the rate we are using in consolidation for the Balance Sheet.))
edited -> No we are using End of period for monthly detailed Balance Sheet
For the P&L Consolidated we use an average rate from the beginning of the year to the end of each period. (not implemented in my PR [#22](https://github.com/acsone/account-financial-reporting/pull/22)

Ok but by default for this PR it's ok with end date of period and if you make the current year BS or P&L it will take today's rate, that's fine.

sbidoul:
Ok, I merge this in `10.0-dev` that will ultimately become version 3.1.

PR #32, opened 2017-10-12
Fix issue with float_is_zero

### LINKED ISSUE #30: aep.py float_is_zero rounding
Are you sure the following:
```
float_is_zero(debit-credit,
precision_rounding=self.dp)
```
is right?
Correct me if I'm wrong but it seems that self.dp which is currency decimal_places is the number of decimals to take, this is an integer like 2 decimals (1.01) and you are passing this in the precision_rounding parameter of float_is_zero.
Shouldn't you use the precision_digits parameters instead?
```
float_is_zero(debit-credit,
precision_digits=self.dp)
```

sbidoul:
closes #30

### DIFF
--- a/mis_builder/CHANGES.rst
+++ b/mis_builder/CHANGES.rst
@@ -7,13 +7,21 @@ Changelog for mis_builder
 .. *
 
   
+10.0.3.0.x (unreleased)
+~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


Martronic-SA (approved):

gurneyalex (approved):

adrienpeiffer (approved):
:+1:

PR #36, opened 2017-10-13
mis_builder_demo

sbidoul:


gurneyalex (approved):

benwillig (approved):
Tested on runbot. I just had to set company_ids field on the report instance because it's required in this version, but it has already been changed in an other PR. So :+1:

PR #38, opened 2017-10-14
[ADD] extended account selectors

### LINKED ISSUE #4: extended account selectors
_From @sbidoul on September 3, 2017 13:9_

Accounting expressions in mis_builder have the form

`(bal|crd|deb)(p|i|e)[account selector]([optional move line domain])?`

The account selector is currently based on account codes (eg `70%`).

Further account filtering can be added in the move line domain, eg `('account_id.type', '=', ...)`. But that is slow because additional database queries must be done, whereas the account selector boils down to one `read_group` per period and move line domain.

So it is desirable to avoid using the move line domains to filter on accounts.

To implement this, the account selector can be extended to support any domain on gl accounts.

_Copied from original issue: OCA/account-financial-reporting#345_

sbidoul:
closes #4

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -19,6 +19,17 @@ from odoo.tools.float_utils import float_is_zero
 from .accounting_none import AccountingNone
 
 
+def _is_domain(s):
+    """ Test if a string looks like an Odoo domain """
+    return s.startswith('(') or \
+        s.startswith("'|'") or \
+        s.startswith("'&'") or \
+        s.startswith("'!'") or \
+        s.startswith('"|"') or \
+        s.startswith('"&"') or \
+        s.startswith('"!"')
+
+
 class AccountingExpressionProcessor(object):
     """ Processor for accounting expressions.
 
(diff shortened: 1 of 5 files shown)


sbidoul:
@jbeficent would you like to test this one?

mis_builder/models/aep.py
```diff
@@ -84,101 +90,105 @@ def __init__(self, companies, currency=None):
...
+            # legacy bal_NNN%
+            acc_domain = self._account_codes_to_domain(account_sel[1:])
+        else:
+            assert account_sel[0] == '[' and account_sel[-1] == ']'
+            inner_account_sel = account_sel[1:-1].strip()
+            if not inner_account_sel:
+                # empty selector: select all accounts
+                acc_domain = tuple()
+            elif inner_account_sel.startswith('('):
+                # account selector is a domain
```
richard-willdooit: The domain parsing is assuming the first element of the domain is a tuple - it could be a '!', '|', or '&'

richard-willdooit (changes requested):
Only really the one suggested enhancement at the account domain...

mis_builder/models/aep.py
sbidoul: Right! Thanks for the in depth review!

sbidoul: Fixed at 0314b8451a0b12228699f0ed4f5b7f329bcec5d4

richard-willdooit (approved):
Cool.

sbidoul:
Small performance improvement to last commit and improved test.

adrienpeiffer (approved):
:+1:

PR #35, opened 2017-10-13
[ADD] multi-company record rule for MIS Budgets

sbidoul:
Fixes #27

Issue #34, opened 2017-10-13
Multi-company budget consolidation


sbidoul:
mis_builder can consolidate actuals from different companies. One can imagine to consolidate budgets from different companies too. To be elaborated.

sbidoul:
@AnizR @sannabelle I think we consider this has been implemented for budget by accounts in https://github.com/OCA/mis-builder/issues/620 and https://github.com/OCA/mis-builder/pull/630

PR #40, opened 2017-10-15
[FIX] a few multi company fixes

sbidoul:
* alt move lines view must have company_id field
* make company_ids optional to facilitate migration and be more
  robust in case the onchange multi_company has not be triggered
  (use company_id if company_ids is empty)

Refs #7

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -141,7 +141,7 @@ class MisReportInstancePeriod(models.Model):
                      '|',
                      ('company_id', '=', False),
                      ('company_id', 'in',
-                      record.report_instance_id.company_ids.ids)])
+                      record.report_instance_id._query_companies().ids)])
                 if current_periods:
                     # TODO we take the first date range we found as current
                     #      this may be surprising if several companies
@@ -235,9 +235,10 @@ class MisReportInstancePeriod(models.Model):
         domain=[('field_id.name', '=', 'debit'),
                 ('field_id.name', '=', 'credit'),
                 ('field_id.name', '=', 'account_id'),
-                ('field_id.name', '=', 'date')],
+                ('field_id.name', '=', 'date'),
+                ('field_id.name', '=', 'company_id')],
         help="A 'move line like' model, ie having at least debit, credit, "
-             "date and account_id fields.",
+             "date, account_id and company_id fields.",
     )
     source_sumcol_ids = fields.One2many(
         comodel_name='mis.report.instance.period.sum',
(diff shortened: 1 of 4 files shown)


mis_builder/models/mis_report_instance.py
```diff
@@ -433,7 +433,14 @@ def _onchange_company(self):
...
         else:
-            self.company_ids = self.company_id
+            self.company_ids = False
+
+    def _get_query_companies(self):
+        """ Return companies to query """
+        if self.multi_company:
+            return self.company_ids or self.company_id
+        else:
+            return self.company_id
```
richard-willdooit, line 446: Definitely cleaner.

mis_builder/models/mis_report_instance.py
```diff
@@ -235,9 +235,10 @@ def _compute_dates(self):
         domain=[('field_id.name', '=', 'debit'),
                 ('field_id.name', '=', 'credit'),
                 ('field_id.name', '=', 'account_id'),
-                ('field_id.name', '=', 'date')],
+                ('field_id.name', '=', 'date'),
+                ('field_id.name', '=', 'company_id')],
```
richard-willdooit, line 239: Hopefully, this wont be a restriction - I think if a model does not have a company_id, then it needs to be re-visited in light of that.  Base Odoo generally gets it right...

Which of course, highlights, budgets will also need to be company aware...

richard-willdooit (approved):
Looks good, and works with or without #38

mis_builder/models/mis_report_instance.py
sbidoul: > if a model does not have a company_id, then it needs to be re-visited 

Indeed. I decided to document that as a breaking change, because making it optional would complicate the code for little benefit. It's easy to adapt such views, and there must not be many in the wild as 3.0 is still very young.

> budgets will also need to be company aware

At this point I just added company_id on budget for the usual record rule (#35). And I created #34 (multi-company budget consolidation) for the future. It's still useful without it.

mis_builder/models/mis_report_instance.py
```diff
@@ -433,7 +433,14 @@ def _onchange_company(self):
                 ('id', 'child_of', self.company_id.id),
             ])
         else:
-            self.company_ids = self.company_id
+            self.company_ids = False
+
+    def _get_query_companies(self):
+        """ Return companies to query """
```
benwillig: I think there should be a ```api.multi``` with a ```self.ensure_one```.

sbidoul: Why? It's just a helper method.

benwillig: Yes, but it must be called on a single recordset or it will crash. It will be the same behavior using the decorator, but IMO it's more explicit and we don't have to read all the method to know if it's a "static" or an object method.

sbidoul: Ok, done.

benwillig (approved):

PR #50, opened 2017-11-08
[FIX] Allow to create a period in a mis report instance before saving (create) this one0

adrienpeiffer:


sbidoul (approved):
Thanks!

sbidoul:
Merging trivial change

PR #51, opened 2017-11-08
[IMP] Improve base date position on mis builder instance form.

adrienpeiffer:
Before
capture du 2017-11-08 12-30-47
After
capture du 2017-11-08 18-58-27

Sorry for the 2 different versions of Odoo

### DIFF
--- a/mis_builder/views/mis_report_instance.xml
+++ b/mis_builder/views/mis_report_instance.xml
@@ -81,7 +81,12 @@
                     <notebook>
                         <page string="Columns" attrs="{'invisible': [('comparison_mode', '=', False)]}">
                             <group>
-                                <field name="period_ids" colspan="4" nolabel="1" attrs="{'required': [('comparison_mode', '=', True)]}" context="{'default_report_instance_id': id}">
+                                <group>
+                                    <field name="date" />
+                                </group>
+                                <group>
+                                </group>
+                                <field name="period_ids" nolabel="1" colspan="4" attrs="{'required': [('comparison_mode', '=', True)]}" context="{'default_report_instance_id': id}">
                                     <tree colors="red:valid==False">
                                         <field name="sequence" widget="handle"/>
                                         <field name="name"/>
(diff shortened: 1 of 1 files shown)


mis_builder/views/mis_report_instance.xml
```diff
@@ -81,18 +81,22 @@
...
-                                        <field name="date_to"/>
-                                        <field name="valid" invisible="1"/>
-                                    </tree>
-                                </field>
-                                <field name="date"/>
+                                <group>
+                                    <field name="date" />
+                                </group>
+                                <group>
+                                </group>
```
sbidoul, line 88: why the empty group?

mis_builder/views/mis_report_instance.xml
```diff
@@ -81,18 +81,22 @@
...
+                            <field name="period_ids" colspan="4" attrs="{'required': [('comparison_mode', '=', True)]}" context="{'default_report_instance_id': id}">
+                                <tree colors="red:valid==False">
+                                    <field name="sequence" widget="handle"/>
+                                    <field name="name"/>
+                                    <field name="source"/>
+                                    <field name="date_from"/>
+                                    <field name="date_to"/>
+                                    <field name="valid" invisible="1"/>
+                                </tree>
+                            </field>
```
sbidoul: Move it inside the above group to remove the empty space between the date and the list?

mis_builder/views/mis_report_instance.xml
adrienpeiffer: This is the workaround that I found to display the select box only on half on the sheet. Before that, I tried a lots of combination with col and colspan attribute without any success ... If you have any idea ...  (nobody in luxembourg has one :cry: ).

mis_builder/views/mis_report_instance.xml
adrienpeiffer: I'll try

adrienpeiffer:
@sbidoul I just added the tree inside the group and updated the screenshot in the description.

mis_builder/views/mis_report_instance.xml
adrienpeiffer: @sbidoul Done

sbidoul (approved):

sbidoul (approved):

PR #52, opened 2017-11-10
[9.0][FIX] Fill some defaults in migration script

MiquelRForgeFlow:
From v8 to v9, some fields has become required (see commits https://github.com/OCA/mis-builder/commit/7e20ff90646fbee20707aeabf294cba77bb239f0 and https://github.com/OCA/mis-builder/commit/db7e6eece53e4a15274582d7f5c83a74c16fc05a). Thus, in case some of them were not filled, they should be filled in the migration.

### DIFF
--- a/mis_builder/migrations/8.0.2.0.0/post-migration.py
+++ b/mis_builder/migrations/8.0.2.0.0/post-migration.py
@@ -2,6 +2,27 @@
 # © 2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from openerp import api, SUPERUSER_ID
+
+
+def fill_defaults(env):
+    cr = env.cr
+
+    default_specs = {
+        'mis.report.instance.period': ['report_instance_id'],
+        'mis.report.kpi': ['report_id'],
+        'mis.report.query': ['report_id'],
+        'mis.report.instance': ['company_id'],
+    }
+
+    for model in default_specs.keys():
+        obj = env[model]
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
LGTM

mis_builder/migrations/8.0.2.0.0/post-migration.py
```diff
@@ -2,6 +2,27 @@
...
+from openerp import api, SUPERUSER_ID
+
+
+def fill_defaults(env):
+    cr = env.cr
+
+    default_specs = {
+        'mis.report.instance.period': ['report_instance_id'],
+        'mis.report.kpi': ['report_id'],
+        'mis.report.query': ['report_id'],
```
sbidoul, line 14: `mis.report.instance.period` without report instance make no sense. Same for kpi and queries without reports. If you have those in your 8.0 database, you should simply drop them. They are the result of bugs (missing unlink or cascade delete).

mis_builder/migrations/8.0.2.0.0/post-migration.py
```diff
@@ -2,6 +2,27 @@
...
+        'mis.report.instance': ['company_id'],
```
sbidoul, line 15: Setting a default for company_id is going to be sensitive to the current default company of the admin user running the migration. In a multi-company context this is delicate. So I'm not sure we should set a default in the migration.

sbidoul:
@mreficent can we close this?

MiquelRForgeFlow:
As you prefer @sbidoul 😉

PR #54, opened 2017-11-14
[9.0] backport v3.1

sbidoul:


sbidoul:
@vdewulf @eficent  you may want to test this.

vdewulf:
Hello, 

I installed this update locally and tested with the templates (VAT) and everything seems OK...
I don't see exactly what use case I could test, don't hesitate to share ideas with me.

One idea though: add a dependency of MIS Builder on account_accountant. Otherwise, there is only the configuration of MIS Builder available, not the reports themselves.

sbidoul:
Merging as it is a straightforward backport, and has been tested independently.

sbidoul:
@vdewulf I created https://github.com/OCA/mis-builder/issues/55 to track your improvement idea.

PR #56, opened 2017-11-17
[10.0] mis_builder python3 compatibility

JordiBForgeFlow:
Adds compatibility with python 3. Now requires python package `future`

### DIFF
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -2,12 +2,15 @@
 # Copyright 2016 Therp BV (<http://therp.nl>)
 # Copyright 2016-2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-
+import sys
 from odoo import api, fields, models, _
 from odoo.exceptions import ValidationError
 
 from .accounting_none import AccountingNone
 from .data_error import DataError
+
+if sys.version_info.major >= 3:
+    unicode = str
 
 
 class PropertyDict(dict):
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -5,7 +5,11 @@
 import datetime
 import re
 from collections import defaultdict
-from itertools import izip
+try:
+    import itertools.izip as zip
+except ImportError:
+    pass
+
 import time
 
 import dateutil
@@ -519,4 +523,4 @@ class AccountingExpressionProcessor(object):
         # or leave that to the caller?
         bals = cls._get_balances(cls.MODE_UNALLOCATED, companies,
                                  date, date, target_move)
-        return tuple(map(sum, izip(*bals.values())))
+        return tuple(map(sum, zip(*bals.values())))
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -3,7 +3,11 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from collections import defaultdict, OrderedDict
-from itertools import izip
+try:
+    import itertools.izip as zip
+except ImportError:
+    pass
+
 import logging
 
 from odoo import _
@@ -233,7 +237,7 @@ class KpiMatrix(object):
         assert len(vals) == col.colspan
         assert len(drilldown_args) == col.colspan
         for val, drilldown_arg, subcol in \
-                izip(vals, drilldown_args, col.iter_subcols()):
+                zip(vals, drilldown_args, col.iter_subcols()):
             if isinstance(val, DataError):
                 val_rendered = val.name
                 val_comment = val.msg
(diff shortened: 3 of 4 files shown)


mis_builder/models/accounting_none.py
```diff
@@ -89,6 +89,7 @@
 True
 """
 
+from builtins import object
```
sbidoul: We don't need this.

In v11, simply remove the `(object)` in class declarations, that will not hinder future merges from 10 to 11 because it's very localized.

mis_builder/models/simple_array.py
```diff
@@ -130,10 +130,10 @@ def __add__(self, other):
     __radd__ = __add__
 
     def __pos__(self):
-        return self.__class__(map(operator.pos, self))
+        return self.__class__(list(map(operator.pos, self)))
 
     def __neg__(self):
-        return self.__class__(map(operator.neg, self))
+        return self.__class__(list(map(operator.neg, self)))
```
sbidoul: I think all this places where `list()` has been inserted are not necessary.

mis_builder/models/mis_report_style.py
```diff
@@ -188,7 +189,7 @@ def render_pct(self, lang, value, dp=1, sign='-'):
     def render_str(self, lang, value):
         if value is None or value is AccountingNone:
             return u''
-        return unicode(value)
+        return str(value)
```
sbidoul: It looks like this would be the only place where we'd need the future module. In which case we can find a way to do without it.

mis_builder/models/accounting_none.py
```diff
@@ -167,9 +168,6 @@ def __repr__(self):
     def __str__(self):
         return ''
 
-    def __nonzero__(self):
-        return False
-
```
sbidoul: If I read the [doc](https://docs.python.org/2/reference/datamodel.html#object.__nonzero__) correctly we need this for python 2. What is weird is that the test of this PR seems to pass, and we have a test for that...

mis_builder/models/accounting_none.py
JordiBForgeFlow: OK. I understand that it will be needed in v10. Otherwise it fails 

2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: ` File "/home/jordi/odoo-addons/10.0/OCA/mis-builder/mis_builder/models/accounting_none.py", line 66, in odoo.addons.mis_builder.models.accounting_none
2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: ` Failed example:
2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: `     bool(AccountingNone)
2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: ` Expected:
2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: `     False
2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: ` Got:
2017-11-17 16:02:01,769 7451 ERROR 10-mis_reports odoo.addons.mis_builder.tests.test_accounting_none: `     True

mis_builder/models/accounting_none.py
JordiBForgeFlow: The part of from builtins import object fixed precisely that.

mis_builder/models/mis_report_style.py
sbidoul: I'd leave `unicode()` here, and at the top of the file, do

```python
if sys.version_info.major >= 3:
    unicode = str
```

JordiBForgeFlow:
@sbidoul I have now removed the list() calls in 10.0 and 11.0-dev PR's.

mis_builder/models/mis_report_style.py
sbidoul: And remove the dependency on `future`

mis_builder/models/accounting_none.py
sbidoul: Ah ok. I'd like to do without the dependency on `future` if we can. And this nonzero method does no harm in python3.

.travis.yml
```diff
@@ -38,6 +38,7 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - export WKHTMLTOPDF_VERSION=0.12.4
   - travis_install_nightly
+  - pip install future
```
sbidoul: remove this, then if green we can merge

JordiBForgeFlow:
@sbidoul Done!

sbidoul:
Merged manually at edc61e986664f74d50dfaa71ed23e17fadddeb19. Thanks!

I also moved the 11.0-dev to 10.0.

JordiBForgeFlow:
Merged 11.0-dev to 11.0 you mean?

sbidoul:
There is no 11.0 yet. At this point the tip of 11.0-dev is at 10.0 (which is edc61e9).

You can rebase #58 on 11.0-dev.

PR #57, opened 2017-11-17
[11.0][mig] mis_builder

JordiBForgeFlow:
Migration to 11.0. Forward port of https://github.com/OCA/mis-builder/pull/56

Commit eeebff9 corresponds to the actual migration to v11.

Some issues that we have found so far:
[ ] Cannot render the widget
[ ] Tests on `TestKpiData` fail - seems related to the creation of `mis.kpi.data.test.item`


sbidoul:
Thanks! Can you please rebase and do the PR to branch 11.0-dev? 

We do it a bit differently here because it's a cohesive repo and I use git merge to maintain feature parity across Odoo series.

PR #58, opened 2017-11-17
[11.0-dev][MIG][mis_builder]

JordiBForgeFlow:
Migration to 11.0. Forward port of #56

Commit eeebff9 corresponds to the actual migration to v11.

Some issues that we have found so far:
[x] Cannot render the widget => DONE
[x] Tests on TestKpiData fail - seems related to the creation of mis.kpi.data.test.item
[x] Refresh button not working => DONE
[x] Add to dashboard part is missing => DONE


JordiBForgeFlow:
@sbidoul This is the good one ;)

mis_builder/models/accounting_none.py
```diff
@@ -167,9 +168,6 @@ def __repr__(self):
     def __str__(self):
         return ''
 
-    def __nonzero__(self):
-        return False
-
```
sbidoul: If I read the [doc](https://docs.python.org/2/reference/datamodel.html#object.__nonzero__) correctly we need this for python 2. What is weird is that the test of this PR seems to pass, and we have a test for that...

JordiBForgeFlow:
I found some other errors:
* When you create a MIS Report template and add a line, and then save, you get the following error:

```Traceback (most recent call last):
  File "/home/jordi/odoo/11.0/odoo/odoo/fields.py", line 944, in __get__
    value = record.env.cache.get(record, self)
  File "/home/jordi/odoo/11.0/odoo/odoo/api.py", line 960, in get
    value = self._data[field][record.id][key]
KeyError: (<odoo.sql_db.Cursor object at 0x7ff035404b70>, 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 646, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 307, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/home/jordi/odoo/11.0/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 683, in dispatch
    result = self._call_function(**self.params)
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 339, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/jordi/odoo/11.0/odoo/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 332, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 927, in __call__
    return self.method(*args, **kw)
  File "/home/jordi/odoo/11.0/odoo/odoo/http.py", line 512, in response_wrap
    response = f(*args, **kw)
  File "/home/jordi/odoo/11.0/odoo/addons/web/controllers/main.py", line 920, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/jordi/odoo/11.0/odoo/addons/web/controllers/main.py", line 912, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/jordi/odoo/11.0/odoo/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/home/jordi/odoo/11.0/odoo/odoo/api.py", line 672, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/home/jordi/odoo/11.0/odoo/odoo/models.py", line 3275, in create
    record = self.browse(self._create(old_vals))
  File "/home/jordi/odoo/11.0/odoo/odoo/models.py", line 3422, in _create
    field.write(self.with_context(rel_context), vals[name], create=True)
  File "/home/jordi/odoo/11.0/odoo/odoo/fields.py", line 2268, in write
    if int(line_sudo[inverse]) != record.id:
  File "/home/jordi/odoo/11.0/odoo/odoo/models.py", line 4654, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/home/jordi/odoo/11.0/odoo/odoo/fields.py", line 948, in __get__
    self.determine_value(record)
  File "/home/jordi/odoo/11.0/odoo/odoo/fields.py", line 1049, in determine_value
    record._prefetch_field(self)
  File "/home/jordi/odoo/11.0/odoo/odoo/models.py", line 2562, in _prefetch_field
    result = records.read([f.name for f in fs], load='_classic_write')
  File "/home/jordi/odoo/11.0/odoo/odoo/models.py", line 2506, in read
    self._read_from_database(stored, inherited)
  File "/home/jordi/odoo/11.0/odoo/odoo/models.py", line 2625, in _read_from_database
    cr.execute(query_str, params)
  File "/home/jordi/odoo/11.0/odoo/odoo/sql_db.py", line 155, in wrapper
    return f(self, *args, **kwargs)
  File "/home/jordi/odoo/11.0/odoo/odoo/sql_db.py", line 232, in execute
    res = self._obj.execute(query, params)
psycopg2.DataError: invalid input syntax for integer: "virtual_53"
LINE 1: ...OM "mis_report_kpi" WHERE "mis_report_kpi".id IN ('virtual_5...
```

sbidoul:
Do you have the same problem when populating the Queries tab?

JordiBForgeFlow:
No, in queries works well.

JordiBForgeFlow:
@sbidoul @lmignon we have migrated the JS part to the new framework. It is now functional, except for the refresh button and add to dashboard.

JordiBForgeFlow:
@sbidoul @lmignon MIS builder now seems to be fully functional in 11.0. Perhaps the way I found to render the widget is not the best, but I had a really hard time finding a way to make it render in the form view and also in the Dashboard.

The Dashboard design seems to have changed with respect to 10.0, and will no longer render a single record in form view. So I had to work around this limitation.

I still have to check if the user context settings are being correctly considered.

lmignon:
@jbeficent Thank you for all this work. I'm busy on others topics and I don't know when I'll have time to look into this problem.

JordiBForgeFlow:
@lmignon no problem! We'll now try to improve the unit tests on the JS side.

mis_builder/models/mis_report.py
```diff
@@ -176,14 +176,17 @@ def _onchange_name(self):
...
+                for expression in kpi.expression_ids:
+                    if expression.subkpi_id:
+                        exprs.append(u'{}\xa0=\xa0{}'.format(
+                            expression.subkpi_id.name, expression.name))
+                    else:
+                        exprs.append(
+                            expression.name or 'AccountingNone')
+                kpi.expression = ',\n'.join(exprs)
+            else:
+                pass
```
sbidoul: Why changing this method?

mis_builder/models/mis_report.py
```diff
@@ -117,7 +117,7 @@ class MisReportKpi(models.Model):
         [(ACC_SUM, _('Sum')),
          (ACC_AVG, _('Average')),
          (ACC_NONE, _('None'))],
-        required=True,
+        required=False,
```
sbidoul: Why changing this, especially in the context of the migration?

mis_builder/models/mis_report.py
```diff
@@ -74,7 +74,7 @@ class MisReportKpi(models.Model):
 
     name = fields.Char(size=32, required=True,
                        string='Name')
-    description = fields.Char(required=True,
+    description = fields.Char(required=False,
```
sbidoul: Why changing this?

mis_builder/models/mis_report.py
JordiBForgeFlow: Sorry, we will revert those silly changes. These are leftovers of our tests to attempt to fix the problem described in the beginning.

mis_builder/models/mis_report.py
HviorForgeFlow: misconfig after fix, now tested on True and working.
Thanks for review

JordiBForgeFlow:
We're really having a hard time migrating the testing part that creates a custom model https://github.com/Eficent/mis-builder/blob/956c2f4ba376692e999565dba11f7dc3c02cd0d0/mis_builder/tests/common.py#L26
Seems that the model is created, but not available when you search it in https://github.com/Eficent/mis-builder/blob/956c2f4ba376692e999565dba11f7dc3c02cd0d0/mis_builder/tests/common.py#L12

I'm kindly asking here to some core experts for some tip on how to move forward with this: @sbidoul @lmignon @hbrunn @StefanRijnhart

hbrunn:
I fail to find out where this function is called, maybe that's the problem? Otherwise: Check the properties `_register` and `_auto` for your test class. Chances are you inherit from `BaseModel`, that defaults those two to `False`. You need them to be `True`

JordiBForgeFlow:
@hbrunn The function is called here https://github.com/Eficent/mis-builder/blob/956c2f4ba376692e999565dba11f7dc3c02cd0d0/mis_builder/tests/test_kpi_data.py#L24

mis_builder/models/mis_report_instance.py
```diff
@@ -527,43 +554,39 @@ def _onchange_dates(self):
...
         }
 
     @api.multi
     def export_xls(self):
         self.ensure_one()
+        context = dict(self.env.context)
+        context1 = {'active_ids': self.ids}
+        if isinstance(context1, pycompat.string_types):
+            context1 = safe_eval(context1, dict(context))
+        context.update(context1)
```
sbidoul: I don't understand this piece. Perhaps `context = dict(self.context, active_ids=self.ids)` ?

sbidoul:
@jbeficent I'll look into this next week.

JordiBForgeFlow:
@hbrunn @sbidoul @lmignon I have finally resolved the way to properly introduce new models during testing. See https://github.com/OCA/mis-builder/pull/58/files#diff-67ee724417786e97c27225738ceb907bR34

Now, IMHO mis_builder starts to be good enough for v11!

I am telling you, this baby has made me sweat blood and tears! ;)

mis_builder_budget/models/mis_report_instance.py
```diff
@@ -63,6 +64,7 @@ def _add_column(self, aep, kpi_matrix, period, label, description):
     @api.multi
     def drilldown(self, arg):
         self.ensure_one()
+        arg = safe_eval(arg)
```
sbidoul: I think this should not be here. Removed in https://github.com/Eficent/mis-builder/pull/1

mis_builder/models/mis_report_instance.py
```diff
@@ -431,6 +433,30 @@ def _default_company_id(self):
...
+        rcontext['mis_report_data'] = mis_report_data
+        rcontext['mis_report_data']['report_name'] = self.name
+        rcontext['mis_instance_id'] = self.id
+        return self.env.ref(
+            'mis_builder.MisReportInstance').render(rcontext)
+
+    @api.multi
+    def get_mis_report_view_from_id(self):
+        for rec in self:
+            return rec.get_mis_report_view_html()
```
sbidoul, line 457: This return inside a for loop is suspicious. If this method is really necessary, there should be an ensure_one and no loop. But then it does the same as get_mis_report_view_html so it's probably useless?

mis_builder/views/mis_report_instance.xml
```diff
@@ -197,19 +197,16 @@
...
                             </field>
                             <field name="source_sumcol_accdet"
                                    attrs="{'invisible': [('source', '!=', 'sumcol')]}"/>
                             <field name="source_cmpcol_to_id"
                                    attrs="{'invisible': [('source', '!=', 'cmpcol')], 'required': [('source', '=', 'cmpcol')]}"
-                                   domain="[('report_instance_id', '=', report_instance_id), ('id', '!=', id)]"
                                    options="{'no_create': True, 'no_open': True}"/>
                             <field name="source_cmpcol_from_id"
                                    attrs="{'invisible': [('source', '!=', 'cmpcol')], 'required': [('source', '=', 'cmpcol')]}"
-                                   domain="[('report_instance_id', '=', report_instance_id), ('id', '!=', id)]"
```
sbidoul, line 212: This domain is important because we can only compare columns of the same report instance. And we cannot compare a column to itself hence 'id' != id. 

Why removing it?

JordiBForgeFlow: It was giving nasty errors. I think that we need to resolve changing the domain dynamically in onchange methods

sbidoul: Hm, I remember now. What we'd ideally need is to refer to `parent.parent` in view domains.

Yet all this did work fine in 9 and 10. Perhaps we should summon rco with an accurate description of the problem. That's not for tonight, though.

Thanks for this huge work btw. Wishing you a happy end of year.

sbidoul:
```
Error:
Odoo Server Error

Traceback (most recent call last):
  File "/home/odoo-11/src/odoo/odoo/fields.py", line 944, in __get__
    value = record.env.cache.get(record, self)
  File "/home/odoo-11/src/odoo/odoo/api.py", line 960, in get
    value = self._data[field][record.id][key]
KeyError: (<odoo.sql_db.Cursor object at 0x7f1f7152b550>, 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/odoo-11/src/odoo/odoo/http.py", line 646, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/odoo-11/src/odoo/odoo/http.py", line 307, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/home/odoo-11/src/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/home/odoo-11/src/odoo/odoo/http.py", line 683, in dispatch
    result = self._call_function(**self.params)
  File "/home/odoo-11/src/odoo/odoo/http.py", line 339, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/odoo-11/src/odoo/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/odoo-11/src/odoo/odoo/http.py", line 332, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/odoo-11/src/odoo/odoo/http.py", line 927, in __call__
    return self.method(*args, **kw)
  File "/home/odoo-11/src/odoo/odoo/http.py", line 512, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo-11/src/odoo/addons/web/controllers/main.py", line 924, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo-11/src/odoo/addons/web/controllers/main.py", line 916, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/odoo-11/src/odoo/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/home/odoo-11/src/odoo/odoo/api.py", line 672, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/home/odoo-11/src/odoo/odoo/models.py", line 3278, in create
    record = self.browse(self._create(old_vals))
  File "/home/odoo-11/src/odoo/odoo/models.py", line 3425, in _create
    field.write(self.with_context(rel_context), vals[name], create=True)
  File "/home/odoo-11/src/odoo/odoo/fields.py", line 2273, in write
    if int(line_sudo[inverse]) != record.id:
  File "/home/odoo-11/src/odoo/odoo/models.py", line 4657, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/home/odoo-11/src/odoo/odoo/fields.py", line 948, in __get__
    self.determine_value(record)
  File "/home/odoo-11/src/odoo/odoo/fields.py", line 1049, in determine_value
    record._prefetch_field(self)
  File "/home/odoo-11/src/odoo/odoo/models.py", line 2565, in _prefetch_field
    result = records.read([f.name for f in fs], load='_classic_write')
  File "/home/odoo-11/src/odoo/odoo/models.py", line 2509, in read
    self._read_from_database(stored, inherited)
  File "/home/odoo-11/src/odoo/odoo/models.py", line 2628, in _read_from_database
    cr.execute(query_str, params)
  File "/home/odoo-11/src/odoo/odoo/sql_db.py", line 155, in wrapper
    return f(self, *args, **kwargs)
  File "/home/odoo-11/src/odoo/odoo/sql_db.py", line 232, in execute
    res = self._obj.execute(query, params)
psycopg2.DataError: invalid input syntax for integer: "virtual_55"
LINE 1: ...OM "mis_report_kpi" WHERE "mis_report_kpi".id IN ('virtual_5...
                                                             ^
```

I'm pretty sure we are hitting an orm bug. If I read the stack trace well it's attempting to prefetch the parent record which is still a virtual_ID.

JordiBForgeFlow:
That"s the nasty error I was talking about that caused so much headache.

sbidoul:
@jbeficent I think I nailed it: https://github.com/odoo/odoo/pull/21944

JordiBForgeFlow:
Ok, will test it in the next days. Have a nice year end!

sbidoul:
@jbeficent I merge this in 11.0-dev so we can collaborate more efficiently by doing addional PR's on to 11.0-dev.

PR #61, opened 2017-12-04
[8.0][FIX] Added a wrpping method for get_pdf() to make it compatible to n…

mathiasfrancke:
Fixed a bug where calling get_pdf() with new API result in errors.

Description:

When the MIS Builder is installed and the get_pdf() method is called it gives an error. It can be seen when debugging the call into the api wrapper that the function pointer new_api() is assigned to "None".

Solution:

Adding a method definition with decorator for "api.v8" does the trick.

### DIFF
--- a/mis_builder/report/report_mis_report_instance.py
+++ b/mis_builder/report/report_mis_report_instance.py
@@ -50,7 +50,7 @@ class Report(models.Model):
                                            context=context)
 
     @api.v8
-    def get_pdf(self, records, report_name, html=None, data=None):
-        return Report.get_pdf(self._model, self._cr, self._uid, records.ids,
-                              report_name, html=html, data=data,
-                              context=self._context)
+    def get_pdf(self, docids, report_name, html=None, data=None):
+        return self._model.get_pdf(self._cr, self._uid,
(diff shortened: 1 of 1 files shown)


sbidoul:
@mathiasfrancke thanks for the fix. I realize it was already fixed on the 9.0 branch is a slightly different way.

Would you mind aligning the code with 

https://github.com/OCA/mis-builder/blob/dde812cc5cadd022eb23393b7517f6c152021265/mis_builder/report/mis_report_instance_qweb.py#L29-L33

so maintainance is a tad easier across series.

sbidoul:
Thanks!

sbidoul (approved):

mathiasfrancke:
@sbidoul Thank you - that's very kind from you. Im sorry for pushing you. This was the first thing I did in the community so I don't yet know how everything is handled yet.

sbidoul:
Fast tracking as this fix has been proven on v9 already.

mathiasfrancke:
I also tested the changes and it works well.

lmignon (commented):
:+1: Code review only

PR #63, opened 2017-12-21
[8.0][FIX] call of api-7 function in api-8 function get_pdf()

kmatthes:
Fixed a bug where calling `get_pdf()` results in a `QWebException` while rendering the template.

Description:
In #61 a wrapper function `get_pdf()` with `api.v8` was added. Unfortunatly not all function parameters where handled well, so a recodset is passed through instead of a list of ids, leading to problems when interpreting the qweb template.

Solution:
Pass ids of the recordset to next function instead of the recordset itself.

### DIFF
--- a/mis_builder/report/report_mis_report_instance.py
+++ b/mis_builder/report/report_mis_report_instance.py
@@ -52,5 +52,5 @@ class Report(models.Model):
     @api.v8
     def get_pdf(self, docids, report_name, html=None, data=None):
         return self._model.get_pdf(self._cr, self._uid,
-                                   docids, report_name,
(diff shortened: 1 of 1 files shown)


sbidoul:
Your fix looks correct. Now I'm wondering how it did go unnoticed on the v9 version.

sbidoul (approved):

kmatthes:
@sbidoul Actually I was wondering myself, why the previous solution was working for 1 of 4 reports in our customer case. I suggest it has something to do with how and if it is iterated over the `doc_ids` in the QWeb templates.

PR #67, opened 2018-01-05
[11.0] migration

sbidoul:
TODO:

- [x] This PR requires https://github.com/odoo/odoo/pull/21944. It's merged in saas-16, wait until it's forward ported to 11.0.
- [x] investigate some changes in mis_builder_template.xml wrt v10
- [x] drilldown is broken because the arguments is sent as string to the backend, probably due to a change in the js framework, the conversion to json must take place [here](https://github.com/OCA/mis-builder/blob/ba767f4909dc62c95333a649e1e7a62da24ab636/mis_builder/static/src/js/mis_builder_widget.js#L111)
- [x] investigate changes in the definition of fields seq1, seq2 in [mis_kpi_data.py](https://github.com/OCA/mis-builder/blob/ba767f4909dc62c95333a649e1e7a62da24ab636/mis_builder/models/mis_kpi_data.py): these should remain related as in v10, or is there an ORM bug?


sbidoul:
Fixed the drilldown test and rebased.

I updated the review TODO list in the PR description.

mis_builder/models/mis_report_instance.py
```diff
@@ -431,6 +430,30 @@ def _default_company_id(self):
...
+        rcontext['mis_report_data'] = mis_report_data
+        rcontext['mis_report_data']['report_name'] = self.name
+        rcontext['mis_instance_id'] = self.id
+        return self.env.ref(
+            'mis_builder.MisReportInstance').render(rcontext)
+
+    @api.multi
+    def get_mis_report_view_from_id(self):
+        for rec in self:
+            return rec.get_mis_report_view_html()
```
sbidoul: This return inside the loop is not correct.

mis_builder/models/mis_report_instance.py
```diff
@@ -431,6 +430,30 @@ def _default_company_id(self):
...
     temporary = fields.Boolean(default=False)
+    mis_report_view = fields.Text(string='MIS Report view',
+                                  compute='_compute_mis_report_view')
+
+    def _compute_mis_report_view(self):
+        """This method is dummy because we use this field as a placeholder
+        to render the actual mis report instance in the html format using the
+        widget"""
+        for rec in self:
+            rec.mis_report_view = ""
```
sbidoul: I'm not comfortable with this dummy field. Surely there must be a better way. Why is it computed btw?

mis_builder/static/src/js/mis_builder_widget.js
```diff
@@ -0,0 +1,121 @@
...
+    drilldown: function(event) {
+        var self = this;
+        var context = self.getParent().state.context;
+        var drilldown = $(event.target).data("drilldown");
+        this._rpc({
+                model: 'mis.report.instance',
+                method: 'drilldown',
+                args: [context.active_id, drilldown],
+                context: context,
+            }).then(function(result){
```
sbidoul: I my understanding of the js framework, there are better, higher level, ways to invoke backend methods than `_rpc`. These higher level methods were used in v10. This is probably the cause of drilldown not working.

mis_builder/views/mis_report_template.xml
```diff
@@ -0,0 +1,72 @@
...
+                                    <t t-esc="row['label']"/>
+                                    <t t-if="row['description']">
+                                      <br/>
+                                      <t t-esc="row['description']"/>
+                                    </t>
+                                </td>
+
+                                    <td t-foreach="row['cells']" t-as="cell"
+                                        t-att="{'style': cell['style'], 'title': cell['val_c']}" class="mis_builder_amount">
+                                        <t t-if="row['cells']">
```
sbidoul: why this `t-if row['cells']` ?

mis_builder/models/kpimatrix.py
```diff
@@ -462,7 +461,13 @@ def as_dict(self):
...
                 if cell is None:
                     # TODO use subcol style here
-                    row_data['cells'].append({})
+                    col_data = {
+                        'val': None,
+                        'val_r': '',
+                        'val_c': '',
+                        'style': None,
+                    }
+                    row_data['cells'].append(col_data)
```
sbidoul: Is this change necessary? If yes why?

sbidoul:
@jbeficent the issues with the complex domains in the view, as well as the MisKpiData tests are resolved.

I still need fully dive into the js part. I've some doubts I wrote in the review comments. To be further investigated.

mis_builder/models/kpimatrix.py
JordiBForgeFlow: Yes. This fix prevented errors raising when you attempted to create lines.

mis_builder/models/mis_report_instance.py
JordiBForgeFlow: You need to use a dummy field in order to be able to render the form view into this field.

 I tested the various methods to render the dynamic view. The method used in v10, using the declaration of <widget>, https://github.com/OCA/mis-builder/blob/10.0/mis_builder/views/mis_report_instance.xml#L17, does not work in v11 anymore.
I tried using client actions as used in https://github.com/odoo/odoo/blob/11.0/addons/stock/views/stock_picking_views.xml#L235, but this would not let me  add the view to the dashboard.

Perhaps there's another way, but I could not find it.

sbidoul: Perhaps we could bind the widget to the `id` field? So we could use the same widget on any m2o field that points to a `mis.report.instance`?

JordiBForgeFlow: The problem is that with this approach we need to define  <field name="mis_report_view" widget="mis_builder_widget">. And it is "mis_report_view" the field that renders the full representation. I would not know how to bind the widget to the id field.

mis_builder/static/src/js/mis_builder_widget.js
JordiBForgeFlow: This seems to be the standard method to call to backend methods in v11.

mis_builder/models/mis_report_instance.py
sbidoul: But where is that `mis_report_view` field computed with a real value? I can't find it in the code.

JordiBForgeFlow: The content is defined in the _render method in the JS side https://github.com/OCA/mis-builder/blob/11.0-dev/mis_builder/static/src/js/mis_builder_widget.js#L44

JordiBForgeFlow: So, the field has a widget, and the widget renders the content of this field. It really does happen to need any content from the field itself.

sbidoul: Yes but that does not use the `mis_report_view` field at all. So the widget could be bound to any field and the result would be identical?

JordiBForgeFlow: Do you suggest <field name=id widget="mis_builder_widget">?

sbidoul: https://github.com/OCA/mis-builder/pull/67/commits/fd26f817ad5582028186b983368299cb31e021db is a first simplification.

mis_builder/views/mis_report_template.xml
JordiBForgeFlow: I think it can be removed.

mis_builder/static/src/js/mis_builder_widget.js
```diff
@@ -0,0 +1,121 @@
...
+    //--------------------------------------------------------------------------
+
+    /**
+     * @override
+     * @private
+     */
+    _render: function () {
+        var self = this;
+        var context = self.getParent().state.context;
+        var rec_id = this.recordData.id || self.getParent().state.context.active_id;
```
sbidoul: Here we must use the field value. Not sure active_id makes sense (TBC).

mis_builder/static/src/js/mis_builder_widget.js
sbidoul: Ah ok then. So we need to convert the `drilldown` arg to json here.

mis_builder/static/src/js/mis_builder_widget.js
JordiBForgeFlow: The problem I had is when you add to the dashboard. It is in those cases where I needed. Because in the dashboard the record is not loaded. This has changed in the definition of the dashboard from v10, as far as I can tell.

mis_builder/static/src/js/mis_builder_widget.js
JordiBForgeFlow: Done

mis_builder/views/mis_report_template.xml
JordiBForgeFlow: Removed.

JordiBForgeFlow:
I will look at this proposal from @lasley to define classes during testing https://github.com/OCA/maintainer-tools/pull/242. Perhaps it is better than what we used, or vice-versa?

lasley:
It was adapted from an Acsone strategy (from Alfodoo), which originally looked very similar to the one you have here. It received a bunch of fixes from @obulkin though as the abstract models began to be implemented in descendent modules. Off the top of my head, I remember there were also some database mutations being missed that caused major headaches.

sbidoul:
@jbeficent @lasley I believe the [solution](https://github.com/OCA/mis-builder/pull/67/files#diff-67ee724417786e97c27225738ceb907b) I reached in this PR for the test model is simple and clean and leaves the database clean. It works because `auto_init` does not commit in v11 (whereas it did in 10).

JordiBForgeFlow:
@lasley @obulkin I also implemented @sbidoul approach here: https://github.com/OCA/server-tools/blob/11.0/base_exception/tests/common.py with success. Perhaps you can review if this proposal works better also for you and we can propose an improved template in maintainer-tools.

sbidoul:
@jbeficent in my tests the teardown part is useful too to avoid warnings. Note this approach is very close to what Odoo does when creating/removing custom models through the UI: that's where I found inspiration in the end.

JordiBForgeFlow:
@sbidoul I had to remove the reardown in the other module in https://github.com/OCA/server-tools/blob/11.0/base_exception/tests/common.py, though, because it would not allow me to uninstall the module. Gave me a nasty error indicating that the model was not found in the registry.

obulkin:
At a glance, it does seem like all the problematic commits have been removed from v11, so there's probably no need to put the registry into test mode in this version. However, you may still want to do something along the lines of the following:

```
cls.registry[cls.test_model._name]._abstract = True
cls.registry[cls.test_model._name]._auto = False
```

This should prevent warnings about the model not having a table, while avoiding any errors caused by descendent code that expects it to be in the registry. At the very least, that was my experience in v10, where my first solution was also to just delete the model from the registry.

sbidoul:
Rebased. @jbeficent with fb3654e I'm now satisfied with the widget. 

I rebased a last time, and removed the WIP tag.

I think we are good now!

jbeficent:
https://media.giphy.com/media/l0MYJnJQ4EiYLxvQ4/giphy.gif
image

JordiBForgeFlow:
@sbidoul Good! Thanks for your help! I will do a last review on monday.

mis_builder/__manifest__.py
```diff
@@ -1,10 +1,9 @@
-# -*- coding: utf-8 -*-
 # Copyright 2014-2017 ACSONE SA/NV (<http://acsone.eu>)
```
simahawk, line 1: you might want to raise year up to 2018 ;)

mis_builder/__manifest__.py
```diff
@@ -1,10 +1,9 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
     'name': 'MIS Builder',
-    'version': '10.0.3.1.1',
+    'version': '11.0.3.1.1',
```
simahawk, line 6: by convention should be "11.0.0.0" but I understand if you want to keep internal versions in line :)

mis_builder/datas/ir_cron.xml
```diff
@@ -7,9 +7,8 @@
             <field name="interval_type">hours</field>
             <field name="numbercall">-1</field>
             <field eval="False" name="doall"/>
-            <field eval="'mis.report.instance'" name="model"/>
-            <field eval="'_vacuum_report'" name="function"/>
-            <field eval="'(24,)'" name="args"/>
+            <field ref="model_mis_report_instance" name="model_id"/>
+            <field name="code">model._vacuum_report(24, 0)</field>
```
simahawk: `hours=24` (default, BTW). The second number is? I don't see it here https://github.com/OCA/mis-builder/blob/fb3654e75551a8d74b8b73b1aec8ac6b2c795816/mis_builder/models/mis_report_instance.py#L466

mis_builder/report/mis_report_instance_qweb.xml
```diff
@@ -1,34 +1,28 @@
...
     <template id="report_mis_report_instance">
-      <t t-call="report.html_container">
+      <t t-call="web.html_container">
         <t t-foreach="docs" t-as="o">
-          <t t-call="report.internal_layout">
+          <t t-call="web.internal_layout">
             <t t-set="matrix" t-value="o._compute_matrix()"/>
             <t t-set="style_obj" t-value="o.env['mis.report.style']"/>
             <div class="page">
+              <link href="/mis_builder/static/src/css/report.css" rel="stylesheet"/>
```
simahawk: any good reason to keep the css here instead of assets?

simahawk (approved):
Code LGTM. Just some questions...

mis_builder/report/mis_report_instance_qweb.xml
sbidoul: good question. @jbeficent do you know why this change was made?

mis_builder/__manifest__.py
sbidoul: Yes, I do the effort of maintaining feature parity across 9 10 11 so this version number expresses that.

mis_builder/datas/ir_cron.xml
sbidoul: @simahawk fixed

astirpe:
I just noticed this line:
https://github.com/OCA/mis-builder/blob/11.0-dev/mis_builder/views/mis_report_instance.xml#L93
and I would ask you to replace the old `colors=` syntax with the new `decoration-*=` way to define the colors of rows.

mis_builder/report/mis_report_instance_qweb.xml
sbidoul: @simahawk is it better now?

sbidoul:
@astirpe `color="red:...` replaced with `decoration-danger`. Thanks for pointing this out.

mis_builder/report/mis_report_instance_qweb.xml
simahawk: @sbidoul should I see any difference? :thinking:

sbidoul: @simahawk yes (https://github.com/OCA/mis-builder/pull/67/commits/0237f5944246ac67111cfdbb220f87a60028ce63). Apparently github has difficulties showing the diff inside the conversation. But the full diff is correct.

simahawk: ahh now I see it. LGTM :smile:

JordiBForgeFlow (approved):

JordiBForgeFlow:
I am good with the proposed changes!

astirpe:
Thank you for this!
I think the default branch can be set to 11.0 ;)

JordiBForgeFlow:
Thank you @sbidoul !!

PR #68, opened 2018-02-04
[FIX] mis_builder_budget: properly handle NameDataError

sbidoul:
Similar to what is done [here](https://github.com/OCA/mis-builder/blob/23a0f11fcf6b1321ed79ffea355292a8c95a9f59/mis_builder/models/mis_report.py#L782-L784).


elicoidal (approved):
LGTM

elicoidal:
(with functional check in OCA instance :wink: )

PR #79, opened 2018-04-14
Add docs directory

sbidoul:
@elicoidal


docs/make.bat
```diff
@@ -0,0 +1,36 @@
+@ECHO OFF
```
elicoidal, line 1: I think you do not need this file (but doesnot hurt)

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
+Introdution
```
elicoidal, line 1: Introduction

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+===========
+
+MIS Builder is an Odoo module that implements a class of Odoo reports where KPI 
+(Key Performance Indicators)
```
elicoidal, line 5: improve the format

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+  and very complex account charts.
+- Use the same template for different reports.
+- Compare data over different time periods.
+- User-configurable styles, rendered perfectly in the UI as well as Excel and PDF exports.
+- Interactive display with drill-down.
+- Export to PDF and Excel.
+- A budgeting module.
+- Evaluate KPI over various data sources, such as actuals, simulation, committed costs
+  (some custom development is required to create the data source).
+- For developers, the accounting balance computation engine is exposed as an easy
```
elicoidal, line 23: easy-to-use

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+- Very fast balance reporting for accounting data, even on million lines databases
+  and very complex account charts.
+- Use the same template for different reports.
+- Compare data over different time periods.
+- User-configurable styles, rendered perfectly in the UI as well as Excel and PDF exports.
+- Interactive display with drill-down.
+- Export to PDF and Excel.
+- A budgeting module.
+- Evaluate KPI over various data sources, such as actuals, simulation, committed costs
+  (some custom development is required to create the data source).
```
elicoidal, line 22: remove `some`

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+===========
+
+MIS Builder is an Odoo module that implements a class of Odoo reports where KPI 
+(Key Performance Indicators)
+are displayed in rows, and time periods in columns. It focuses on very fast reporting
+on accounting data but can also use data from any other Odoo model.
+
+It features the following key characteristics:
+
+- User configurable: end users can create new report templates without development,
```
elicoidal, line 11: The list is a little bit eclectic: it mixes functional features on one side and usability on the other. It would maybe be a good idea to split it in 2.

docs/index.rst
```diff
@@ -0,0 +1,24 @@
+.. MIS Builder documentation master file, created by
+   sphinx-quickstart on Sat Apr 14 14:08:55 2018.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to MIS Builder's documentation!
+=======================================
+
+Management Information System reports for Odoo: easily build super fast,
```
elicoidal, line 9: " easily build, super fast,"
(lol: sounds like fp :wink: )

docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
elicoidal, line 1: In your case, it implies we maintain all the build technical files in all modules: it looks heavy and error prone (conf.py could become inconsistent over modules). I supposed this is the way readthedoc.io works (never used it). 

I adopted a different approach about Sphinx with a dockerized server, building in a separate environment to build the doc based on `index.rst` files only. Some of the work is published here: https://github.com/Elico-Corp/readthedoc-oca-docker. This is only a PoC that works 'fine' here: https://oca-doc.elico-corp.com/ and based on the README files but I have the same concept for the company documentation for which I only need an `index.rst` file in each directory (I could publish it too). 
We would just need to start a docker in a server and gather all files published in the repos/docs directory.

My point is having the Sphinx files here is fine for one repo but looks looks heavy to maintain if we want a generic solution for OCA documentation.

docs/index.rst
```diff
@@ -0,0 +1,24 @@
+.. MIS Builder documentation master file, created by
+   sphinx-quickstart on Sat Apr 14 14:08:55 2018.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to MIS Builder's documentation!
```
elicoidal, line 6: I would suggest a line before to create a document title

docs/index.rst
```diff
@@ -0,0 +1,24 @@
...
+=======================================
+
+Management Information System reports for Odoo: easily build super fast,
+beautiful, custom reports such as P&L, Balance Sheets and more.
+
+.. toctree::
+   :maxdepth: 3
+   :caption: Contents:
+
```
elicoidal, line 15: I suggest to add the following (to be tested though in readthedoc):
```
.. sectnum::

.. header:: |DOCNO| - |TITLE| - |DATE|

.. footer:: Copyright Creative Common - Odoo Community Association
```

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+  using simple Excel-like formulas.
+- Very fast balance reporting for accounting data, even on million lines databases
```
elicoidal, line 13: million-line

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+  and very complex account charts.
```
elicoidal, line 14: and -> with

docs/intro.rst
```diff
@@ -0,0 +1,25 @@
...
+- Use the same template for different reports.
+- Compare data over different time periods.
+- User-configurable styles, rendered perfectly in the UI as well as Excel and PDF exports.
```
elicoidal, line 17: perfectly rendered

docs/conf.py
```diff
@@ -0,0 +1,159 @@
...
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = u'MIS Builder'
+copyright = u'2018, Odoo Community Association (OCA)'
```
elicoidal, line 23: do you need to add the license?

elicoidal:
@sbidoul I will prepare the PR based on your file structure

docs/Makefile
sbidoul: So far the config is just the result of running `sphinx-quickstart` and entering the doc title and author. So that's how [sphinx works out of the box](http://www.sphinx-doc.org/en/1.6/tutorial.html). readthedocs just happens to autodetect that structure.

We can imagine a mechanism to provide OCA defaults for the sphinx configuration in the same way across repos. We must be careful to not restrict sphinx features and extension mechanisms though. Perhaps we could have a default OCA config that each repo could import in its own `docs/conf.py`. I'd prefer that over an OCA specific doc builder: if it remains standard sphinx, we can benefits from tools such as readthedocs.io.

Anyway it's probably too early an optimisation. AFAICT, this is going to be the second repo with doc so far (the first one being [connector](https://github.com/OCA/connector/tree/10.0/connector/doc)).

One thing I'd like to investigate first is a mechanism to somehow have the addon README.rst generated by including rst fragments into the boilerplate template. That would facilitate it's maintenance. In your PR you can think about having short fragments that are reusable by including them in README.rst that would be great.

docs/make.bat
sbidoul: I figured it would be nice to leave it so non-technical doc authors can build the doc on windows too.

docs/index.rst
sbidoul: These directives [must not be used with sphinx](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives). 
* section numbering is done by adding `:numbered:` in the toctree directive
* header/footer are generated from `conf.py` variables and the theme/templates

docs/conf.py
sbidoul: CC BY license added done at https://github.com/OCA/mis-builder/commit/b6c6bf5e2700b6a115efa0d9f820f5fadfd36a14#diff-cac3dc3bfd42bc240bc9eb924f8dbfdd

docs/Makefile
sbidoul: @elicoidal even better, we should build an OCA sphinx theme with all the defaults we want, so `conf.py` can become as small as we want yet completely extensible.

We can use this project to experiment with the theme then package it when we are satisfied with the result.

Current result is [here on readthedocs](http://oca-mis-builder.readthedocs.io/en/10.0/).

elicoidal: > One thing I'd like to investigate first is a mechanism to somehow have the addon README.rst generated by including rst fragments into the boilerplate template. 

Complex because each module only has fragment while the documentation presents a full intertwined use case with multiple modules (in your case only 2 but in BR we have almost 20). I can try to rework the structure but I fear it will hit some wall.

sbidoul: @elicoidal here is more concretely what I have in mind.

In each module, add a `docs/` directory containing `introduction.rst`, `installation.rst`, `configuration.rst`, `usage.rst`, `roadmap.rst` and `credits.rst`. I'd add `changes.rst` too.

Each of these file must contain a reStructuredText fragment *without section title*. Only introduction, usage (and maybe credits) must be present, the other are optional.

From that and the addon manifest (for the module name, license, development_status) we can easily generate README.rst in our nightly job.

It is also easy to include the fragment where we see fit in the repo-level documentation eg with `.. include:: ../addon/docs/introduction.rst`, as I already did with the `CHANGES.rst`.

What do you think? We can try here if it works we can propose to `https://github.com/OCA/maintainer-tools/tree/master/template/module`.

elicoidal: @sbidoul I think this is the way to go. It might not be 100% generic (or better said genericity might not be a fit for 100% of cases) but IMO this will cover most of cases. Later on we might think about adding some crossover documents.
This is partially what I did yesterday (#80) but I need to go one step further on the split. I will do that today (hopefully)

elicoidal: I am thinking about the following structure:
* mis_builder.intro.rst
* mis_builder_budget.intro.rst
And mix them all in the index.rst.
Building the README should not be that complex
etc...

PR #80, opened 2018-04-15
[10.0][ADD][WIP] add functional doc version1

elicoidal:
@sbidoul please check this first draft and comment: I have taken some liberties so do not hesitate to feedback so that I can adjust :smile: . 

Some additional comments:
* I couldnot set up a logo in the conf.py as it needs to be local and cannot be URL.
* I personally prefer the theme `classic` but did not want to impose myself :wink:
* I am not sure the following lines work well in this kind of document (works fine in rst2pdf in single file):
```
.. sectnum::
.. header:: MIS Builder's documentation
.. footer:: Copyright Odoo Community Association - Creative Common
```
* I tested in my local with a simple  `make html`  and output looks OK (except for above)


elicoidal:
Besides: please confirm the Creative Common License

docs/intro.rst
```diff
@@ -1,25 +1,89 @@
...
 
-- User configurable: end users can create new report templates without development,
-  using simple Excel-like formulas.
-- Very fast balance reporting for accounting data, even on million lines databases
-  and very complex account charts.
-- Use the same template for different reports.
-- Compare data over different time periods.
-- User-configurable styles, rendered perfectly in the UI as well as Excel and PDF exports.
+- User-friendly configuration: end users can create new report templates without
+  development, sing simple Excel-like formulas.
```
sbidoul: s/sing/using/

docs/setup.rst
```diff
@@ -0,0 +1,99 @@
+Install the module in Odoo
+==========================
```
sbidoul, line 2: This section is not specific to this repo. Maybe this should become a page on the OCA website where we can point to?

docs/setup.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Using pip
+*********
+An easy way to install it with all its dependencies is using pip:
+
+#. Recover the code from pip repository
+
+   .. code-block:: sh
+
+      pip install odoo10-addon-mis_builder odoo-autodiscover
```
sbidoul: add `--pre` to be sure to get the last `.devN` version. That's still a part that annoys me that we cannot control precisly which version goes to pypi. Anyway, that's another other story.

docs/usage.rst
```diff
@@ -0,0 +1,280 @@
...
+budgetable. You also need to configure the accumulation method on the KPI according
+to their type.
+
+The accumulation method determines how budgeted values spanning over a time period
+are transformed to match the reporting period.
+
+* **Sum**: values of shorter period are added, values of longest or partially overlapping
+  periods are adjusted pro-rata temporis (eg monetary amount such as revenue).
+* **Average**: values of included period are averaged with a pro-rata temporis weight.
+  Typically used for values that do not accumulate over type (eg a number of employees).
```
sbidoul: s/over type/over time/

docs/usage.rst
```diff
@@ -0,0 +1,280 @@
...
+
+#. Select the KPI to be budgeted
+#. Select the date range or From/To
+#. provide Amount in Company currency
+#. Provide the corresponding Analytic account
+
+.. note:: To be noted that the budget could be prepared in Excel and easily imported via Odoo Standard Function.
+
+   .. image:: images/14.png
+      :width: 1800
```
sbidoul: Note for this screenshot you can keep only the following columns:
* kpi_expression_id: with the kpi name or kpi_name.subkpi_name in presence of subkpi (no need to lookup kpi xmlids, the name search will do its magic as it knows the budget context)
* date_from
* date_to
* amount
* analytic account

The budget_id column is defaulted from the context when in the budget item views. And the date range is just a UI helper to enter date from/to.

elicoidal:
@sbidoul please check this PoC.

We would need to decide/fix the following:
* Flat index.rst or use of include (first easier to maintain but second better IMO for the title levels)
* Fix the first level structure
* Level underlying convention
* Final directory structure and Filenames

docs/setup.rst
elicoidal: Either this or we add it in the README.md
@sbidoul

elicoidal:
@sbidoul any next step or still waiting about https://github.com/OCA/maintainer-tools/issues/335 ?

sbidoul:
I made a last proposal in OCA/maintainer-tools#335

Note the text fragments there should remain relatively short (ie if they become too large, they must be reduced to "quick start" introduction for the README, the more complete version being expanded in the repo-level doc).

elicoidal:
> Note the text fragments there should remain relatively short (ie if they become too large, they must be reduced to "quick start" introduction for the README, the more complete version being expanded in the repo-level doc).

Not sure to full understand

elicoidal:
@sbidoul shall we close this one?

sbidoul:
@elicoidal why no! I merged the readme changes. The fragments naming scheme should be stable now. So you can redo this PR by including the relevant fragments at the right place. You need to redo the editorial changes you made to DESCRIPTION and USAGE too (sorry about that). My suggestion is to adapt USAGE.rst so it becomes a kind of quickstart section of the larger documentation here.

elicoidal:
@sbidoul OK! would you accept one single PR for the whole repo?
I might be able to do it this week end.

elicoidal:
@sbidoul I am working on the documentation and I have the following questions / issues:
* the install section: If the idea is to explain how to install a repo or module in Odoo, I think we'd better have a common webpage that would be referred as part of the boilerplate. I am not sure what else could there be in that section (other topics would mostly be part of CONFIGURATION see next)
* I am missing a CONFIGURATION section on top of usage
* What about the images? usually we use rst links but I have seen that you use github static content links which is OK but only when the PR is merged which is kind of annoying for the review

sbidoul:
@elicoidal 

* About the install section: agreed we need a common web page for that, so there is no need for one in the mis_builder doc, feel free to drop it if you like.
* If you add a readme/CONFIGURE.rst, it will be included in the README in a Configuration section. That said, for mis builder, one could argue that the configuration is the usage or vice versa. But feel totally free to make one if you think it's easier to understand.
* Regarding the images, I used github links for images that end up in the readme, so they render correctly everywhere (pypi, apps.odoo.com, inside Odoo, etc). For images to be used in the sphinx doc but not in readme fragments, just add them somewhere in docs (usually docs/_static/) and reference them using relative paths.

elicoidal:
Superseded by #92

PR #87, opened 2018-04-20
Add pbal, nbal accounting expressions

### LINKED ISSUE #86: Filtering positive/negative account balances
Use case
=====

In Spain, we have several sections in the Balance Sheet that includes the sum of balances that are greater than 0 in one side, and the opposite in the other side. They are for example the accounts 551. But the balance is not taken as a whole, but account per account. Example:

- You have in the account 551001 positive balance: then it should go to one side.
- You have in the account 551002 negative balance: then i should go to the other side.

So putting the expression `(bale[551%] if bale[551%] > 0)` doesn't work, and you can't define the expression if a fixed way like `(bale[551001] if bale[551001] > 0) + (bale[551002] if bale[551002] > 0) + ...` because there are infinite accounts: you have 551001, 551002, but also 5510001, 5510002... (with 7 digits), and so on.

Proposed solution
========

Add an account expansion mechanism at expression level. This can be an optional fifth letter in the expressions (for example, `a`). If included, mis_builder will switch the expression before passing it to the "SQL parsing" for making the expansion by account. Example: if you put the expression `+(balea[551%] if balea[551%] > 0)`, it searches on the system the account selector, and unfolds the expression with the result:
```
+(bale[551000] if bale[551000] > 0)+(bale[551001] if bale[551001] > 0)+(bale[551002] if bale[551002] > 0)
```
and then continues the normal pipe

I don't know enough yet the mis_builder code, but maybe with a little guidance I can implement it.

sbidoul:
Sum accounts with positive/negative balance only.

So, for example, `pbale[551%]` will show the sum of all 551 accounts with a positive ending balance.

TODO
- [x] check it covers the use case @pedrobaeza 
- [x] unit test
- [x] handle decimal precision => not necessary in this case
- [x] update legend

Closes #86

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -80,7 +80,7 @@ class AccountingExpressionProcessor(object):
     MODE_UNALLOCATED = 'u'
 
     _ACC_RE = re.compile(
-        r"(?P<field>\bbal|\bcrd|\bdeb)"
+        r"(?P<field>\bbal|\bpbal|\bnbal|\bcrd|\bdeb)"
         r"(?P<mode>[piseu])?"
         r"\s*"
         r"(?P<account_sel>_[a-zA-Z0-9]+|\[.*?\])"
@@ -384,6 +384,10 @@ class AccountingExpressionProcessor(object):
                                          (AccountingNone, AccountingNone))
                 if field == 'bal':
                     v += debit - credit
+                elif field == 'pbal' and debit >= credit:
+                    v += debit - credit
+                elif field == 'nbal' and debit < credit:
+                    v += debit - credit
                 elif field == 'deb':
                     v += debit
                 elif field == 'crd':
@@ -421,6 +425,16 @@ class AccountingExpressionProcessor(object):
                                      (AccountingNone, AccountingNone))
             if field == 'bal':
                 v = debit - credit
+            elif field == 'pbal':
+                if debit >= credit:
+                    v = debit - credit
+                else:
+                    v = AccountingNone
+            elif field == 'nbal':
+                if debit < credit:
+                    v = debit - credit
+                else:
+                    v = AccountingNone
             elif field == 'deb':
                 v = debit
             elif field == 'crd':
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -35,13 +35,15 @@ class AccountingExpressionProcessor(object):
 
     Expressions of the form <field><mode>[accounts][optional move line domain]
     are supported, where:
-        * field is bal, crd, deb
+        * field is bal, crd, deb, pbal (positive balances only),
+          nbal (negative balance only)
         * mode is i (initial balance), e (ending balance),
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Thank you very much for the quick reaction! Let me known when do you want me to test (I'm on holidays today though)

sbidoul:
It should work as is, so you can test when you want. I'll not work on it before next week though. I mainly need your confirmation it solves your use case before finishing it cleanly.

mis_builder/models/aep.py
```diff
@@ -421,6 +425,10 @@ def f(mo):
                                      (AccountingNone, AccountingNone))
             if field == 'bal':
                 v = debit - credit
+            elif field == 'pbal' and debit >= credit:
+                v += debit - credit
```
pedrobaeza: Here (and below) should be assignation or we get:
```
UnboundLocalError: local variable 'v' referenced before assignment
```

mis_builder/models/aep.py
```diff
@@ -421,6 +425,10 @@ def f(mo):
                                      (AccountingNone, AccountingNone))
             if field == 'bal':
```
pedrobaeza, line 428: We should put above `v = AccountingNone` for avoiding a later error

pedrobaeza:
With the 2 changes I have mentioned inline, this works. This is the positive part (`pbale[551%]`):

seleccion_034

and this is the negative (`nbale[551%]`):

seleccion_035

coming from a simple journal entry:

seleccion_033

The only cons I see is that the other account with moves still appears in the account expansion detail, but I think it can be complicated to hide it. @jbeficent what do you think? Does it deserve to invest time in hiding it?

mis_builder/models/aep.py
sbidoul: That was a copy-paste error indeed.

mis_builder/models/aep.py
sbidoul: Not sure about this one, once the other error has been fixed.

pedrobaeza: This is still missing

sbidoul: Ah yes, I see it now

pedrobaeza: Yes, I needed to add it for avoiding error later

sbidoul:
> Does it deserve to invest time in hiding it?

Normally the "Hide Empty" style should be enough. I don't know why these line show as 0 instead of empty in your example though.

pedrobaeza:
That's because we can't hide empty values on the P&L and BS.

JordiBForgeFlow:
@sbidoul @pedrobaeza I think that it is good as it is now. Showing all the accounts won't hurt. It's good to go for me!

JordiBForgeFlow (approved):

pedrobaeza:
Thank you very much for this! Now waiting for the back/forward-porting of these features to 9.0/10.0

PR #90, opened 2018-05-05
[WIP] Generated readme proof of concept

sbidoul:
This is to illustrate https://github.com/OCA/maintainer-tools/pull/339

Here is the resuling README.rst for [mis_builder](https://github.com/acsone/mis-builder/blob/10.0-generated-readme/mis_builder/README.rst), [mis_builder_budget](https://github.com/acsone/mis-builder/blob/10.0-generated-readme/mis_builder_budget/README.rst) and [mis_builder_demo](https://github.com/acsone/mis-builder/blob/10.0-generated-readme/mis_builder_demo/README.rst).

To be noted, compared to the current readme template:
- a table of contents after the description
- maintainers avatars in the Maintainers section (from the new `maintainers` manifest key)
- maturity level badge (from the new `development_status` manifest key), pending final concept from Valeria)
- addition of an Authors section (from the manifest `author` key, with OCA removed)
- better how to contribute link at the bottom, including a link to the github repo

Please comment on the structure in https://github.com/OCA/maintainer-tools/pull/339, not here.


elicoidal (approved):
LGTM

mis_builder/README.rst
```diff
@@ -1,6 +1,19 @@
-.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+.. !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+   !! This file is generated by oca-gen-addon-readme !!
+   !! changes will be overwritten.                   !!
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
pedrobaeza: Shouldn't this be PNG for visibility in the Apps?

sbidoul: Good catch, thanks!

pedrobaeza:
About the separation in authors and contributors, is this only in mis-builder or something general you want to put?

sbidoul:
Authors come from the manifest. Contributors from CONTRIBUTORS.rst. 
So it would be everywhere.

(could please keep all comments in https://github.com/OCA/maintainer-tools/pull/339, instead od here)

docs/history.rst
```diff
@@ -0,0 +1,12 @@
+History
```
elicoidal, line 1: I thought the titles should not be in the separate files.

sbidoul: There should not be section titles in the rst fragments that are included in the README. My reasoning is that we want to avoid the README to become too complex, and avoid section level conflicts when including the fragments in a sphinx doc. Here we are in the sphinx doc where the author controls everything so there are no such constraints.

Note HISTORY.rst is not included in the README and for that one I kept section titles in the [example](https://raw.githubusercontent.com/acsone/mis-builder/10.0-generated-readme/mis_builder_budget/HISTORY.rst).

sbidoul:
@OCA/mis-builder-maintainers

PR #92, opened 2018-05-27
 [ADD][WIP] User documentation using README with fragments 

elicoidal:
Supersedes #80 
This is an advanced WIP:
* not clear the place where to store the images so that it is visible when editing the fragments and easy to convert in the final README (current place in `readme/images` makes it easy).
* another option would be that the script automatically converts in the final document to static Github URL 
* not clear the reference links within the fragments. If a repo documentation is built later, this might need discussion or rule (it should not be able to add a link to a reference outside the module)
* the titles levels should be in guidelines. Not important for standalone modules, it would be essential if someday we want to build a common documentation with Sphinx
* I took the liberty to suggest some modifications in the `README.md`

cc @OCA/mis-builder-maintainers


sbidoul:
@elicoidal there is a misunderstanding on the purpose of the readme fragments.

The readme fragments [must not contain section titles](https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/USAGE.rst), otherwise that will most likely break the readme generator. The only exception so far is HISTORY.rst.

In my view it's not possible to fit the full documentation of something as complex as mis builder in a README format. Hence the sphinx approach.

The idea is to have short introductory content in the readme fragments, and reuse them by inclusion in a broader sphinx documentation like [here](https://raw.githubusercontent.com/OCA/mis-builder/10.0/docs/quickstart.rst) or [here](https://raw.githubusercontent.com/OCA/mis-builder/10.0/docs/history.rst). And then add specific content in the sphinx `docs` directory and [index.rst](https://github.com/OCA/mis-builder/blob/10.0/docs/index.rst), to generate the full documentation (possibly on [readthedocs](http://oca-mis-builder.readthedocs.io/en/10.0/)).

elicoidal:
@sbidoul 
indeed, I thought the README would cover the whole documentation (as in this PR).

What kind of file structure do you expect in the `docs` directory? following the same file structure as `readme`?

sbidoul:
Really, the readme won't work for a big document. If only because of the images.

I don't have any specific structure in mind for a sphinx doc. Whatever works best for a repo-level detailed documentation. There is a skeleton in the 10.0 branch.

elicoidal:
OK: will make a proposal based on the initial PR #80

PR #98, opened 2018-06-13
[11.0][mis_builder] Implement hide_always

JordiBForgeFlow:
Allows to hide kpis when this flag 'hide_always' is set in the assigned style.

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -446,7 +446,8 @@ class KpiMatrix(object):
 
         body = []
         for row in self.iter_rows():
-            if row.style_props.hide_empty and row.is_empty():
+            if (row.style_props.hide_empty and row.is_empty()) or \
+                    row.style_props.hide_always:
                 continue
             row_data = {
                 'row_id': row.row_id,
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -36,6 +36,7 @@ PROPS = [
     'dp',
     'divider',
     'hide_empty',
+    'hide_always',
 ]
 
 TYPE_NUM = 'num'
@@ -137,6 +138,8 @@ class MisReportKpiStyle(models.Model):
                                default='1')
     hide_empty_inherit = fields.Boolean(default=True)
     hide_empty = fields.Boolean(default=False)
+    hide_always_inherit = fields.Boolean(default=True)
+    hide_always = fields.Boolean(default=False)
 
     @api.model
     def merge(self, styles):
--- a/mis_builder/report/mis_report_instance_qweb.xml
+++ b/mis_builder/report/mis_report_instance_qweb.xml
@@ -66,7 +66,7 @@
                 </div>
                 <div class="mis_tbody">
                   <t t-foreach="matrix.iter_rows()" t-as="row">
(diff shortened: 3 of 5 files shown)


mis_builder/tests/test_render.py
```diff
@@ -209,6 +212,7 @@ def test_merge(self):
         self.assertEquals(style_props, {
             'color': '#00FF00',
             'dp': 1,
+            'hide_always': False,
         })
```
sbidoul: why do you need to change this test?

JordiBForgeFlow:
We're not going to use this feature. So closing the PR for now.

JordiBForgeFlow:
I reconsidered and we'll use this feature :)

sbidoul:
This PR should work, right?

JordiBForgeFlow:
Yes, it works. I have been using it to prepare the financial ratios for
spain

sbidoul:
Cool!

JordiBForgeFlow:
@sbidoul See the financial ratios proposed for Spain. https://github.com/OCA/l10n-spain/pull/871 Perhaps you might be interested to use a similar approach for BE, LU

sbidoul:
Nice! So you are overriding the reports to add new lines with the ratios, correct?

JordiBForgeFlow:
I had to duplicate the reports from l10n_es_mis_report, not override.

Even when the xmlids are the same, since the reports belong to a different
module the complete xmlid is different.

I duplicated the balance sheet and profit and loss, hiding all the lines so
that the ratio kpis can refer to them.

Either you do that, or you directly add the ratios to the BS and P&L, but
that can be inconvenient if you want to distribute these reports, without
the ratios.

sbidoul:
I see. I'll dig this idea of having the formulas of one report available in the computations of another.

sbidoul:
Have you seen my comment about the changed test above? Except that this PR looks good to me.

JordiBForgeFlow:
Yes, I attended your comment

sbidoul:
I don't understand. The test is still modified.

JordiBForgeFlow:
Hmm will check

mis_builder/tests/test_render.py
JordiBForgeFlow: 'hide_always' does not have a corresponding 'hide_always_inherit', and correspondingly will be returned here https://github.com/OCA/mis-builder/blob/11.0/mis_builder/models/mis_report_style.py#L157

JordiBForgeFlow: Perhaps I should add 'hide_always_inherit'?

sbidoul: I was wondering too

JordiBForgeFlow: Hmm, now that I think about it, when you hide a higher level it should always hide lower levels. So I will create 'hide_always_inherit'.

JordiBForgeFlow:
@sbidoul I have now added 'hide_always_inherit' and makes total sense.

sbidoul (approved):
Thanks! I'll merge this in 10.0 and port to 9 and 11

sbidoul:
Merged at c3f41b933ddf4da6fac4362aecb4e2b4d33e9e0a

sbidoul:
Ported to 11 and 9

PR #107, opened 2018-07-08
in AEP, use active_test=False

sbidoul:
This is important to consider all move lines in computations.
Without this, a formula based on deactivated taxes would
return no result, which is wrong since the corresponding
move lines do exist.

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -189,6 +189,8 @@ class AccountingExpressionProcessor(object):
 
     def done_parsing(self):
         """ Replace account domains by account ids in map """
+        account_model = self.env['account.account'].\
+            with_context(active_test=False)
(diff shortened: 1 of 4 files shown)


sbidoul:
@jbeficent @benwillig do you have an opinion on this one?

benwillig:
In my opinion, it's logic to disable the active test. But maybe it should be an option in the MIS report instance (or maybe a checkbox in the instance view). If not the user should be informed that we also get data linked to inactive objects, because if the user check the journal items, he might be confused about the differences between the report and the data shown in the list view.

sbidoul:
@benwillig I thought about that too, but I could not imagine a realistic situation where this could cause in issue in practice. Do you have one in mind?

benwillig:
No I have no real use case.

sbidoul:
Tested in production.

PR #108, opened 2018-07-08
int(), float() and round() support for AccountingNone

sbidoul:


sbidoul:
Merging this trivial fix.

PR #111, opened 2018-08-01
[10.0][IMP] Add as_value() method to export current report as values.…

robyf70:
This PR is just a proposed improvement in order to associate, where possible, a data range to a column and propagate this information down to the kpimatrix object so that it's available as normal date object. 

I've also added a new as_value() method as well as new compute_values() in MisReportInstance in order to get raw dictionary values where the key maps a row of columns.

All of this is in preparation of a new module where we can plot some data in a dashboard.

Let me know your thought.

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -77,7 +77,7 @@ class KpiMatrixRow(object):
 
 class KpiMatrixCol(object):
 
-    def __init__(self, key, label, description, locals_dict, subkpis):
+    def __init__(self, key, label, description, locals_dict, subkpis, date_from=None, date_to=None):
         self.key = key
         self.label = label
         self.description = description
@@ -85,6 +85,8 @@ class KpiMatrixCol(object):
         self.colspan = subkpis and len(subkpis) or 1
         self._subcols = []
         self.subkpis = subkpis
+        self.date_from = date_from
+        self.date_to = date_to
         if not subkpis:
             subcol = KpiMatrixSubCol(self, '', '', 0)
             self._subcols.append(subcol)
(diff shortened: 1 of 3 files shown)


sbidoul:
Hi! Thanks for contributing to mis builder.

By design we want to have the KPI matrix as dumb as possible. That goal is not fully achieved yet but the general idea is that it should [remain a kind of presentation layer](https://github.com/OCA/mis-builder/blob/515290d86c10fd02892420cb30dcd167fba0831d/mis_builder/models/kpimatrix.py#L25-L28) with very limited knowlege of the actual meaning of the content. That's why I'm a bit reluctant to add additional information on the KpiMatrix.

However there is one place where the `period_id` is known (ie id of mis.report.instance.column, where you have date_from/date_to): it is in the [drilldown_arg](https://github.com/OCA/mis-builder/blob/515290d86c10fd02892420cb30dcd167fba0831d/mis_builder/models/mis_report.py#L789-L792) that is [attached to the cell](https://github.com/OCA/mis-builder/blob/515290d86c10fd02892420cb30dcd167fba0831d/mis_builder/models/kpimatrix.py#L139).

Do you think you could use that for your use case?

mis_builder/models/kpimatrix.py
```diff
@@ -486,3 +488,35 @@ def as_dict(self):
             'header': header,
             'body': body,
         }
+
+    def as_value(self):
+        values = {}
+        columns = {col.key: {'label': col.label,
+                             'date_from': col.date_from,
+                             'date_to': col.date_to} for col in self.iter_cols()}
```
sbidoul, line 496: the subcols should be taken into account here

sbidoul: could this `as_value` method be implemented as an option to `as_dict()` to return raw values instead of the col_data rendering we have now?

robyf70:
Hi @sbidoul,

I agree totally with you. Regarding date_from/to ranges, this seems stored as locals_dict, right?

Regarding period_id, yes I've got it! But also I've found that when you have calculated columns you have to find in different way, this value, see my patch. My goal is to get a dictionary value of rows associated with a list of dictionaries representing columns details.

mis_builder/models/kpimatrix.py
robyf70: yes! A very minimal return value from `as_dict()` would be nice, maybe as new key `raw_value` having the representation returned by my `as_value()` along `header` and `body` keys?

Issue #91, opened 2018-05-07
Cash Flow Forecast


jjscarafia:
Anyone working or planning a cash flow forecast using mis_builder?

If I'm no wrong it would be great to allow account.move.line analysis via de date_maturity field. It could be solved by mis_bilder queries by I think it would be better (and easier) to allow using date_maturity directly with some parameter on kpis.

@sbidoul any thoughts to share?

BR!

sbidoul:
Interesting idea. Could you elaborate a little bit on the kind of report you have in mind?

gonzaloruzafa:
Hi @sbidoul! The main goal would be to give visibility to forecasted funds flowing in or out of the company. Using the date_maturity of the account move lines (usually Payable or Receivable Accounts), you could project when that account is expected to become liquid cash (going in or out). With that info (you can use any account), you can project your future financial situation and take decisions.

**Example:** - Today: 07/05/2018

_Current Bank Balance: 500k_
* Payables move lines
_101200 Account Payable - 100k - Date maturity 15/05/2018_
_101200 Account Payable - 50k - Date maturity 21/05/2018_

* Receivable move lines
_101200 Account Receivable - 20k - Date maturity 15/05/2018_
_101200 Account Receivable - 800k - Date maturity 30/05/2018_

* Draft account.move.line for hipotetical in/out flow
_xxxxxx Other expected in  - 10k - Date maturity 15/05/2018_
_yyyyyy Other expected Out - 5k  - Date maturity 30/05/2018_

![seleccion_275](https://user-images.githubusercontent.com/3332717/39722487-4ef2e6a0-5219-11e8-8201-cb8bf198b1e0.png)

jjscarafia:
@sbidoul exactly, the idea is more less the one explained by @gonzaloruzafa 
Just to add a new boolean "use_date_maturity" kpis so that move lines are grouped by that field instead of date field.

Perhaps another great addition would be to think on some model (similar to budget) to add some forecast in/out flows. But this could also be achieved using a draft account.move with many account.move.lines representing that forecast in/out flows

sbidoul:
Thanks for the example. 

At first glance, adding "use date maturity" on kpis does not fit very well with the current model and algorithms.

What you can do today, without changing mis_builder is create a view model such as

```sql
SELECT
  date_maturity AS date
  company_id,
  account_id,
  debit, 
  credit
FROM account_move_line
WHERE ... account type in payable, receivable ...
```

[...]

PR #112, opened 2018-08-08
adding date field choise on instance and instance periods (draft, proposal)

jackjack82:
The goal is to have a report based on different date field on the move line.
The date field in on instance periods, in order to enable comparison between accounting/cash flow date in different columns.

The date field is also available on the report instance with "apply" button, in order to apply the date directly to all periods.

image

Improvements/open points:
- rename new field (e.g. date_field_name)?
- what if the date field is empty in destination model? Set a "undefined" line/field or warning?
- remove date field selction from non-"actuals" columns
- update automatic tests
- test flow adding new object to the MIS builder (other than account.move.line)

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -222,7 +222,8 @@ class AccountingExpressionProcessor(object):
     def get_aml_domain_for_expr(self, expr,
                                 date_from, date_to,
                                 target_move,
-                                account_id=None):
+                                account_id=None,
+                                date_field='date'):
         """ Get a domain on account.move.line for an expression.
 
         Prerequisite: done_parsing() must have been invoked.
@@ -252,7 +253,8 @@ class AccountingExpressionProcessor(object):
             if mode not in date_domain_by_mode:
                 date_domain_by_mode[mode] = \
                     self.get_aml_domain_for_dates(date_from, date_to,
-                                                  mode, target_move)
+                                                  mode, target_move,
+                                                  date_field)
         assert aml_domains
         # TODO we could do this for more precision:
         #      AND(OR(aml_domains[mode]), date_domain[mode]) for each mode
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Isn't better to create an SQL view from account_move_line with the date field you want and use that source? I think this complicates things without a real benefit.

jackjack82:
This was a request from a client and there is an open issue https://github.com/OCA/mis-builder/issues/91 . We may add additional dates on move lines, e.g. payment date. 

This can also be partially solved using existing queries. If not interested you can close the PR and the issue. It was a proposal, we have no final reports or case use yet.

Thank you, 
Giacomo

pedrobaeza:
Let's wait for @sbidoul's opinion. He's on holidays, so this might delay a bit, but if there's no hurry, it's OK to wait.

sbidoul:
@jackjack82 thanks for this PR!

My first reaction is this adds quite a bit of complexity for something that can be done today with a SQL view such as the one shown in https://github.com/OCA/mis-builder/issues/91#issuecomment-387402890 (albeit with some development).

However, I'm curious about your customer's use case. Maybe by collecting those use case such as #91 an interesting solution will emerge.

jjscarafia:
thanks @jackjack82 this is quite near to what we where looking for on https://github.com/OCA/mis-builder/issues/91.

I understand the caution of @sbidoul and @pedrobaeza . For me one use case would be for cash flow exactly as shown on https://github.com/OCA/mis-builder/issues/91. I believe is not much complexity to be added but I agree that another use case would be needed to think on a change like this.

Regarding the possibility of an SQL view from account_move_line, it's a good idea but obviously it's much easier just to be able to make a report based on the due_date. I believe being able to use due_date on mis reports would extend possibilities, not just for cash flow Forecast.

In summary, I agree with @sbidoul on collecting more use cases

Issue #73, opened 2018-03-11
Control display of from/to dates in column headers


sbidoul:
Displaying date from-to in the column header is not always useful. For columns with relative periods it helps but for columns named manually by the user, it is often useless.

This display could be controlled by a flag on `mis.report.instance.period`.

PR #117, opened 2018-08-24
[11.0][IMP] mis_builder: description of column disabled by default

astirpe:
Our customers seem to be a little annoyed by the description in the columns of the printed reports. The description is in facts most of the time redundant (the column name would be enough) and makes the look&feel of the report bloated.

With this PR, the description in the columns of the MIS reports will be not printed by default.
To re-enable it, set the flag *Display Columns Description* to true in the MIS report instance.

### DIFF
--- a/mis_builder/__manifest__.py
+++ b/mis_builder/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'MIS Builder',
-    'version': '11.0.3.2.2',
+    'version': '11.0.3.3.0',
     'category': 'Reporting',
     'summary': """
         Build 'Management Information System' Reports and Dashboards
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -424,6 +424,9 @@ class MisReportInstance(models.Model):
     no_auto_expand_accounts = fields.Boolean(
         string='Disable account details expansion',
     )
+    display_columns_description = fields.Boolean(
+        help="Display the date range details in the columns of the reports.",
+    )
     comparison_mode = fields.Boolean(
         compute="_compute_comparison_mode",
         inverse="_inverse_comparison_mode")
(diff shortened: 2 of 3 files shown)


elicoidal (approved):
LGTM

fclementic2c:
good idea

mis_builder/models/mis_report_instance.py
```diff
@@ -592,6 +595,10 @@ def _add_column_actuals(
             raise UserError(_("Column %s with actuals source "
                               "must have from/to dates.") %
                             (period.name,))
+
+        if not self.display_columns_description:
+            description = ''
```
sbidoul, line 600: I'd rather do this in _compute_matrix below so it applies to all column types

mis_builder/models/mis_report_instance.py
```diff
@@ -424,6 +424,9 @@ def _default_company_id(self):
     no_auto_expand_accounts = fields.Boolean(
         string='Disable account details expansion',
     )
+    display_columns_description = fields.Boolean(
+        help="Display the date range details in the columns of the reports.",
```
sbidoul, line 428: in the column headers?

mis_builder/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     'name': 'MIS Builder',
-    'version': '11.0.3.2.2',
+    'version': '11.0.3.3.0',
```
sbidoul, line 6: I'll manage the version bump, no need to do it in the PR.

sbidoul (commented):
Thanks! This actually implements #73.

Overall looks good to me. Just a couple of minor comments.

My initial idea was to control that per column, but it's good to do it at the report level. We can still add finer grained control later if needed.

If possible for you, can you do the PR towards 10.0? I'll then forward and backport to 11 and 9.

sbidoul:
Can you also add your name in contributors.

And, icing on the cake, add a file readme/newsfragments/73.feature with the changelog entry (I'm toying with towncrier to manage the changelog). (I can do this myself of course)

astirpe:
@sbidoul thank you! I opened a separate PR https://github.com/OCA/mis-builder/pull/118

sbidoul:
Superseded by #118

PR #118, opened 2018-08-27
[IMP] Description of column disabled by default

astirpe:
With this PR, the description in the columns of the MIS reports will be not printed by default.
To re-enable it, set the flag Display Columns Description to true in the MIS report instance.

Implements #73.
Supersedes https://github.com/OCA/mis-builder/pull/117.


sbidoul (approved):
Looks good. This changes the default behaviour but this is a better default for the majority of reports, and it is easy to adapt by the end user, so ok. Thanks!

yung-wang (approved):
Looking good

pedrobaeza (approved):

sbidoul:
Thanks! I'll port to 9 and 11 soon.

PR #123, opened 2018-09-10
[IMP] add evaluate method to mis.report

sbidoul:
Add new simplified method to evaluate a mis report over a time period.


benwillig (approved):

robinkeunen (approved):
The diff looks good to me and I tested locally but I didn't dig deep into the underlying logic (ie `prepare_kpi_matrix` and `declare_and_compute_period`)

PR #33, opened 2017-10-12
Version 3.1

sbidoul:
This is the integration branch that will become version 3.1. 

Everything here has been reviewed in [separate PR's](https://github.com/OCA/mis-builder/milestone/2?closed=1).

See [changelog](https://github.com/OCA/mis-builder/pull/33/files#diff-45868279802e9dac0162d9042ab7cf83).

PR #24, opened 2017-10-02
analytic filtering mechanism in mis_builder (was mis_builder_analytic_filter)

adrienpeiffer:
Include https://github.com/OCA/account-financial-reporting/pull/243 into mis_builder module

closes #15
closes #42 
closes #44

PR #134, opened 2018-10-23
WIP: [10.0] - Combined analytical axes report 

sbejaoui:
This PR adds a wizard to evaluate one mis-builder template with all possible combinations of given analytical axes

based on:

- #33 
- #24  
- #50  
- #51


sbidoul:
@sbejaoui as this is going to be a slow report, I don't want this in mis builder core. Please make it a separate module. Also I think this is not generic enough for OCA.

PR #143, opened 2018-11-28
wip refactor generic JS

mpanarin:
This is a refactor of mis-builder widget to follow a v11 API, as well as being a generic view. This will eventually supersed the https://github.com/OCA/mis-builder/pull/133


sbidoul:
Oh. So it's not a widget anymore?

We liked to have it a widget so it could be in principle included in any form.

Could we keep it close to the existing design for v11 (and just adding the analytic filter), and discuss a possible refactor for v12?

sbidoul:
Also note, if you work on the v11 port of the analytic filters: I strive to keep the code base very similar across version so I can git merge changes from one to the other.

So to do 11: do a git merge of 7b03a3b20be06c137c10444cb00feea16112a8aa in 11.0 and manually fix the widget and the get_pdf method, please do no other changes. Then I'll finish (changelog) and release 11.0.3.3.0.

simahawk:
@sbidoul if we split the rendering you can do it where you want... a widget, a view, etc :wink:

sbidoul:
@simahawk sounds interesting indeed. But please let's keep a straitghtforward port so we can release 11.0.3.3.0 with minimum fuss, then we can talk big refactoring.

Also, if we do a big refactoring like this I'd like to make sure it works identically in 12 before committing it in 11, to ease maintenance afterwards.

benwillig:
As @sbidoul said, I think it's better to keep a widget to render the view. The ```MisReportRenderer``` would simply instanciate it and destroy it when switching from a record to another one. I don't think it would be a lot of work to change it. I didn't work a lot with the v11 framework, but in my opinion, it would be wrong to instanciate a view type renderer in a field widget for example.

simahawk:
@benwillig nope, is not wrong... is the purpose of splitting the concerns to different components and reuse them.
Indeed, is what the FieldX2Many widget does: reuse the list view renderer :wink:

benwillig:
@simahawk I would say it's because we need to get the form and list views of the related model. So there is an explicit call of the FormRenderer and TreeRenderer in x2many widgets?

simahawk:
Yes, but that's the point: you have something that is able to render something else: you can reuse it.
This is the example:

https://github.com/odoo/odoo/blob/f5c3dafb045f4eb5e832c45596dd59396667ea09/addons/web/static/src/js/fields/relational_fields.js#L956
https://github.com/odoo/odoo/blob/f5c3dafb045f4eb5e832c45596dd59396667ea09/addons/web/static/src/js/fields/relational_fields.js#L1014

And this is how we did it for the matrix widget (a widget, not a view):

https://github.com/OCA/web/blob/11.0/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js#L181

Here is used for the widget and the widget is only responsible to prepare the data and communicate w/ the field. So, if we want to make this matrix a view we can w/out rewriting the whole rendering.

benwillig:
Ok thanks for the information. I guess that when we instanciate thoses renderers in views, only datas are reloaded when changing to another record, right? If it's the case, I don't think it fits with a mis report instance, because when displaying a MIS report, you have to reload everything, view, data, filters, and options.

I understand that it is better to split. That's why I think if we had a simple class which simply extends the Widget class, we could reuse it in the renderer. And it would work the same way as it does now. I 
 didn't check, but what part of the BasicRenderer is really needed to display a mis report? Actually, in 11.0 branch, the widget extends the AbstractField class, it shouldn't. It should extends the Widget class. And we would use that class in the field widget.

sbidoul:
Closing this for now since #229 seems to be on a good track. Thanks for the effort here. Maybe some ideas can be revived in the future when migrating to the OWL...

PR #148, opened 2018-12-21
mis_builder_analytic: New module

ernestotejeda:
Depends on:

- [x] https://github.com/OCA/mis-builder/pull/149

Cc @Tecnativa
TT14942

PR #149, opened 2019-01-09
Configurable account.account model (can use any model that has a code field)

ernestotejeda:
The account_id field of the model selected in 'Move lines source'
in the Period form (mis.report.instance.period) can be a Many2one
relationship with any class that has the field code

Cc @Tecnativa
TT14942

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -89,7 +89,7 @@ class AccountingExpressionProcessor(object):
         r"(?P<ml_domain>\[.*?\])?"
     )
 
-    def __init__(self, companies, currency=None):
+    def __init__(self, companies, currency=None, account_model=None):
         self.env = companies.env
         self.companies = companies
         if not currency:
@@ -112,6 +112,9 @@ class AccountingExpressionProcessor(object):
         # a first query to get the initial balance and another
         # to get the variation, so it's a bit slower
         self.smart_end = True
+        # Account model
+        self._account_model = self.env[
+            account_model or 'account.account'].with_context(active_test=False)
 
     def _account_codes_to_domain(self, account_codes):
         """Convert a comma separated list of account codes
@@ -188,8 +191,6 @@ class AccountingExpressionProcessor(object):
 
     def done_parsing(self):
         """ Replace account domains by account ids in map """
-        account_model = self.env['account.account'].\
-            with_context(active_test=False)
         for key, acc_domains in self._map_account_ids.items():
             all_account_ids = set()
             for acc_domain in acc_domains:
@@ -197,7 +198,8 @@ class AccountingExpressionProcessor(object):
                     acc_domain,
                     [('company_id', 'in', self.companies.ids)],
                 ])
-                account_ids = account_model.\
+                account_ids = self._account_model.\
+                    with_context(active_test=False).\
                     search(acc_domain_with_company).ids
                 self._account_ids_by_acc_domain[acc_domain].\
                     update(account_ids)
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -147,12 +147,12 @@ class KpiMatrixCell(object):
 
 class KpiMatrix(object):
 
-    def __init__(self, env):
+    def __init__(self, env, account_model=None):
         # cache language id for faster rendering
         lang_model = env['res.lang']
         self.lang = lang_model._lang_get(env.user.lang)
         self._style_model = env['mis.report.style']
-        self._account_model = env['account.account']
+        self._account_model = env[account_model or 'account.account']
         # data structures
         # { kpi: KpiMatrixRow }
         self._kpi_rows = OrderedDict()
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -455,6 +455,10 @@ class MisReport(models.Model):
     _name = 'mis.report'
     _description = 'MIS Report Template'
 
+    def _default_move_lines_source(self):
+        return self.env['ir.model'].search(
+            [('model', '=', 'account.move.line')]).id
+
     name = fields.Char(required=True,
                        string='Name', translate=True)
     description = fields.Char(required=False,
@@ -470,6 +474,30 @@ class MisReport(models.Model):
     subkpi_ids = fields.One2many('mis.report.subkpi', 'report_id',
                                  string="Sub KPI",
                                  copy=True)
+    move_lines_source = fields.Many2one(
+        comodel_name='ir.model',
+        string='Move lines source',
+        domain=[('field_id.name', '=', 'debit'),
+                ('field_id.name', '=', 'credit'),
+                ('field_id.name', '=', 'account_id'),
+                ('field_id.name', '=', 'date'),
+                ('field_id.name', '=', 'company_id')],
+        default=_default_move_lines_source,
+        required=True,
+        help="A 'move line like' model, ie having at least debit, credit, "
+             "date, account_id and company_id fields.",
+    )
+    account_model = fields.Char(
+        compute='_compute_account_model',
+        string='Account model',
+    )
+
+    @api.multi
+    @api.depends('move_lines_source')
+    def _compute_account_model(self):
+        for record in self:
+            record.account_model = record.move_lines_source.field_id.filtered(
+                lambda r: r.name == 'account_id').relation
 
     @api.onchange('subkpi_ids')
     def _on_change_subkpi_ids(self):
@@ -536,7 +564,7 @@ class MisReport(models.Model):
     @api.multi
     def prepare_kpi_matrix(self):
         self.ensure_one()
-        kpi_matrix = KpiMatrix(self.env)
+        kpi_matrix = KpiMatrix(self.env, self.account_model)
         for kpi in self.kpi_ids:
             kpi_matrix.declare_kpi(kpi)
         return kpi_matrix
@@ -544,7 +572,7 @@ class MisReport(models.Model):
     @api.multi
     def _prepare_aep(self, companies, currency=None):
         self.ensure_one()
-        aep = AEP(companies, currency)
+        aep = AEP(companies, currency, self.account_model)
         for kpi in self.kpi_ids:
             for expression in kpi.expression_ids:
                 if expression.name:
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -284,6 +284,20 @@ class MisReportInstancePeriod(models.Model):
          'Period name should be unique by report'),
     ]
 
+    @api.constrains('source_aml_model_id')
+    def _check_source_aml_model_id(self):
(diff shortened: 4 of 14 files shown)


pedrobaeza:
@sbidoul this is needed for #148

sbidoul:
Interesting. I'll look into it within a couple of weeks.

mis_builder/models/mis_report_instance.py
```diff
@@ -669,10 +669,19 @@ def _compute_matrix(self):
...
+
+            account_model = None
+            if period.source == 'actuals_alt' and period.source_aml_model_id:
+                account_field = period.source_aml_model_id.field_id.filtered(
+                    lambda r: r.name == 'account_id')
+                account_model = account_field.relation
+
+            aep = self.report_id._prepare_aep(self.query_company_ids,
+                                              self.currency_id,
+                                              account_model=account_model)
```
sbidoul: Looking at this in more details, I think it does not make sense to have a different account_model for different columns, because of account details expansion which would become meaningless if different account objects are used in different columns.

sbidoul (changes requested):
This require more thinking.

mis_builder/models/mis_report_instance.py
sbidoul: IOW, this may work, but the account model must be configured globally at the level of mis.report, and passed to the AEP constructor, as well as to the KPI matrix constructor.

sbidoul: Or, better, define the AML model used for Actuals at the mis.report level, and make sure Actual (alternative) models used in columns have the same account model in the account_id m2o.

**updated**: mis.report instead of mis.report.instance

pedrobaeza:
@sbidoul are you saying that each time you want to create an instance, you have to define all of these things? I think all this information should be in the template, not the instance.

sbidoul:
@pedrobaeza yes, correct, I meant the template.

pedrobaeza:
Well, that can an option, but why complicate it when it can be auto-inferred? I don't get the advantage.

sbidoul:
@pedrobaeza I don't understand your last comment.

pedrobaeza:
As we have the model of `account_id` field, we already know it, so why re-define it on template level?

sbidoul:
That's not what I suggest in https://github.com/OCA/mis-builder/pull/149#discussion_r247365221

What I suggest is
* add a field named "Actuals move lines model" on `mis.report` (default = `acccount.move.line`)
* from that, the account_id model can be found
* in mis.report.instance.period configuration, enforce the account_id field used for the "Actuals (alternative)" models uses the same account_id model as the main one

pedrobaeza:
But this is what I'm saying that is not really needed, as you already know the m2o model from the model definition. Why do you need to enforce that thing? It would make sense if you enforce all things, like using a template with an specific actual data, but only for the model...

Other thing that I don't understand is that the data source is specified at instance level instead of template. Maybe a template can work with several data sources, but this should be specified by a m2m, but not having to be configured each time for each report...

sbidoul:
The data source is specified at the *column* level in the instance. There are no columns in the template. This allows having a column for actuals, another for committed, and a third for budget for example.

I'm just saying that we must enforce the same account model is used for the data sources used in all columns. This is necessary because when you expand accounts they are displayed in rows and the rows must make sense for all columns.

mis_builder/models/mis_report_instance.py
sbidoul: Another way to say it is that one way or another, the same AEP must be used for all columns and the account model must therefore be known before the `for period` loop.

pedrobaeza:
@ernestotejeda can you do the requested changes or is there something you don't understand?

ernestotejeda:
Yes @pedrobaeza, I can try to do the requested changes according to https://github.com/OCA/mis-builder/pull/149#issuecomment-453938499. I agree with @sbidoul, thanks.

ernestotejeda:
@sbidoul i have done the changes proposed

rafaelbn:
Hi @ernestotejeda This branch has conflicts that must be resolved, could you rebase?

@sbidoul could you give a feedback of the last changes?

mis_builder/models/aep.py
```diff
@@ -112,6 +112,10 @@ def __init__(self, companies, currency=None):
         # a first query to get the initial balance and another
         # to get the variation, so it's a bit slower
         self.smart_end = True
+        # Account model
+        self._account_model = account_model
+        if account_model is None:
+            self._account_model = self.env['account.account']
```
sbidoul:
```suggestion
        self._account_model = self.env[account_model or 'account.account'].with_context(active_test=False)
```

mis_builder/models/aep.py
```diff
@@ -188,16 +192,15 @@ def parse_expr(self, expr):
...
         for key, acc_domains in self._map_account_ids.items():
             all_account_ids = set()
             for acc_domain in acc_domains:
                 acc_domain_with_company = expression.AND([
                     acc_domain,
                     [('company_id', 'in', self.companies.ids)],
                 ])
-                account_ids = account_model.\
+                account_ids = self._account_model.\
+                    with_context(active_test=False).\
```
sbidoul, line 202:
```suggestion
```

mis_builder/models/kpimatrix.py
```diff
@@ -147,12 +147,14 @@ def __init__(self, row, subcol,
...
-    def __init__(self, env):
+    def __init__(self, env, account_model=None):
         # cache language id for faster rendering
         lang_model = env['res.lang']
         self.lang = lang_model._lang_get(env.user.lang)
         self._style_model = env['mis.report.style']
-        self._account_model = env['account.account']
+        self._account_model = account_model
+        if account_model is None:
+            self._account_model = env['account.account']
```
sbidoul:
```suggestion
        self._account_model = env[account_model or 'account.account']
```

mis_builder/models/mis_report.py
```diff
@@ -454,6 +458,33 @@ class MisReport(models.Model):
...
+        default=_default_move_lines_source,
+        required=True,
+        help="A 'move line like' model, ie having at least debit, credit, "
+             "date, account_id and company_id fields.",
+    )
+    account_model = fields.Many2one(
+        compute='_compute_account_model',
+        comodel_name='ir.model',
+        string='Account model',
+    )
```
sbidoul, line 493: make this field a `Char`

mis_builder/models/mis_report.py
```diff
@@ -520,15 +551,15 @@ def copy(self, default=None):
     @api.multi
     def prepare_kpi_matrix(self):
         self.ensure_one()
-        kpi_matrix = KpiMatrix(self.env)
+        kpi_matrix = KpiMatrix(self.env, self.env[self.account_model.model])
```
sbidoul:
```suggestion
        kpi_matrix = KpiMatrix(self.env, self.account_model)
```

mis_builder/models/mis_report.py
```diff
@@ -520,15 +551,15 @@ def copy(self, default=None):
...
         for kpi in self.kpi_ids:
             kpi_matrix.declare_kpi(kpi)
         return kpi_matrix
 
     @api.multi
     def _prepare_aep(self, companies, currency=None):
         self.ensure_one()
-        aep = AEP(companies, currency)
+        aep = AEP(companies, currency, self.env[self.account_model.model])
```
sbidoul:
```suggestion
        aep = AEP(companies, currency, self.account_model)
```

mis_builder/models/mis_report_instance.py
```diff
@@ -268,6 +268,20 @@ def _compute_dates(self):
...
     ]
 
+    @api.constrains('source_aml_model_id')
+    def _check_source_aml_model_id(self):
+        for record in self:
+            if record.source_aml_model_id:
+                record_model = record.source_aml_model_id.field_id.filtered(
+                    lambda r: r.name == 'account_id').relation
+                period_model = record.report_id.account_model.model
+                if record_model != period_model:
```
sbidoul: s/period_model/report_account_model/
s/record_model/period_account_model/

mis_builder/models/mis_report_instance.py
```diff
@@ -703,7 +718,8 @@ def drilldown(self, arg):
         account_id = arg.get('account_id')
         if period_id and expr and AEP.has_account_var(expr):
             period = self.env['mis.report.instance.period'].browse(period_id)
-            aep = AEP(self.query_company_ids, self.currency_id)
+            aep = AEP(self.query_company_ids, self.currency_id,
+                      self.env[self.report_id.account_model.model])
```
```suggestion
                      self.report_id.account_model)
```

mis_builder/tests/test_data_sources.py
```diff
@@ -85,7 +85,10 @@ def setUp(self):
         # create report
         self.report = self.env['mis.report'].create(dict(
             name='test report',
+            move_lines_source=self.env['ir.model'].search(
+                [('model', '=', 'account.move.line')]).id,
         ))
+        self.report._compute_account_model()
```
sbidoul: It should not be necessary to change the test since `move_lines_source` has a default value.

mis_builder_demo/data/mis_report.xml
```diff
@@ -6,6 +6,8 @@
   <record id="mis_report_expenses" model="mis.report">
     <field name="name">Demo Expenses</field>
     <field name="style_id" ref="mis_report_expenses_style1"/>
+    <field name="move_lines_source" search="[('model', '=', 'account.move.line')]"/>
+    <field name="account_model" search="[('model', '=', 'account.account')]"/>
```
sbidoul: This should not be necessary since there is a default value.

sbidoul (commented):
@ernestotejeda @rafaelbn This is going into the right direction. I made a few relatively minor comments.

Could you make this PR to 10.0? I'll then forward port quickly as soon as it is merged.
Also could you add a file named `readme/newsfragments/<pr or issue number>.feature` that provides a short description of the feature that will be included automatically in the changelog?

Thanks for this work.

rafaelbn:
Hi @ernestotejeda please rebase, @pedrobaeza @sbidoul could we discuss this? 

https://github.com/OCA/mis-builder/pull/148 depends on this

Thanks

ernestotejeda:
@rafaelbn, @sbidoul , @pedrobaeza I have done the changes proposed by @sbidoul

sbidoul:
Thanks, I'll review soon.

sbidoul:
@ernestotejeda I ported your commit to 10.0 and slightly amended it to resolve the conflict that was due to #242. It is in https://github.com/OCA/mis-builder/pull/240 (ce736132d01c31053ed49192fd2e165f9857d0e0). Would you like to have quick look at it?

I removed the changes to the test, because the change should be fully backward compatible. 

I also changed the way you added contributors, to conserve the style of adding contributors chronologically, I hope you don't mind.

sbidoul:
Closing as it is included in 10.0-staging and will be soon merged in 11.0. Shout if you see any issue.

PR #150, opened 2019-01-13
Forward port 10.0 to 11.0 [DO NOT MERGE]

sbidoul:
This contains everything from 10.0 except the js part of the analytic filters.


pedrobaeza:
@sbidoul how is the work to migrate to 11.0/12.0 going? This is starting to be a bit urgent as this is the base for our localizations as you know. Do you need help in something?

sbidoul:
@pedrobaeza this PR as well as #151 (for porting to 12) are working fine. They both contain exactly the same code as 10.0, with the notable exception of the analytic filter in the preview widget. *This is not blocking*, as all the other feature work fine without that.

So as soon as someone else than me does a quick functional test on this PR and #151, I'll push them to the main branches.

Now about migrating the analytic filters in the widget, this is harder than I initially thought. @simahawk showed me that the changes in the Odoo js Field Manager that make that difficult. There must be a workaround but it's still elusive.

elicoidal:
something to investigate: I just logged in the [runbot](http://3356308-11-0-3-3-c99493.runbot3.odoo-community.org) and MIS Builder is installed but not the Invoicing menu visible so there is no way to access by default the menu. 
It looks like a dependency is missing?

elicoidal:
I had to install manually `account_invoicing`

elicoidal:
I added admin to `full accounting feature` group

image

Current: When click on print button form view is open
Expected: the report is printed

I did some other small tests (add lines+total) and didnot see any other issue

sbidoul:
@elicoidal about the groups, yes, there is only a dependency on `account`. I like to leave it this way to not force unneeded dependencies (in some circumstances the engine is useful without the menus, and in practice the missing menus issue only occurs on runbot, as people normally install invoicing or accounting anyway).

About the print button, could it be you clicked slightly off the icon center?

Issue #102, opened 2018-06-27
MIS Budget 


fclementic2c:
Hi @sbidoul 

I fear a big limitation on MIS budget. Tell me if I got something wrong here.

Context:
- A report template can be used by different companies ok... but only if based on same chart of account. (I know that technically it works but we will have funny numbers if your 3% for instance point to a different CoA. Right?)
- A MIS budget can and must have only 1 report template. 

use case: 
A group has companies in France, Switzerland and Germany with 3 different chart of accounts. 
Having 3 different P&L report template is not an issue.
But how can you handle the budget? 
If you create 3 budget (1 for each country P&L template) you cannot go any further to aggregate budgets.
Don't we have an issue on MIS Builder budget assuming that a budget must be linked to only 1 template ?
Even in mono-company environment, it means that you can see your budget through only 1 report. A department manager cannot have his own department report then.

In my view, we miss 1 model between the kpi and the budget item (let's call it Budget account) 
We would need to be able to map a KPI  to a Budget item account (this budget account would be shared cross company).

What's your opinion Stéphane ? Do you agree ? Does it sounds technically do-able ? and, If yes, would it be a big job ?

Thanks

Frederic

fclementic2c:
just thinking further:
that would suppose that instead of a flag on line expressions you would have to choose a budget account.
Or (but much more complex) we would need possibility to build report template based on budget accounts.

sbidoul:
Not looked in details to your question yet, but a first hint:

> only if based on same chart of account

This is true only if you base the report template on account codes.

You can also create reports based on, say, account tags, eg `bal[('tag_ids...', '=', '...')]`.

fclementic2c:
@sbidoul Yes, but I recently discovered that if you base you KPI on tags you cannot use subkpi based on operating units or analytic tags :|

sbidoul:
> if you base you KPI on tags you cannot use subkpi based on operating units or analytic tags

That needs to be investigated. Can you open a separate issue about that?

For the rest the fundamental philosophy of mis budgets is that we budget on KPIs (+ possibly analytic dimensions). It is useful for instance if you want to budget on non accounting data (eg budget on a number of employees, or similar). I'm not too sure how a concept of budget account would fit into that.

If you want to do consolidation with different chart of accounts, using account tags or types is the way to go.

[...]

PR #153, opened 2019-01-23
[WIP] Working with External Reports' KPIs

hugouchoasborges:
- **[ADD] MisReport Code Field Added**

Here was added a new field, **code**, which contains the report's name without any special characters and/or spaces.
The method called to do that was __python_var(var_str)_, as is done with the _name_ field from _kpi_ids_.

- **[IMP] MIS Reports working with external Report's KPIs**

This commit allows the user entering a 'dot notation' expression for a KPI. i.e, _report_01.receipt_05_.
Doing so, MIS Reports will compute the value held by the "receipt_05 KPI" from the "report_01 MisReport".

### DIFF
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -769,9 +769,6 @@ class MisReportInstancePeriod(models.Model):
             ('report_id', 'in', sub_report_ids)
         ])
 
-        id_tree = self.env.ref('mis_builder.'
-                               'mis_report_instance_view_tree')
-
         id_form = self.env.ref('mis_builder.'
                                'mis_report_instance_result_view_form')
 
@@ -783,7 +780,7 @@ class MisReportInstancePeriod(models.Model):
                 'view_type': 'list',
                 'views': [(False, 'list'), (False, 'form')],
                 'view_id': False,
-                'domain': str([('id', 'in', sub_report_ids)]),
+                'domain': str([('id', 'in', report_ids.ids)]),
                 'target': 'new',
                 'context': context,
             }
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -426,8 +426,6 @@ class MisReport(models.Model):
         self.ensure_one()
         res = {}
 
-        recomputed_dict = {}
-
         localdict = {
             'registry': self.pool,
             'sum': _sum,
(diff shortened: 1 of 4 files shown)


sbidoul:
Thanks for this very interesting contribution!

Allow me some time to check the design, I need to verify the concept will work nicely on version 9 and greater.

JordiBForgeFlow:
@sbidoul @HugoUchoasBorges as I understand this would mean, for example, that you could define a MIS Report for a "Balance Sheet", and then have another MIS Report "Balance Sheet KPI" that fetches kpi's from "Profit&Loss" to create new KPIS (e.g. Current Assets / Total Assets).

mis_builder/models/mis_builder.py
```diff
@@ -439,20 +450,36 @@ def _compute(self, lang_id, aep,
...
                 except ZeroDivisionError:
                     kpi_val = None
                     kpi_val_rendered = '#DIV/0'
                     kpi_val_comment += '\n\n%s' % (traceback.format_exc(),)
                 except (NameError, ValueError):
-                    recompute_queue.append(kpi)
+                    if '.' in kpi.expression:
+                        report_descript, kpi_name = kpi.expression.split('.')
+                        inherit = kpi.search(
+                            [('report_id.description', '=', report_descript),
```
mileo: report_id.code

mis_builder/models/mis_builder.py
```diff
@@ -489,6 +516,32 @@ def _compute(self, lang_id, aep,
...
+                    kpi_ids=inherit_record.kpi_ids,
+                )
+
+                report_descript, kpi_name = kpi.expression.split('.')
+
+                inherit = kpi.search(
+                    [('report_id.description', '=', report_descript),
+                     ('name', '=', kpi_name)]
+                )
+
```
mileo: DEAD code from line 525 to 530

mis_builder/models/mis_builder.py
```diff
@@ -498,6 +551,7 @@ def _compute(self, lang_id, aep,
                 # so we stop trying
                 break
             # try again
+
```
mileo: Remove this space

mileo (changes requested):

fclementic2c:
> @sbidoul @HugoUchoasBorges as I understand this would mean, for example, that you could define a MIS Report for a "Balance Sheet", and then have another MIS Report "Balance Sheet KPI" that fetches kpi's from "Profit&Loss" to create new KPIS (e.g. Current Assets / Total Assets).

if I understand like @bjeficent, this is the fix for the bigger weakness of MIS builder for finance reports !

sbidoul:
@jbeficent that's similar to what we discussed at the sprint indeed.

@fclementic2c if that's the bigger weakness for you, you should have created an enhancement issue to describe the requirement :)

fclementic2c:
@sbidoul I already did : https://github.com/OCA/mis-builder/issues/102

mis_builder/models/mis_builder.py
```diff
@@ -439,20 +450,36 @@ def _compute(self, lang_id, aep,
...
                     localdict[kpi.name] = kpi_val
                 except ZeroDivisionError:
                     kpi_val = None
                     kpi_val_rendered = '#DIV/0'
                     kpi_val_comment += '\n\n%s' % (traceback.format_exc(),)
                 except (NameError, ValueError):
-                    recompute_queue.append(kpi)
+                    if '.' in kpi.expression:
+                        report_descript, kpi_name = kpi.expression.split('.')
```
sbidoul: @HugoUchoasBorges this `split` is not going to work with complex expressions. There maybe many `.` in the expression.

The design I had in mind when I thought about that last year was something like this:
* add on `mis.report` a m2m with a list of dependent `mis.report`
* the code you add on `mis.report` looks like a great idea
* then at the beginning of _compute (before the recompute loop), evaluate the dependent reports and add their localdict as some sort of dataclass in the localdict of the main report with the dependent report code as key name

In v9+ there is an `evaluate` method on `mis.report` that simplifies the evaluation of dependent report a little bit. You can probably get inspiration there.

Anyway, this PR is cool!

sbidoul:
@fclementic2c #102 sounds like something else to me. We closed it BTW, was there a feature request in there that I missed?

I created #155 to track this feature.

mileo:
Video of the current progress: https://drive.google.com/open?id=19m75wOzJ5VMxMKQJDObAekpy1UfA-mVn

Last step: Fetch the context after the redirect

cc: @HugoUchoasBorges

mileo:
New video: https://drive.google.com/drive/folders/1bVDdDLSG3x7EShHxYiOgL5rhO-AOUTFV

mis_builder/models/mis_builder.py
```diff
@@ -800,7 +936,21 @@ def display_settings(self):
     def compute(self):
         self.ensure_one()
 
-        aep = self.report_id._prepare_aep(self.root_account)
+        sub_report_ids = self.env.context.get('sub_report_ids')
+
+        if sub_report_ids:
+            report_id = self.env['mis.report'].browse(sub_report_ids)
```
mileo: This isn't working.

mileo (changes requested):

mis_builder/models/mis_builder.py
hugouchoasborges: https://github.com/OCA/mis-builder/pull/153/commits/8afb3c68b4d6d4755b7a6c798602be89b69574ff

mis_builder/static/src/js/mis_builder.js
```diff
@@ -78,7 +78,13 @@ openerp.mis_builder = function(instance) {
         },
         generate_content: function() {
             var self = this
-            context = new instance.web.CompoundContext(self.build_context(), self.get_context()|| {}) 
+
+            old_context = self.field_manager.ViewManager.ActionManager.inner_action.context;
```
mileo: In my opinion you'd better refactor the "getcontext" to get inner action context then append mis_report_instance_id if need. 

This way the context will be correct for any new modifications that need it.

mileo (changes requested):

mis_builder/static/src/js/mis_builder.js
hugouchoasborges: https://github.com/OCA/mis-builder/pull/153/commits/154dd3c9536420e829d0ae80aec3f98c29ba88d5

mileo:
Hi, we plan to port v8 feature to v12.

Can we follow the same approach? Any suggestion?

Thanks in advance

sbidoul:
@mileo I'm planning to work on this in the next ocuple of weeks. The approach I have in mind is simpler, but I need to check a couple of things before. i'll keep you informed.

mileo:
> @mileo I'm planning to work on this in the next ocuple of weeks. The approach I have in mind is simpler, but I need to check a couple of things before. i'll keep you informed.

Great, we plan to work on it this week... We can assume that. If you can speak what you have in mind, we will be happy to implement it.

Issue #109, opened 2018-07-30
New style option: "page break before"


sbidoul:
Allow forcing a page break before a KPI.

Not sure yet if it should be a style option or a flag on the KPI.

hbrunn:
In https://github.com/OCA/account-financial-reporting/pull/1076 I now propose to introduce a `split_after` flag on KPIs that causes a new table to be rendered after a KPI with the flag set (why not in styles? Because then we can't share styles between non-splitting and splitting KPIs, and I think KPI is the right level of abstraction anyways).

Works quite nicely for my horizontal use case, but in this version I made it general enough that you can split your KPIs to any amount of tables. Page breaks should be a simple followup then.

Would you still prefer a patch on mis_builder, or should I start with `mis_builder_horizontal` to use as incubator for this feature? (I prefer that)

PR #156, opened 2019-01-25
Render table in one or more columns

gabrielcardoso21:
Sometimes it's necessary to show two or more tables alongside one another. The original module mis_builder didn't have this feature by default, so I added it. It was necessary, though, to change what used to be returned from the method compute from the model mis_report_instance, which can cause trouble if you are overriding this method.

### DIFF
--- a/mis_builder/models/mis_builder.py
+++ b/mis_builder/models/mis_builder.py
@@ -69,6 +69,8 @@ class MisReportKpi(models.Model):
     description = fields.Char(required=True,
                               string='Description',
                               translate=True)
+    column = fields.Integer(string='Column',
+                            default=1)
     expression = fields.Char(required=True,
                              string='Expression')
     default_css_style = fields.Char(string='Default CSS style')
@@ -818,18 +820,18 @@ class MisReportInstance(models.Model):
             kpi_values_by_period_ids[period.id] = kpi_values
 
         # prepare header and content
-        header = []
-        header.append({
+        header = [{
             'kpi_name': '',
             'cols': []
-        })
+        }]
         content = []
         rows_by_kpi_name = {}
         for kpi in self.report_id.kpi_ids:
             rows_by_kpi_name[kpi.name] = {
                 'kpi_name': kpi.description,
                 'cols': [],
-                'default_style': kpi.default_css_style
+                'default_style': kpi.default_css_style,
+                'column': kpi.column,
             }
             content.append(rows_by_kpi_name[kpi.name])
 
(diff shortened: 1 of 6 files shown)


mileo (approved):

mileo:
Video / PDF / XLS of my test:

https://drive.google.com/drive/folders/1qYPTPxoDhg-Ho8ac36f9Wcy8-1r20xai?usp=sharing

mileo:
@gabrielcardoso21 please fix travis errors: https://travis-ci.org/OCA/mis-builder/builds/484373470?utm_source=github_status&utm_medium=notification

sbidoul:
Interesting PR. Do you think you could make it with a concept of "page/column break" as hinted in https://github.com/OCA/mis-builder/issues/109 ?

gabrielcardoso21:
> Interesting PR. Do you think you could make it with a concept of "page/column break" as hinted in #109 ?

@sbidoul It's a great idea! I mean, it's not very easy to make a page break, but a column break configurable at the kpi would save much time in configuration. Mainly because the way I made it the user has to configure the column of the kpi lines one by one. I'll sure work on that, thanks!

mileo:
There is also another feature that is in our backlog to allow the display of this report:

image

**With the description only in the first column.**

www.investidorpetrobras.com.br/download/6111

gabrielcardoso21:
Something I forgot to mention, when you name you kpi like header_{number of the column}, the description of this kpi appears in the header of it's respective column. Like this example, where I named the kpi 'header_1' with description 'ATIVO' and other I named 'header_2' with description 'PASSIVO': 
captura de tela de 2019-01-25 14-39-47

mileo:
image

Source: http://www.abgf.gov.br/wp-content/uploads/2017/04/ABGF_DemonstracoesContabeis-2016-e-pareceres.pdf

lili94:
hi please how to make tow colomun with different expression please help

mileo:
@lili94 please see: https://drive.google.com/drive/folders/1qYPTPxoDhg-Ho8ac36f9Wcy8-1r20xai

lili94:
Exactly but in odoo v11 can you help me please

mileo:
Hi, we plan to port v8 feature to v12.

Can we follow the same approach? Any suggestion?

Thanks in advance

sbidoul:
Great!

Can you make it with a "Column break before" flag on the KPI. Also maybe with an explicit "Column title" field that appears when column break is ticked (I think it's easier to discover than a conventional kpi names such as `header_`). 

Can you please make the PR to 10.0? Forward porting to 11 and 12 (and 13) will then be very easy.

PR #158, opened 2019-02-07
[11.0][mis_builder] define the comparison method in the mis.report.instance.period

JordiBForgeFlow:
This PR adds a field `compare_method` in the `mis.report.instance.period`, allowing to define multiple types of comparison (by percentage and by difference) simultaneously for the same KPI.

image

image

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -186,13 +186,13 @@ class KpiMatrix(object):
         return col
 
     def declare_comparison(self, cmpcol_key, col_key, base_col_key,
-                           label, description=None):
+                           label, description=None, compare_method=None):
         """ Declare a new comparison column.
 
         Invoke the declare_* methods in display order.
         """
         self._comparison_todo[cmpcol_key] = \
-            (col_key, base_col_key, label, description)
+            (col_key, base_col_key, label, description, compare_method)
         self._cols[cmpcol_key] = None  # reserve slot in insertion order
 
     def declare_sum(self, sumcol_key, col_to_sum_keys,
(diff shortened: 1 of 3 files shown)


sbidoul:
Interesting. The problem I see with this is that some comparison methods do not make sense for some KPIs.

For instance if a KPI is itself a percentage (like a profit margin), comparing with the Percentage method does not make sense. Some other kpis may be strings, where pct or diff comparison do not make sense either.

Not sure what to do about it. Maybe kpis should say which are their preferred and allowed comparison methods, and the period could override it if compatible.

This could be done in a second step.

sbidoul:
Don't forget to add a test. If possible can you do the PR on 10, and add a file name 158.feature in the `readme/newsfragments/` to populate the changelog? I'll then take care of forward/packporting.

JordiBForgeFlow:
@sbidoul OK, I'll proceed as you suggest.

But if a KPI is not suitable for comparison, I guess that the person would not set 'compare' as the source in the period.

sbidoul:
> But if a KPI is not suitable for comparison, I guess that the person would not set 'compare' as the source in the period.

@jbeficent my point is that different KPI (rows) in a report can have different types, hence different applicable comparison methods. So overriding all of them in one column can create meaning less comparisons for some cells in the column.

JordiBForgeFlow:
Oh, I see. Perhaps this means that when you define a KPI you should indicate the allowed comparison methods instead of proposing a single comparison method?

Continuing the example, the KPI for the profit margin should defined to allow "difference" as the only allowed comparison method. If you don't define a comparison method in the Period, the one from the KPI would be used by default. 

I understand that this means a change to the existing data model.... :S

sbidoul:
Yes, we need a new field such as "Allowed comparison methods" on mis.report.kpi.

elicoidal:
IMO at some point you need some kind of condition for the comparison.
eg: compare if % or compare if positive
or simply "compare if" with additional field
or best would be a simply "Calculate" with calculation field + a format field so that you can do whatever you want.   
It might get tricky and complex to set up but I assume this would be advanced user configuration stage anyway

sbidoul:
Testing the bot that labels stale PRs. @JordiBForgeFlow do you still have interest in this PR?

Issue #138, opened 2018-11-18
Analytic filter in report columns definition


bunhin:
I have a question,
is it possible if i have a kpi lines and want to display the total for each analytic account together as columns?, I mean each analytic account as a column in the report and displayed together and then put other period data as second row as the comparison such as below :

KPI and period                           analytic acc-1      analytic acc-2 analytic acc-3  ..and so on     Total

expenses                                   total an-acc-1      total-an-acc2  Total-an-acc3  ...and son       Total
expenses other period               total an-acc-1-op   ...........             ................... 

How to achieve this if it is possible with current state of mis-builder. Thank you

HaraldPanten:
Hi @bunhin ,

In my opinion, it shouldn't be posted as an issue, as it is a question...

I'm not sure if it's possible to do this with the new functionallity; filtering by analytic account. 

But, I'm pretty sure that it can be done using Sub KPI's functionallities.

Try something like:

                          Period 1                                              Period 2                                    ...
         an. acc. 1   an. acc. 2  an. acc. 3         an. acc. 1  an. acc. 2  an. acc.3         ...   ...   ...

KPI 1

KPI 2

KPI 3

....

Am I right @sbidoul ?

bunhin:
Hi @HaraldPanten, yes it should be posted as question, i am sorry for not put the question label
Thank you for response and may be i need to explore further with the sub.kpi if so.  My understanding at current state the column is should be defined (not dynamically added as number of analytic account grows for example) but it is for child rows detail

sbidoul:
Hi,

I'm fine with questions in github issues. I'm adding the question label when needed, no worries.

Using subkpi can work, if you use the move line filter in expression such as `balp[][('analytic_account_id.code', '=', 'XYZ')]`.

Another way to do it would be to add an analytic filter field on `mis.report.instance.period`, similar to what is done in [mis_builder_operating_unit](https://github.com/OCA/operating-unit/blob/10.0/mis_builder_operating_unit/model/mis_builder.py). 

I think it makes sense to have that in the core mis_builder module, and I'm willing to review a PR about that.

So I'm adding the enhancement label :)

HaraldPanten:
Cool, thx!

bunhin:
Hi again,
Thanks to @sbidoul 
One more question, is it possible to have a column that calculate percentage of a kpi line value of a period, **to a selected** other kpi line value on the same period.
sample of use case:
[...]

Issue #96, opened 2018-06-04
Generic move line domain at report and column level


sbidoul:
As a generalization of analytic filters, and to ease implementation of simple cases (such mis_builder_operating_unit), it may be useful to have a generic move line filter domain on
- mis.report.instance
- mis.report.instance.period

fclementic2c:
@sbidoul sorry I do not follow you. Could you explain further please ?

sbidoul:
For instance in mis_builder_operating_unit, you can add a move line filter by operating unit on reports and columns. 

This enhancement proposal is to generalize the mechanism to let a user define an arbitrary move line domain on reports and columns.

fclementic2c:
ok got it.

PR #168, opened 2019-03-22
[11.0] add mis_builder_subkpi_template

grindtildeath:
:warning: This is stil a work in progress. :warning: 

Description :

This module extends the functionality of MIS-Builder to add MIS Report Sub-KPI
templates.

Sub-KPI Templates allows to predefine a set of sub-KPIs to be used in a
MIS Report template.

A move_line_domain field is defined on Sub-KPI Template Lines, and allows to
specifiy an extra domain to be applied on the move lines for this sub-KPI.

When entering a KPI expression, it will not be needed to fill the expressions
for each sub-KPI, as the extra move line domain is applied automatically on
KPI expressions when flagging the KPI as multi.

Moreover, this module adds a sub-KPI filter at the MIS Report Instance level to
select which sub-KPIs should be displayed on the report.

I'm on holidays next week, but would be happy to read about any suggestions when I'll be back on Apr 1st.
cc @fclementic2c @yvaucher 

EDIT:  Depends on:
 - [ ] #170


mis_builder_subkpi_template/models/mis_report_instance.py
```diff
@@ -0,0 +1,72 @@
...
+        store=True,
+    )
+    subkpis_filter_active = fields.Boolean(
+        string='Sub-KPIs filter active',
+        default=False,
+    )
+    filter_subkpi_ids = fields.Many2many(
+        'mis.report.subkpi',
+        string='Sub-KPIs to display',
+    )
```
sbidoul: How is this different from the `subkpi_ids` field on `mis.report.instance.period`.
Is it the goal to simplify configuration by filtering subkpis at once for all columns?
In this case, it's best to add this feature in the core module.

fclementic2c: It is the same idea you are right.
In our case, subkpis = department so our accountant want to be able to display a P&L with department in columns then filter quickly from the report to a specific department. 

If we have your blessing, we can add this feature in the core module, yes :)

fclementic2c:
More important for us, @sbidoul : What do you think about the idea of having subkpis tempate that you could define separately from report template.
ex : you create subkpis for departments then you can create several reports based on the template "departments" -> expression will be built automatically with the multi syntax.

sbidoul:
> What do you think about the idea of having subkpis template

I must think about it. This sounds reminiscent of #138. Or a kind of generalization of mis_builder_operating_unit. subkpi template might not be the ideal name for the concept, not sure.

Can you give an example of subkpis template in your case? (screenshot or sample config)

fclementic2c:
@grindtildeath has our test DB, he will send you that next week when he is back

mis_builder_subkpi_template/models/mis_report_instance.py
grindtildeath: @sbidoul Yes it is, so I will extract the feature and open another PR to target the core module.

grindtildeath:
@sbidoul 
This is the example I used for our subKPI template departments:
image

We use https://github.com/OCA/account-analytic/tree/11.0/analytic_base_department to have the department_id filled on the analytic account.

It's therefore much easier to define multi subkpis through the onchange, instead of having to repeat the same domain multiple times:
Peek 2019-04-01 15-53

sbidoul:
Thanks for the screenshots. I understand better.

I've the nagging feeling that it's pushing the subkpi concept a bit too far, and it we could reach the same goal more easily with https://github.com/OCA/mis-builder/issues/96. 

But we can let this module incubate in bit. Could you just move the PR to OCA/mis-builder-contrib?

grindtildeath:
@sbidoul 
I extracted all the commits from this PR into multiple PRs :
 - SubKPI file on report instance: https://github.com/OCA/mis-builder/pull/170
 - mis_builder_subkpi_template module: https://github.com/OCA/mis-builder-contrib/pull/1
 - Preparation of subkpi expression: https://github.com/OCA/mis-builder/pull/171

I guess we can close this one and merge the others after reviews ?

PR #174, opened 2019-04-18
[fix] fix for storno accounting

badbole:
In storno accounting negative ammounts on credit or debit side is normal thing, 
this fix will take all ammounts (positive or negative) with same logic ( different from zero ammount. 

Applicable in all countries where storno accounting is applied .
(check module: account storno) 
Curtesy of Decodio dev team


badbole:
@sbidoul can this pass?

sbidoul:
Looks good, thanks.

PR #182, opened 2019-04-29
[FIX] propagate context to xlsx report

### LINKED ISSUE #178: The Analytic Account filter don't work when exporting to xls
Hello,

I have version 11.0.3.3.0 of "mis_builder".
I am using Analytic Account to filter the reports, the filter is applied correctly in the preview and when generating the pdf, but when exporting to xls the filter don't apply.

Is it a bug? Do I need to configure something else?

Thanks

sbidoul:
fixes #178


sbidoul:
Tested again.

PR #193, opened 2019-05-10
[IMP] Add date/time of report generation in XLS export (issue #191)

### LINKED ISSUE #191: Add date/time of report generation in XLS export
This could be added at the bottom of the report.

apineux:
closes #191

### DIFF
--- a/mis_builder/report/mis_report_instance_xlsx.py
+++ b/mis_builder/report/mis_report_instance_xlsx.py
@@ -7,6 +7,8 @@ import logging
 import numbers
 
 from odoo.report import report_sxw
+from odoo import _, fields
+from datetime import datetime
 
 from ..models.accounting_none import AccountingNone
 from ..models.data_error import DataError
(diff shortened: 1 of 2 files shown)


sbidoul:
@apineux is covered by the ACSONE ECLA.

mis_builder/report/mis_report_instance_xlsx.py
```diff
@@ -142,6 +144,20 @@ def generate_xlsx_report(self, workbook, data, objects):
...
+        footer_format = workbook.add_format({
+            'italic': True, 'font_color': '#202020', 'size': 9})
+        now = datetime.now()
+        lang_model = self.env['res.lang']
+        lang = lang_model._lang_get(self.env.user.lang)
+        date_format = lang.date_format
+        time_format = lang.time_format
+        create_date = _("Generated on {} at {}").format(
+            now.strftime(date_format),
+            now.strftime(time_format))
```
sbidoul: I think there something better to get now in the user timezone and language. Maybe `tools.format_tz()`?

sbidoul:
Can you also add a file named `191.feature` in readme/newfragments. This project uses [towncrier](https://pypi.org/project/towncrier/) to generate the changelog.

mis_builder/report/mis_report_instance_xlsx.py
apineux: I'm not sure that we want to depend on a function declared in addons.mail.models.mail_template?
We can still use:
```suggestion
utc_now = datetime.now()
tz = pytz.timezone(self.env.user.tz or "UTC")
now = pytz.utc.localize(utc_now).astimezone(tz)
create_date = _("Generated on {} at {}").format(
    now.strftime(date_format),
    now.strftime(time_format))
```

sbidoul: @apineux indeed depending on the mail module is not a good idea.
I think `fields.Datetime.context_timestamp` is the way to go.

sbidoul:
The newfragment should contain the changelog entry as a short paragraph;

sbidoul (changes requested):
Better way to get local time

apineux:
@sbidoul Can you validate (or unvalidate) this development? Thanks

sbidoul (approved):

sbidoul:
Squashed. Please /ocabot merge

Issue #165, opened 2019-02-22
Setting Report from-date to year start ("Year-to-date")


alyngsoe:
We often need to specify a report column as "Year-to-date", meaning:
From date: first day of the actual year (from base date)
To date: Base date
Is there a way to do this ?

![year-to-date](https://user-images.githubusercontent.com/12607073/53232076-9c77e280-368a-11e9-9050-69231926e0e6.png)

sbidoul:
From the top of my head, this is not currently possible in an automated way. You need to adjust the end date manually.

sbidoul:
Closed via #194

PR #194, opened 2019-05-14
mis_builder: Add support for "Year to date" relative date type in periods

MaxyMoos:
This change seemed pretty straightforward and works without any issues so far on V11 so I backported to V10 directly.

Hopefully there isn't a side-effect or issue cause by this modification that I haven't been able to spot with my current usage/workflow with mis_builder :crossed_fingers:

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -79,7 +79,8 @@ class MisReportInstancePeriod(models.Model):
                  'report_instance_id.comparison_mode',
                  'date_range_type_id',
                  'type', 'offset', 'duration', 'mode',
-                 'manual_date_from', 'manual_date_to')
+                 'manual_date_from', 'manual_date_to',
+                 'is_ytd')
     def _compute_dates(self):
         for record in self:
             record.date_from = False
@@ -161,6 +162,9 @@ class MisReportInstancePeriod(models.Model):
                         record.date_from = periods[0].date_start
                         record.date_to = periods[-1].date_end
                         record.valid = True
+            if record.mode == MODE_REL and record.valid and record.is_ytd:
+                record.date_from = fields.Date.from_string(record.date_to)\
+                    .replace(day=1, month=1)
 
     _name = 'mis.report.instance.period'
     _description = 'MIS Report Instance Period'
@@ -181,6 +185,11 @@ class MisReportInstancePeriod(models.Model):
          ('y', _('Year')),
          ('date_range', _('Date Range'))],
         string='Period type'
+    )
+    is_ytd = fields.Boolean(
+        default=False,
+        string='Year to date',
+        help='Forces the start date to Jan 1st of the relevant year'
     )
     date_range_type_id = fields.Many2one(
         comodel_name='date.range.type',
--- a/mis_builder/views/mis_report_instance.xml
+++ b/mis_builder/views/mis_report_instance.xml
@@ -231,6 +231,7 @@
                         <group name="relative" attrs="{'invisible': [('mode', '!=', 'relative')]}" colspan="4">
                             <group>
                                 <field name="type" attrs="{'required': [('mode', '=', 'relative')]}"/>
+                                <field name="is_ytd"/>
                                 <field name="date_range_type_id"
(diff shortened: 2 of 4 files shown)


mis_builder/models/mis_report_instance.py
```diff
@@ -135,6 +135,16 @@ def _compute_dates(self):
                 record.date_from = fields.Date.to_string(date_from)
                 record.date_to = fields.Date.to_string(date_to)
                 record.valid = True
+            elif record.mode == MODE_REL and record.type == 'ytd':
+                date_from = d.replace(month=1, day=1)
+                date_from = date_from + \
+                    relativedelta(years=record.offset)
+                today = fields.Date.context_today(record)
+                date_to = fields.Date.from_string(today) + \
+                    relativedelta(years=record.offset)
```
sbidoul: @MaxyMoos I've a little doubt about this. Let's say you have a base date of '2018-05-05' and offset 0.
This would give you a period of '2018-01-01' to '2019-05-05'. That might be unexpected.

Also, I'm wondering if we don't need variants such as 'year to end of month', etc. For accounting reports for instance, `today` is rarely a very useful date.

So maybe this ytd option should instead be a flag next to the 'mode' and `type` fields where `date_to` is computed with the current algorithm and `date_start` replaces month and day by 1 in date_start.

sbidoul:
refs #165

mis_builder/models/mis_report_instance.py
MaxyMoos: @sbidoul Indeed that sounds like a much better approach. I pushed such modifications with some tests just now.

mis_builder/views/mis_report_instance.xml
```diff
@@ -231,6 +231,7 @@
                         <group name="relative" attrs="{'invisible': [('mode', '!=', 'relative')]}" colspan="4">
                             <group>
                                 <field name="type" attrs="{'required': [('mode', '=', 'relative')]}"/>
+                                <field name="is_ytd" force_save="1"/>
```
sbidoul: @MaxyMoos is force_save necessary? Why?

mis_builder/models/mis_report_instance.py
```diff
@@ -75,11 +75,13 @@ class MisReportInstancePeriod(models.Model):
     """
 
     @api.multi
+    @api.onchange('is_ytd')
```
sbidoul: Why this onchange decorator? I would have thought the depends was enough.

mis_builder/views/mis_report_instance.xml
MaxyMoos: Well, to be honest, I can't remember now what problem I was trying to solve by using this...
But trying it out quickly today I can't seem to find any issues arising without having the `force_save` attribute, so I'll get rid of it.

mis_builder/models/mis_report_instance.py
MaxyMoos: It is indeed enough, the depends triggers the onchange, I dropped the decorator in the latest push

MaxyMoos:
@sbidoul Thanks a lot for taking the time to review & comment, I force pushed the suggested changes on my branch :+1:

rgl1234:
Hello @MaxyMoos 
I validated this improvement in V11 and could not find any issues so far.... I would appreciate to have this functionality soon in V11.
René

rgl1234:
Hello @MaxyMoos 
I just faced an issue but I am not sure if it is related to this improvement or not. I tried to add a sub-kpi 

subkpi_test

but then I always get the following error when I want to view the report

subkpi_error

René

sanrav:
Functional testing successful. Also with sub-kpi's.
@rgl1234 you should try with expressions like :
image

leemannd:
Hello @sbidoul as @rgl1234 has tested it into V11 with success, do you have time to forward-port it to V11?

sbidoul:
@leemannd yes, I'll merge #211 and do all the ports and a release this week.

leemannd:
Ok. If it makes you gain time I can cherry-pick them and open a PR

sbidoul:
No, I'll do a git merge to port all 10.0 latest changes at once.

Issue #125, opened 2018-09-17
error while mixing kpis and subkpis (and managed to get  "Probably not your fault... but I'm really curious to know") 


aminesoulaymani:
I got an error, while my logs doesn't show anything special, so i cannot actually describe it.
how to reproduce this error:
1-create a mis builder template with: 1 subkpis. no queries. add a kpi with "multi" ticked, add the previous subkpi, put a number as expression; value type, comparison method and so doesn't matter. add another kpi, put a number as expression, other options doesn't matter either.
2- create a report, select your template, click on preview. everything perfect, it should work.
3- now go back to the template, add another subkpi. actually, there is no need to add it to the main kpi, the related report should crash if you try to preview it. adding the new subkpi to any kpi results the same crash and message "Probably not your fault... but I'm really curious to know how you managed to raise this error so I can handle one more corner case!"

best regards

 Odoo 10.0-20171015 (Community Edition)
MIS Builder 10.0.3.1.1 
Mis Budget not installed.

sbidoul:
Thanks for the detailed scenario. I'll try to reproduce as soon as possible.

In the meantime could you try the same with #116 and tell me if it works better?

sbidoul:
@aminesoulaymani #116 is now merged. Could you try again?

sbidoul:
Closing since no feedback. Check #116 and #196.

PR #196, opened 2019-05-20
[IMP] Better error message when expressions are not tuples in presence of SubKPIs

MaxyMoos:
Maybe related to #125

I got the "_Probably not your fault... but I'm really curious to know how you managed to raise this error so I can handle one more corner case!_" error message by using a "title row" KPI with _AccountingNone_ (or empty) value while having another KPI using subKPIs in the same report.

During report computation, the _AccountingNone_ KPI throws an error:

(mis_builder/models/mis_report.py, L686)
```python
if subkpis and not kpi.multi:
    if isinstance(vals[0], tuple):
        vals = vals[0]
        assert len(vals) == col.colspan
    elif isinstance(vals[0], DataError):
        vals = (vals[0],) * col.colspan
    else:
        raise UserError(_("Probably not your fault... but I'm "
                          "really curious to know how you "
                          "managed to raise this error so "
                          "I can handle one more corner "
                          "case!"))
```

We arrive with vals[0] equal to _AccountingNone_ so we neither enter the _tuple_ nor the _DataError_ condition, and the error is thrown.
I fixed it by adding an explicit check on _AccountingNone_ below the DataError check.

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -36,6 +36,14 @@ from .mis_kpi_data import (
 _logger = logging.getLogger(__name__)
 
 
+class SubKPITupleLengthError(UserError):
+    pass
+
+
+class SubKPIUnknownTypeError(UserError):
+    pass
+
+
 class AutoStruct(object):
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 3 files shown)


sbidoul:
Could you try `(AccountingNone, AccountingNone, AccountingNone)` as expression for your "title" KPI - assuming you have 3 subkpis? Or check "multi" on that KPI and put AccountingNone for each subkpi expression?

If that works as I suspect I'd say the error message needs to be improved rather than changing the algorithm.

MaxyMoos:
Just figured out by reading the existing unit tests (a bit too late...) that it is indeed supposed to work the way you explained rather than define a single `AccountingNone` value in my "title" KPI.

Just tested it and it does work perfectly when used this way, sorry about that, I am closing this PR.

sbidoul:
No problem, thanks for reporting. I'll keep it open to improve the error message.

MaxyMoos:
The least I could do is try & make it happen then: I force-pushed such modifications on my branch.
Let me know if that looks OK to you or if you'd rather pick it up from there.

mis_builder/models/mis_report.py
```diff
@@ -693,6 +697,16 @@ def _declare_and_compute_col(self,
                         assert len(vals) == col.colspan
                     elif isinstance(vals[0], DataError):
                         vals = (vals[0],) * col.colspan
+                    elif vals[0] == AccountingNone:
```
sbidoul: Actually I think this could simply be a `else` and therefore replace the other weird error message. I think we understand it well enough now. So the error could be `KPI {} has type {} while a tuple was expected. This can be fixed by ...`

sbidoul (commented):
Thanks a lot for the contrib!

mis_builder/models/mis_report.py
sbidoul: BTW, the assert on the tuple length 3 lines above could also be replaced by a similar friendly error message.

MaxyMoos: Thanks for the suggestions, that's done and pushed :+1:

sbidoul (approved):
Looks good.

Do not hesitate to add your name in the contributors list in (readme/CONTRIBUTORS) if you want.

MaxyMoos:
Glad I could help!

PR #201, opened 2019-05-28
[11.0][IMP] add analytic account filter to mis_builder reports

AdriaGForgeFlow:
This PR is intended to migrate the analytic account filtering feature that was available in v10.

image

cc ~ @Eficent @sbidoul

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -295,7 +295,7 @@ class MisReportInstancePeriod(models.Model):
         mis_report_filters = self.env.context.get('mis_report_filters', {})
         for filter_name, values in mis_report_filters.items():
             if values:
-                value = values.get('value')
+                value = values.get('value')[0]
                 operator = values.get('operator', '=')
                 filters.append((filter_name, operator, value))
         return filters
@@ -463,6 +463,9 @@ class MisReportInstance(models.Model):
     analytic_account_id = fields.Many2one(
         comodel_name='account.analytic.account', string='Analytic Account',
         oldname='account_analytic_id')
+    analytic_account_filter_id = fields.Many2one(
+        comodel_name='account.analytic.account',
+        string='Analytic Account Filter')
     hide_analytic_filters = fields.Boolean(default=True)
 
     @api.onchange('company_id', 'multi_company')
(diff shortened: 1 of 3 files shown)


mis_builder/static/src/js/mis_report_widget.js
```diff
@@ -74,13 +101,104 @@ odoo.define('mis_builder.widget', function (require) {
...
+            self.analytic_account_id_m2o.prependTo(self.get_mis_builder_filter_box());
+
+            self.analytic_account_id_m2o.$external_button.hide();
         },
 
         refresh: function () {
             this.replace();
         },
 
+        _onFieldChanged: function (event) {
```
apineux: Event if camelCase is the preferred style, you should use the underscores style to be consistent with the rest of the file.

apineux (commented):
LGTM except a small style choice.

apineux:
@ageficent I tested your PR and there is a problem with the filter.
You cannot set an empty value once it contains a value.

mis_builder/static/src/js/mis_report_widget.js
```diff
@@ -74,13 +101,104 @@ odoo.define('mis_builder.widget', function (require) {
...
 
         refresh: function () {
             this.replace();
         },
 
+        on_field_changed: function (event) {
+            var self = this;
+            if (event && event.data.changes) {
+                var changes = event.data.changes;
+                if (changes['analytic_account_filter_id']) {
```
apineux, line 194:
```suggestion
                if (changes['analytic_account_filter_id'] !== undefined) {
```
In case the value is "false"

mis_builder/models/mis_report_instance.py
```diff
@@ -463,6 +463,9 @@ def _default_company_id(self):
     analytic_account_id = fields.Many2one(
         comodel_name='account.analytic.account', string='Analytic Account',
         oldname='account_analytic_id')
+    analytic_account_filter_id = fields.Many2one(
```
apineux:
```suggestion
```
This field is not necessary

mis_builder/models/mis_report_instance.py
```diff
@@ -463,6 +463,9 @@ def _default_company_id(self):
...
+        comodel_name='account.analytic.account',
+        string='Analytic Account Filter')
```
sbidoul, line 468: Unfortunately adding a field sounds like a blocking issue? It has security implication (user needs write access on the template). And it probably break when several users preview the same report?

apineux: It is executing a write on the field analytic_account_filter_id so, yes @sbidoul you are right.

sbidoul:
Thanks a lot for this work @ageficent, and sorry for the delay in reviewing. As mentioned in previous comments adding a new field, and requiring write access to the report for users in preview mode is unfortunately a showstopper. So I'll close this in favor of #229. Your review there is most welcome.

Issue #20, opened 2017-10-01
monetary type


sbidoul:
A monetary kpi type which behaves exactly as the numeric type, except it renders with the proper monetary sign.

sbidoul:
@Martronic-SA you had a branch somewhere with that, did you?

Martronic-SA:
yes, here it is [#5](https://github.com/Martronic-SA/account-financial-reporting/pull/5)
basically this commit [#9b220a8c155d1f68e9486cb05421bb83dbcce622](https://github.com/Martronic-SA/account-financial-reporting/commit/9b220a8c155d1f68e9486cb05421bb83dbcce622)

PR #208, opened 2019-07-04
Adds Monetary type for data values

Martronic-SA:
Adds Monetary type for data values. This will format the number as currency value.

### DIFF
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -179,7 +179,7 @@ class MisReportKpiStyle(models.Model):
 
     @api.model
     def render_mon(self, lang, value, prefix=None, suffix=None,
-            currency=None):
+                   currency=None):
         # format number following user language
         if value is None:
             return u''
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -42,6 +42,7 @@ PROPS = [
 TYPE_NUM = 'num'
 TYPE_PCT = 'pct'
 TYPE_STR = 'str'
+TYPE_MON = 'mon'
 
 CMP_DIFF = 'diff'
 CMP_PCT = 'pct'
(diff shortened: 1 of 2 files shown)


sbidoul:
Refs #20 

Hi, thanks for reviving this feature :)

Looks good overall. Could you add a test (in test_render.py) and propose it to 10.0 instead of 11? I'll then merge and port to 11 myself, the code should be identical.

mis_builder/models/mis_report_style.py
```diff
@@ -163,16 +164,48 @@ def merge(self, styles):
...
             return self.render_str(lang, value)
 
+    @api.model
+    def render_mon(self, lang, value, prefix=None, suffix=None,
+                   currency=None):
+        # format number following user language
+        if value is None:
+            return u''
+        elif value is AccountingNone:
+            value = 0.0
```
sbidoul: AccountingNone must render as an empty string.

Martronic-SA: Why did I changed that .... probably a good reason but I don't remember..

mis_builder/models/mis_report_style.py
```diff
@@ -163,16 +164,48 @@ def merge(self, styles):
...
+        if not currency:
+            currency = self.env.user.company_id.currency_id
+        fmt = "%.{0}f".format(currency.decimal_places)
+        r = lang.format(
+            fmt, currency.round(value),
+            grouping=True, monetary=True).replace(r' ', u'\N{NO-BREAK SPACE}')
```
sbidoul, line 191: Can you also add the part with non-breaking-hyphen from render_num?

mis_builder/models/mis_report_style.py
Martronic-SA: I think that we've had clients who prefered to have 0.00 EUR instead of a empty cell in their reports

sbidoul: MIS Builder goes to great lenght at distinguishing 0 from no data.

For situations where customers insist at having 0.0 displayed for no data, it can be handled in expressions like this: `balp[...] or 0.0`.

Martronic-SA:
I've made a PR to the v 10.0
https://github.com/OCA/mis-builder/pull/210

PR #209, opened 2019-07-04
Automatic periods for Mis Report

Martronic-SA:
The aim is to generate automatic columns in the report based on a subdivision of time.
First you choose the main period (This Year, Last Year, This Month, Last Month or custom) and the wanted subdivision like week, month, year or no subdivision.
The report will then autogenerate the columns:

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -386,7 +386,7 @@ class AutomaticPeriod(object):
 
     def _get_additional_query_filter(self, query):
         return []
-    
+
     def _get_additional_move_line_filter(self):
         return []
 
@@ -427,7 +427,7 @@ class MisReportInstance(models.Model):
     use_automatic_periods = fields.Boolean(string='Use Automatic Periods',
                                            default=False)
     subdivision = fields.Selection(
-        [('no',_('No')),
+        [('no', _('No')),
          ('week', _('Week')),
          ('month', _('Month')),
          ('year', _('Year'))], string='Default subdivision', default='no')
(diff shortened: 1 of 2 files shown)


Martronic-SA:
The automatic columns are not stored anywhere, its basically standard python class instanciated at each request.

sbidoul:
Hi, thanks for this contrib. It's an interesting feature, I like it.

However I'm not very comfortable with the fact it does not create real `mis.report.instance.period` records.

A alternative could be a kind of "repeat period" wizard that would appear as a button on a period, and would ask parameters for cloning it over time?

Martronic-SA:
The final idea behind that feature is to be able to change the columns in the report on the fly like here below (this is our own devs on mis_builder):

That's why I dont want to store the columns. User can ask for the current year/month or last year/month  or custom dates and subdivide by week or month or three months and year. It's "just" the display that changes

sbidoul:
I see, very interesting!

I'll look deeper then to check the impacts on other parts (such as analytic filters etc).

mis_builder/models/mis_report_instance.py
```diff
@@ -733,7 +871,11 @@ def _compute_matrix(self):
         aep = self.report_id._prepare_aep(
             self.query_company_ids, self.currency_id)
         kpi_matrix = self.report_id.prepare_kpi_matrix()
-        for period in self.period_ids:
+        if not self.use_automatic_periods:
```
Martronic-SA, line 892: The changes between standard periods and automatic periods is made here. I think that the whole machinery doesn't see the difference and it has no influence on the general behavior.

Martronic-SA:
In that PR, the definition for automatic periods is only done in the mis report instance, but there isn't the javascript with the ability to change the periods on the fly. Do you think that the javascript part should be part of that PR? (only Period selection and subdivision)

sbidoul:
I prefer to handle the js part in a separate PR... and we'll need to see how to handle porting of that js across Odoo versions as that part evolves much more rapidly :(

Martronic-SA:
Here : https://github.com/Martronic-SA/mis-builder/tree/mis_report_automatic_periods_js
I've added the corresponding javascript if you want to take a look at it and make some testings.

PR #210, opened 2019-07-05
Adds Monetary type for data values (v10.0)

Martronic-SA:
Adds Monetary type for data values. This will format the number as currency value.
See discussion here : https://github.com/OCA/mis-builder/pull/208


sbidoul:
Refs #20

mis_builder/models/kpimatrix.py
```diff
@@ -164,6 +164,8 @@ def __init__(self, env):
         self._sum_todo = {}
         # { account_id: account_name }
         self._account_names = {}
+        # id of the currency
+        self.currency_id = None
```
sbidoul, line 168: I suggest naming this currency instead of currency_id and passing it in the KpiMatrix constructor.

mis_builder/models/mis_report_style.py
```diff
@@ -154,16 +155,49 @@ def merge(self, styles):
...
+    @api.model
+    def render_mon(self, lang, value, prefix=None, suffix=None,
+                   currency=None):
+        # format number following user given or user currency
+        if value is None or value is AccountingNone:
+            return u''
+        if not currency:
+            currency = self.env.user.company_id.currency_id
+        elif isinstance(currency, int):
+            currency = self.env['res.currency'].browse(currency)
```
sbidoul, line 180: this test for int is superfluous, IMO, lets force people to pass a proper currency object

mis_builder/models/mis_report_style.py
```diff
@@ -154,16 +155,49 @@ def merge(self, styles):
...
+            grouping=True, monetary=True).replace(r' ', u'\N{NO-BREAK SPACE}')
+        r = r.replace('-', u'\N{NON-BREAKING HYPHEN}')
+        if not prefix:
+            prefix = u''
+        if not suffix:
+            suffix = u''
+        if not prefix and currency.position == 'before':
+            prefix = u'{symbol}\N{NO-BREAK SPACE}'
+        elif not suffix:
+            suffix = u'\N{NO-BREAK SPACE}{symbol}'
```
sbidoul, line 193: Why does prefix/suffix have precedence over the symbol?

sbidoul (commented):
Can you check why tests fail?

sbidoul:
The Excel export will need to support the monetary type too.
#211 prepares the ground for this (while fixing another bug).

mis_builder/models/kpimatrix.py
Martronic-SA: Should we have here in KpiMatrix class the id of the currency or the currency itself?
The value stored here is then passed to the Mis Report Style render method.

mis_builder/models/mis_report_style.py
Martronic-SA: Maybe if you want to keep the number monetary formatting and rounding but have something else as currency symbol. by the way you can't replace the currency symbol by nothing.
For example 1000.04 if CHF would give 1'000.05 CHF and you may want 1'000.05 Swiss Francs.
Maybe we should give the possibility to show/hide currency symbol or choose between currency code / symbol.

Issue #220, opened 2019-07-29
Percentage point (pp) values do not export correctly to Excel


sbidoul:
To reproduce, create a KPI of type Percentage, create two columns, and a comparison column.

In the comparison column, the KPI is displayed as `.pp`. In Excel it shows at percentage, multiplied by 100.

sbidoul:
This will be part of 3.5. It is testable on the staging PRs.

PR #223, opened 2019-08-29
[FIX] Percentage point (pp) values do not export correctly to Excel

apineux:
Fix issue https://github.com/OCA/mis-builder/issues/220

### DIFF
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -204,10 +204,26 @@ class MisReportKpiStyle(models.Model):
     def compare_and_render(self, lang, style_props, type, compare_method,
                            value, base_value,
                            average_value=1, average_base_value=1):
+        """
+        :param lang: res.lang record
+        :param style_props: PropertyDict with style properties
+        :param type: num, pct or str
+        :param compare_method: diff, pct, none
+        :param value: value to compare (value - base_value)
+        :param base_value: value compared with (value - base_value)
+        :param average_value: value = value / average_value
+        :param average_base_value: base_value = base_value / average_base_value
+        :return: tuple with 4 elements
+            - delta = comparison result (Float or AccountingNone)
+            - delta_r = delta rendered in formatted string (String)
+            - style_r = PropertyDict with style properties
+            - delta_type = Type of the comparison result (num or pct)
+        """
         delta = AccountingNone
         style_r = style_props.copy()
+        delta_type = TYPE_NUM
         if isinstance(value, DataError) or isinstance(base_value, DataError):
-            return AccountingNone, '', style_r
+            return AccountingNone, '', style_r, delta_type
         if value is None:
             value = AccountingNone
         if base_value is None:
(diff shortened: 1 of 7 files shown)


sbidoul:
Thanks. Can you add a test for that and a short code comment?

mis_builder/readme/newsfragments/220.bugfix
```diff
@@ -0,0 +1 @@
+Mis builder export in XLSX should keep the percentage point suffix (pp) instead of replacing it with %
```
sbidoul: "In Excel export, keep the percentage point suffix (pp) instead of replacing it with %."

This will read better in the changelog.

sbidoul:
@apineux .pp values look good, but % seem to have regressed. Can you check?
Also, the the underlying value for pp seems to be multiplied by 100. Since it's a difference of percentages my first intuition is that it should be -0.1025 in this example.

image

sbidoul:
Can you rebase to get the travis fix from the main branch?

apineux:
> @apineux .pp values look good, but % seem to have regressed. Can you check?
> Also, the the underlying value for pp seems to be multiplied by 100. Since it's a difference of percentages my first intuition is that it should be -0.1025 in this example.
> 
> image

@sbidoul It is not possible to send -0.1025 to excel and display it as -10.25 if the suffix is not %. That's the way excel is working.

sbidoul:
> It is not possible to send -0.1025 to excel and display it as -10.25 if the suffix is not %

Right, indeed. When the divider style is used for other numeric cells, the original value is lost in the excel export.

mis_builder/models/mis_report_style.py
```diff
@@ -274,7 +274,8 @@ def to_xlsx_style(self, type, props, no_indent=False):
             if props.dp:
                 num_format += u'.'
                 num_format += u'0' * props.dp
-            num_format += '%'
+            # A comparison TYPE_PCT can keep its suffix
+            num_format += props.suffix if props.keep_suffix else '%'
```
sbidoul: I'm not too comfortable with this `keep_suffix`.

Fundamentally, the root cause of this bug is that the comparison result is not of TYPE_PCT but is rather TYPE_NUM. 
The type of the comparison is different than the type of the compared values. This is the case when comparing two TYPE_NUM in percentage mode too, and it looks like what's in place for that can be considered as a workaround too.

So it think would be overall cleaner to:
- adapt `compare_and_render` to return the type of the comparison result as a fourth return value (the TYPE_NUM path can be adapted too)
- add a type property to `KpiMatrixCell`
- use that cell type property in priority over row.kpi.type if it is set when rendering

It's a bigger change but I hope the result should be much easier to understand.

apineux:
Now the KpiMatrixCell has a new property called "val_type".
By default this property has the same value as the KPI type (num or pct).
In case of comparison, this val_type can change:
- Comparison of PCT & PCT will return a NUM result
- PCT Comparison of NUM & NUM will return a PCT result

The rendering in xlsx is now based on that val_type instead of the KPI type.

sbidoul:
@apineux thanks. It's much cleaner and future proof.

I pushed another commit with a few minor additional improvements to related code (mainly using `render` instead of `render_num` in `compare_and_render` which is now possible since we properly handle the type).

sbidoul (approved):

Issue #222, opened 2019-08-28
mis_builder_demo: improve commitment view


sbidoul:
Improve the commited purchase SQL view :

- analytic tags
- currency (convert to company currency)

@apineux

sbidoul:
Also, the draft invoices part of the sql view needs improvement:

- ~~handle out_* invoices~~
- take the invoice type in computing the sign

sbidoul:
And while we are at it we could rename this to commitment instead of purchase commitment, and start including draft customer invoices/refunds. Draft sales can be left for future work.

sbidoul:
This was done as part of #224

PR #228, opened 2019-09-03
[10.0][IMP] Add analytic tags as filter for mis_builder

apineux:
The improvement adds a new filter (on account.analytic.tag) in the mis builder preview.
Capture d’écran de 2019-09-03 14-59-45

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -310,7 +310,8 @@ class MisReportInstancePeriod(models.Model):
         for filter_name, values in mis_report_filters.items():
             if values:
                 value = values.get('value')
-                operator = values.get('operator', '=')
+                value = [value] if not isinstance(value, list) else value
+                operator = values.get('operator', 'in')
                 filters.append((filter_name, operator, value))
         return filters
 
@@ -476,6 +477,9 @@ class MisReportInstance(models.Model):
     analytic_account_id = fields.Many2one(
         comodel_name='account.analytic.account', string='Analytic Account',
         oldname='account_analytic_id')
+    analytic_tag_id = fields.Many2one(
+        comodel_name='account.analytic.tag', string='Analytic Tag',
+        oldname='account_tag_id')
     hide_analytic_filters = fields.Boolean(default=True)
 
     @api.onchange('company_id', 'multi_company')
@@ -499,14 +503,20 @@ class MisReportInstance(models.Model):
     @api.model
     def get_filter_descriptions_from_context(self):
         filters = self.env.context.get('mis_report_filters', {})
-        analytic_account = filters.get('analytic_account_id', {})
-        analytic_account_id = analytic_account.get('value')
+        analytic_account_id = \
+            filters.get('analytic_account_id', {}).get('value')
         filter_descriptions = []
         if analytic_account_id:
             analytic_account = self.env['account.analytic.account'].browse(
                 analytic_account_id)
             filter_descriptions.append(
                 _("Analytic Account: %s") % analytic_account.display_name)
+        analytic_tag_id = filters.get('analytic_tag_ids', {}).get('value')
+        if analytic_tag_id:
+            analytic_tag = self.env['account.analytic.tag'].browse(
+                analytic_tag_id)
+            filter_descriptions.append(
+                _("Analytic Tag: %s") % analytic_tag.name)
         return filter_descriptions
 
     @api.multi
@@ -592,6 +602,10 @@ class MisReportInstance(models.Model):
         if self.analytic_account_id:
             context['mis_report_filters']['analytic_account_id'] = {
                 'value': self.analytic_account_id.id,
+            }
+        if self.analytic_tag_id:
+            context['mis_report_filters']['analytic_tag_ids'] = {
+                'value': self.analytic_tag_id.id,
             }
 
     @api.multi
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -310,9 +310,14 @@ class MisReportInstancePeriod(models.Model):
         for filter_name, values in mis_report_filters.items():
             if values:
                 value = values.get('value')
-                value = [value] if not isinstance(value, list) else value
-                operator = values.get('operator', 'in')
-                filters.append((filter_name, operator, value))
+                if not value:
+                    continue
+                if not isinstance(value, list):
+                    value = [value]
+                many_ids = self.report_instance_id.resolve_2many_commands(
+                    filter_name, value, ['id'])
+                for m in many_ids:
+                    filters.append((filter_name, 'in', [m['id']]))
         return filters
 
     @api.multi
@@ -477,9 +482,8 @@ class MisReportInstance(models.Model):
     analytic_account_id = fields.Many2one(
         comodel_name='account.analytic.account', string='Analytic Account',
         oldname='account_analytic_id')
-    analytic_tag_id = fields.Many2one(
-        comodel_name='account.analytic.tag', string='Analytic Tag',
-        oldname='account_tag_id')
+    analytic_tag_ids = fields.Many2many(
+        comodel_name='account.analytic.tag', string='Analytic Tags')
     hide_analytic_filters = fields.Boolean(default=True)
 
     @api.onchange('company_id', 'multi_company')
@@ -511,12 +515,13 @@ class MisReportInstance(models.Model):
                 analytic_account_id)
             filter_descriptions.append(
                 _("Analytic Account: %s") % analytic_account.display_name)
-        analytic_tag_id = filters.get('analytic_tag_ids', {}).get('value')
-        if analytic_tag_id:
-            analytic_tag = self.env['account.analytic.tag'].browse(
-                analytic_tag_id)
+        analytic_tag_value = filters.get('analytic_tag_ids', {}).get('value')
+        if analytic_tag_value:
+            analytic_tag_names = self.resolve_2many_commands(
+                'analytic_tag_ids', analytic_tag_value, ['name'])
             filter_descriptions.append(
-                _("Analytic Tag: %s") % analytic_tag.name)
+                _("Analytic Tags: %s") %
+                ', '.join([rec['name'] for rec in analytic_tag_names]))
         return filter_descriptions
 
     @api.multi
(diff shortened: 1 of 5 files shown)


fclementic2c:
I usually use analytic tags as subkpi (ex: departement or Business Units).
Think to test it will not break in these cases.

sbidoul:
@fclementic2c it should not break anything if you don't use the new filter. And these analytic filters are applied with a AND on move line domains that would already be in expressions.

mis_builder/models/mis_report_instance.py
```diff
@@ -500,14 +504,19 @@ def _compute_query_company_ids(self):
...
-        analytic_account = filters.get('analytic_account_id', {})
-        analytic_account_id = analytic_account.get('value')
+        analytic_account_id = filters.get('analytic_account_id', {}).get('value')
         filter_descriptions = []
         if analytic_account_id:
             analytic_account = self.env['account.analytic.account'].browse(
                 analytic_account_id)
             filter_descriptions.append(
                 _("Analytic Account: %s") % analytic_account.display_name)
+        analytic_tag_id = filters.get('analytic_tag_ids', {}).get('value')
```
sbidoul: s/analytic_tag_ids/analytic_tag_id/ ?

sbidoul:
@benwillig could you review this one please as you are very familiar with this part of the code?

mis_builder/models/mis_report_instance.py
apineux: Yes because the filter is a m2o (select one tag at a time) but we are looking on the field "analytic_tag_ids" on move.line. Therefore the name of the filter is the name of the field on which we will search.

benwillig (approved):
Code review. LGTM

sbidoul:
@apineux can you add `228.feature` for then changelog?

pedrobaeza:
Why not doing both filters as m2m. In tags is even more needed as it's usual to want to filter by several tags at the same time. It's not too much change at code level (in operator instead of =) and using m2m tags widget makes a similar UI.

apineux:
It was my first idea to use a m2m as filter but some people find confusing the fact that filtering with a m2m is searching with OR between each element selected in the m2m.

M2o or M2m for these filter is a debatable topic.

pedrobaeza:
With a proper help it should be good explained, and having m2m serves for filtering one or several accounts/tags. This way, you can only filter by one. Other option is to provide a char field with domain widget for building advanced domains, or even to have 4 fields for AND and OR operations (not my best option, but preferred over not having the option).

sbidoul:
> some people find confusing

I am some people :)

So this PR as it is now allows to filter on *one* tag, like one can filter on *one* analytic account, it's a simple extension of what exists now. Let's call this solution 0.

If we allow to select more than one tag in the filter box, we have two possibilities
1. filter on move lines which have *one of* the selected tags: in this case the analytic account filter should be changed too to work the same way for symmetry reasons; this means a data model changes and a bigger impact on existing deployments
2. filter on move lines which have *all* the selected tags: this is probably the most useful for people using tags for several analytic dimensions; for example if people use tags for, say, projects and departments, then it's useful to filter on project, or on departement, or on a project within a department; in this use case with multiple dimensions, doing a OR is probably counter productive

I'm ok with solution 0. 

I'm not in favor of 1. because it's a data model change, and more importantly it covers only some more use cases while raising a lot of questions at the same time (AND/OR, combinations, etc).

The more I think to it the more I think solution 2. is best for this PR.

And also, in another PR, investigate how hard it would be implement the domain widget. Or propose a selection of existing domain filters. That would be a completely generic solution. It can be implemented in two steps: first a domain filter in the backend view (with no widget in the preview), then add the domain filter in the preview.

pedrobaeza:
OK for solution 2 in this PR, but the OR is needed as well as said as I have that user case. Adding the domain widget is easy and I can put a JS developer for that if you want.

apineux:
Solution 2 has been implemented.
All the tags selected in the filter must belong to the record (account.move.line) so that the record is returned.

**Suggestion**: I can add a checkbox (or selection field) to be able to choose between AND search or OR search for the tags.
Let me know if it is a good idea.

mis_builder/models/mis_report_instance.py
```diff
@@ -310,8 +310,14 @@ def _get_filter_domain_from_context(self):
...
-                operator = values.get('operator', '=')
-                filters.append((filter_name, operator, value))
+                if not value:
+                    continue
+                if not isinstance(value, list):
+                    value = [value]
+                many_ids = self.report_instance_id.resolve_2many_commands(
+                    filter_name, value, ['id'])
+                for m in many_ids:
+                    filters.append((filter_name, 'in', [m['id']]))
```
sbidoul: @apineux this part now ignores of the `operator`. I think it is important to preserve it because some extensions depend on it. I've added a few unit tests on the 10.0 branch to make sure the important parts are guaranteed. cc/ @benwillig 

I'm wondering if we should not handle a new operator here (named `all`?), since what is actually does is not supported by any of the standard Odoo operators.

sbidoul (changes requested):
Could you rebase on 10.0 and adapt to preserve the operator?

sbidoul:
Analytic tags are shown before analytic accounts in the widget. I kind of remember they are displayed after (or on the right of) analytic accounts in most places in Odoo. Could you check that too?

mis_builder/models/mis_report_instance.py
```diff
@@ -307,11 +307,22 @@ def _onchange_source(self):
...
-            if values:
-                value = values.get('value')
-                operator = values.get('operator', '=')
-                filters.append((filter_name, operator, value))
+        for filter_name, domain in mis_report_filters.items():
+            if domain:
+                value = domain.get('value')
+                operator = domain.get('operator')
+                if not value:
+                    continue
```
sbidoul: Is this continue important? Testing for "is not set" (= False) could make sense in some situations.

sbidoul:
A few trailing remarks. I notice GitHub does not display them all correctly in context. They are best viewed in https://github.com/OCA/mis-builder/commit/79670680011aa1b4e40690d33428f632ff653fd2

sbidoul (approved):

sbidoul:
Thanks @apineux !

@vdewulf if you are still doing mis builder doc today, you might like to test (and document?) this one: it adds support for analytic tags.

djulien19:
Hey,
The tags appears above, instead of inside the box:
image

vdewulf:
@apineux tested on runbot instance http://3389032-228-9c516b.runbot1.odoo-community.org
I activated the analytic accounting, added some invoices with analytic account "Camp to Camp - Camptocamp" and the "Contract" analytic tag, and others with analytic account "Administrative" and "R&D" analytic tag.

In the Demo Expenses report, when selecting an analytic account, everything works fine.
When I select an analytic tag (for example "Contract", but same error occurs with the other analytic tag) and click on refresh, I've got this error:


Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 642, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 684, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 334, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 327, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 942, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 507, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 895, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 887, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 680, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/odoo/build/OCA/mis-builder/mis_builder/models/mis_report_instance.py", line 789, in compute
    kpi_matrix = self._compute_matrix()
  File "/home/odoo/build/OCA/mis-builder/mis_builder/models/mis_report_instance.py", line 781, in _compute_matrix
    self._add_column(aep, kpi_matrix, period, period.name, description)
  File "/home/odoo/build/OCA/mis-builder/mis_builder_budget/models/mis_report_instance.py", line 64, in _add_column
    aep, kpi_matrix, period, label, description)
  File "/home/odoo/build/OCA/mis-builder/mis_builder/models/mis_report_instance.py", line 750, in _add_column
    aep, kpi_matrix, period, label, description)
  File "/home/odoo/build/OCA/mis-builder/mis_builder/models/mis_report_instance.py", line 726, in _add_column_actuals_alt
    no_auto_expand_accounts=self.no_auto_expand_accounts,
  File "/home/odoo/build/OCA/mis-builder/mis_builder/models/mis_report.py", line 813, in declare_and_compute_period
    aml_model)
  File "/home/odoo/build/OCA/mis-builder/mis_builder/models/aep.py", line 349, in do_queries
    ['account_id', 'company_id'], lazy=False)
  File "/.repo_requirements/odoo/odoo/models.py", line 1940, in read_group
    result = self._read_group_raw(domain, fields, groupby, offset=offset, limit=limit, orderby=orderby, lazy=lazy)
  File "/.repo_requirements/odoo/odoo/models.py", line 1963, in _read_group_raw
    query = self._where_calc(domain)
  File "/.repo_requirements/odoo/odoo/models.py", line 4070, in _where_calc
    e = expression.expression(domain, self)
  File "/.repo_requirements/odoo/odoo/osv/expression.py", line 643, in __init__
    self.parse()
  File "/.repo_requirements/odoo/odoo/osv/expression.py", line 821, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field u'analytic_tag_ids' in leaf "<osv.ExtendedLeaf: (u'analytic_tag_ids', 'in', [1]) on mis_committed_purchase (ctx: )>"


The display is better now than it was (cf. Julien's comment here above): the analytic tag is displayed in the field and not above anymore.

sbidoul:
@vdewulf the error you got is linked to the commit purchases data source (from mis_builder_demo) which is not yet adapted for analytic tag (this is covered in #222). If the column involving committed purchases is removed, it should work better.

vdewulf:
@sbidoul @apineux I tested again with a new template to avoid the issue encoutered yesterday.
Everything seems to work correctly. The analytic tag is well displayed, and the combination of filters (analytic account + analytic tag) works as well.
Just have to click on refresh button every time. Not an issue but could be a nice improvement for next year!

sbidoul:
Technical review and functional testing done at #OCAdays. Merging on 10.0-staging to do some integration testing before release.

PR #234, opened 2019-09-30
Update usage.rst

djulien19:


mileo:
Depends of https://github.com/OCA/mis-builder/pull/235

mileo:
This PR must be close because the text are already in https://github.com/OCA/mis-builder/pull/235 with the images

PR #241, opened 2019-10-11
[FIX] analytic tags in budget item overlap detection

sbidoul:


sbidoul:
Moving foward with this as it is a fairly obvious bug. Without this it is impossible to create budget items with different analytic tags.

PR #242, opened 2019-10-14
[10.0][IMP] mis_builder: Show company of account detail in multicompany

etobella:
With this PR, we are able to see the company of the account on the label.
image

This PR does not change the behavior when `multi_company` is not selected.
image

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -148,7 +148,7 @@ class KpiMatrixCell(object):
 
 class KpiMatrix(object):
 
-    def __init__(self, env):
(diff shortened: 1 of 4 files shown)


sbidoul:
Makes sense. +1

Could you do it as a PR to `10.0-staging`? I'm preparing a release across versions real soon now and bring it to 11 from there.

Also it would be great if you could look if there is an easy spot to add a test for that.

etobella:
Sure!

etobella:
Tests added, it should be :green_apple: now.

@jbeficent

JordiBForgeFlow (approved):
:+1:

sbidoul (approved):

PR #229, opened 2019-09-05
[11.0][MIG] Add filter on mis builder preview - Analytic Account & Analytic Tags

apineux:
The purpose of this PR is to migrate the filter (Account analytic account) on the mis_builder preview. In addition we also migrate the PR: https://github.com/OCA/mis-builder/pull/228

PR #144, opened 2018-12-04
[WIP][11.0] analytic account filter porting from v10

simahawk:
replaces #133 
depends on https://github.com/OCA/reporting-engine/pull/259

PR #245, opened 2019-10-21
[11.0] analytic filters in widget, take 5

sbidoul:
This is the last and final attempt to have a working and clean implementation of the analytic filters in the widget.

Building on previous tries and inspiration provided in #133, #143, #144, #201 and #229.

The main trick here is to combine the use of `BasicModel.makeRecord` (proposed by @apineux in #229) with the `StandaloneFieldManagerMixin` to handle events and manage the field widgets.

This PR also preserves the same data stucture for the `mis_report_filters` context key, so there are no backend code changes.

It works as in on 12.

### DIFF
--- a/mis_builder/static/src/css/custom.css
+++ b/mis_builder/static/src/css/custom.css
@@ -20,19 +20,10 @@
   text-decoration: underline;
 }
 
-.odoo .oe_mis_builder_buttons {
-  padding-bottom: 10px;
-  padding-top: 10px;
-}
-
 .oe_mis_builder_content {
   padding: 10px;
 }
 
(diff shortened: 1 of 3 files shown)


lmignon (approved):
LGTM (Code + Functional). A little bit surprised that the refresh is not automatic when we change filters values. I suppose that's to avoid unwanted time consuming calls when the user want to change more than one filter parameter..

sbidoul:
> refresh is not automatic

Yes that, and currently calling `refresh()` causes flickering.

So that can be left for future improvement.

sbidoul:
Thanks for the review!

PR #248, opened 2019-11-11
GitHub action for linting

sbidoul:
An experiment...


lmignon:
Interesting... Is the goal to decrease the number of travis runners by build?

PR #250, opened 2019-12-12
[IMP] new extension module

badbole:
Initial commit of new extension module.

Module enables filtering out lines with zero values, 
Shorten reports, while preserving all data visible.
Specialy usefull in cases where there is a lot of KPI lines.


sbidoul:
There is a (perhaps too little known) `Hide Empty` style option that should have more or less the same effect.

badbole:
> There is a (perhaps too little known) `Hide Empty` style option that should have more or less the same effect.

Agree on: "too little know" ... 
could you provide example if possible? 
does it work the same in all reports? (web, pdf, xls)

sbidoul:
> Agree on: "too little know" ...
> could you provide example if possible?

The [screenshot for Styles in the documentation](https://oca-mis-builder.readthedocs.io/en/10.0/usage.html#styles) shows it (at the bottom). Admittedly it lacks some explanations. You can uncheck "Hide Empty Inherit" and then tick the new "Hide Empty" box that appears.

> does it work the same in all reports? (web, pdf, xls)

Yes

badbole:
Excelent, THANKS!
will hace a look!
closing PR

sbidoul:
Thanks. Don't hesitate to do a documentation PR (10.0 branch, `docs` directory) with your findings.

PR #252, opened 2020-01-11
[MIG][WIP] 13.0 from 12.0

sbidoul:


sbidoul:
There still a few test failures to investigate, but it should otherwise be good enough to test on runbot.

mis_builder/__manifest__.py
```diff
@@ -12,10 +12,10 @@
     "website": "https://github.com/OCA/mis-builder/",
     "depends": [
         "account",
-        "board",
+        # "board",
```
MiquelRForgeFlow: why?

mis_builder_demo/models/mis_committed_purchase.py
```diff
@@ -36,7 +36,7 @@ class MisCommittedPurchase(models.Model):
         string="Analytic Tags",
     )
 
-    @api.model_cr
+    @api.model
```
MiquelRForgeFlow:
```suggestion
```

MiquelRForgeFlow (commented):
Minor things:

- `mis.kpi.data.test.item` needs `_description`

```diff
- def default_get(self, fields):  # pylint: disable=redefined-outer-name
+ def default_get(self, default_fields):
```
 
```diff
- ('company_id','child_of',[user.company_id.id])
+ ('company_id','in',company_ids)
```

mis_builder_demo/models/mis_committed_purchase.sql
```diff
@@ -74,12 +74,12 @@ CREATE OR REPLACE VIEW mis_committed_purchase AS (
           WHEN (ail.price_subtotal / COALESCE(cur.rate, 1.0))::decimal(16,2)  < 0 THEN (ail.price_subtotal / COALESCE(cur.rate, 1.0))::decimal(16,2)
           ELSE 0.0
         END AS credit
-        FROM account_invoice_line ail
-            LEFT JOIN account_invoice ai ON ai.id = ail.invoice_id
+        FROM account_move_line ail
+            LEFT JOIN account_move ai ON ai.id = ail.move_id
```
MiquelRForgeFlow, line 78:
```suggestion
            LEFT JOIN account_move ai ON (ai.id = ail.move_id and not ail.exclude_from_invoice_tab)
```
the same below in line 106

sbidoul: Ah, yes, this horror. Done

sbidoul:
>  mis.kpi.data.test.item needs _description

Done in 10.0, will be forward ported

> - def default_get(self, fields):  # pylint: disable=redefined-outer-name
> + def default_get(self, default_fields):

Done in 10.0, will be forward ported

> - ('company_id','child_of',[user.company_id.id])
> + ('company_id','in',company_ids)

Done

Thanks for the review!

mis_builder_demo/models/mis_committed_purchase.sql
```diff
@@ -74,13 +74,15 @@ CREATE OR REPLACE VIEW mis_committed_purchase AS (
...
             LEFT JOIN currency_rate cur on (cur.currency_id = ai.currency_id and
                 cur.company_id = ai.company_id and
-                cur.date_start <= coalesce(ai.date_invoice, now()) and
-                (cur.date_end is null or cur.date_end > coalesce(ai.date_invoice, now())))
-        WHERE ai.state = 'draft' AND ai.type IN ('in_invoice', 'out_refund')
+                cur.date_start <= coalesce(ai.invoice_date, now()) and
+                (cur.date_end is null or cur.date_end > coalesce(ai.invoice_date, now())))
+        WHERE ai.state = 'draft'
+          AND ai.type IN ('in_invoice', 'out_refund')
+          AND ail.exclude_from_invoice_tab
```
MiquelRForgeFlow:
```suggestion
          AND NOT ail.exclude_from_invoice_tab
```

mis_builder/__manifest__.py
sbidoul: this is resolved

mis_builder_demo/models/mis_committed_purchase.sql
sbidoul: good catch, thanks!

sbidoul:
This one is now complete. Does anyone want to do so more tests?

pedrobaeza:
I think it's OK to merge it as is and if anything else migrating localizations we can do further PRs.

HaraldPanten:
Yep, we can merge. I'll test it functionally and I'll tell if I find that something is not working well.

Thanks!

sbidoul:
Ok, let's go. After that I'll release 3.6 with the new features for 10,11,12,13.

I'll stop active maintenance of 9: I'll merge backports and fixes if someone do the PRs.

PR #254, opened 2020-02-10
[10.0] staging for version 3.6

sbidoul:
- [x] 9.0 is now in maintenance mode
- [x] some refactoring (#258)
- [x] Budget by account (similar to account_move_budget, but works with periods, ~~and allow a combined budget on KPIs and accounts~~) (#259)
- [x] Sub reports (use KPIs from other reports #155, #256)
- [x] Column-level analytic filters (#138, #261)

PR #256, opened 2020-02-11
[ADD] subreports

### LINKED ISSUE #155: Access to KPIs from other reports in KPI expressions, aka subreports
A report template `mis.report` can have access to KPI's provided by other reports templates.

This would mean, for example, that you could define a MIS Report for a "Balance Sheet", and then have another MIS Report "Balance Sheet Ratios" that fetches kpi's from "Balance Sheet" to create new KPIS for the ratios (e.g. Current Assets / Total Assets).

Design ideas:
* on `mis.report` add a list of dependent reports, each with a name
* when evaluating a report (in `declare_and_compute_period`), add recursively evaluate dependent reports to add entry in `locals_dict` for each of them
* investigate the impact on mis_builder_budget

In the example above, the ratios report could then have expressions such as `balance_sheet.current_assets / balance_sheet.total_assets`).

sbidoul:
Closes #155 

With this PR, you can add named "sub reports" on  a report template.
When that is done, KPIs from sub reports can be used in expressions of the the parent report with the following syntax `<subreport name>.<kpi name>`.

~~The code is WIP and will be refactored, but the concept and data model should be ok.~~

cc/ @fclementic2c @mileo @JordiBForgeFlow


mileo:
Hi, @sbidoul thanks! 

I ran the tests on the runbot, it seems to be working well. 

http://3408304-256-4348a7.runbot1.odoo-community.org/web#id=4&view_type=form&model=mis.report.instance&active_id=4

However, I don't know if I didn't know how to configure it, but the budget columns didn't work.

Another very interesting feature that we did in the PR of v8, was the possibility for the user, when viewing the report, to navigate to the subreport by clicking on the cell. https://github.com/OCA/mis-builder/pull/153/files#diff-3678f60340076198a6cc144a9fa4a107L37

Can we help you in any way?

Best regards

sbidoul:
> the budget columns didn't work.

Yes, that's expected. That should work after the refactoring.

Regarding the drilldown I'm not sure. Currently the drilldown shows move lines or budget lines. I tend to think showing other reports could be confusing because there could be many of them for different periods, or different preset analytic filters. Maybe something could be possible with the the features that generates reports instances on the fly. Need to think about it.

> Can we help you in any way?

Thanks for proposing! Let me go through the refactoring first.

sbidoul:
@mileo in the end, budget with subreports is not that obvious. Budget by account will work though.
For budget by KPIs to work with subreports, we probably need to extend `mis.budget` to either create a budget that allows budgeting KPIs from all subreports. Or make the budget source a m2m so the user can provide a budget for each subreport when defining a budget column. And that probably requires finding a good design that allows combining budget by account and budget by KPI in a single report. So that definitely requires more design thinking before doing a PR.

So this specific PR is done :)

mis_builder/models/mis_report.py
```diff
@@ -470,6 +470,13 @@ def _default_move_lines_source(self):
         compute="_compute_account_model", string="Account model"
     )
 
+    @api.depends("kpi_ids", "subreport_ids")
+    def _compute_all_kpi_ids(self):
+        for rec in self:
+            rec.all_kpi_ids = rec.mapped("kpi_ids") | rec.subreport_ids.mapped(
```
MiquelRForgeFlow:
```suggestion
            rec.all_kpi_ids = rec.kpi_ids | rec.subreport_ids.mapped(
```

fclementic2c (approved):
I tested a reports fetching KPIs from 3 others subreports : test ok
Fantastic :)

mis_builder/models/mis_report.py
sbidoul: @MiquelRForgeFlow thanks!

sbidoul:
Merging into the 3.6 staging branch. People interested to test further can do so in #254 

/ocabot merge

PR #267, opened 2020-03-12
[12.0] staging for version 3.6

sbidoul:
Migration of #266


rvalyi:
Hello, @sbidoul 
thanks for this update. I'm not sure how your analytic filter works exactly, but what I know is that sometimes in a balanced double accounting move, the analytic account is carried on the other leg from the account.move.line you want to filter (I had this issue when selling projects). Do you see some workaround for such cases?

sbidoul:
@rvalyi not sure what you mean. Does not seem related to this PR though. Can you open an issue with an example?

fclementic2c:
LGTM
:+1:

PR #270, opened 2020-03-19
[10.0] Switch order of computes to allow comparisons of sum columns

svalaeys:
This is my use case. 

A user is using a particular budget for 6 month of the year, and another budget for the other 6 months. We have a column for each in the report, and a column to tally them up and have a full-year budget. Then we compare this total column to the realized of the year.

Right now, the sum column cannot be used as part of a comparison. It errors out on the following line of the kpimatrix.py:

 `    def compute_comparisons(self):
        """ Compute comparisons.

        Invoke this after setting all values.
        """
        for (
            cmpcol_key,
            (col_key, base_col_key, label, description),
        ) in self._comparison_todo.items():
            col = self._cols[col_key]
            base_col = self._cols[base_col_key]
            common_subkpis = self._common_subkpis([col, base_col])`

because base_col  is None at this point.

However, with the proposed fix, sum columns are prepared before the compute_comparison method runs and it works fine.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -791,8 +791,8 @@ class MisReportInstance(models.Model):
                 date_to = self._format_date(period.date_to)
                 description = _("from %s to %s") % (date_from, date_to)
             self._add_column(aep, kpi_matrix, period, period.name, description)
-        kpi_matrix.compute_comparisons()
         kpi_matrix.compute_sums()
+        kpi_matrix.compute_comparisons()
         return kpi_matrix
 
     def compute(self):


svalaeys:
After updating mis_builder to its latest version, this pull request is now neither up-to-date nor needed. I am closing it. Thanks for everything.

sbidoul:
@svalaeys thanks for the feedback. That is surprising because nothing has changed in that area.

Re-reading your use case, note that you create a budget for the full year with lines for different periods in the same budget. Then you can create 3 columns using the same budget S1, S2, Total Year without using sum columns.

svalaeys:
Mmmm, that’s worrisome 😉 I was wondering if I had tested this enough (for reasons only due to my lack of discipline, bringing in the new code was quite a feat for me). I will test some more before I go forward, but it's working fine. Maybe it was always working fine and I had messed up something in the previous version (I am gifted like that!)

Regarding your comment on my use case, in my case I am trying to combine different sources. For example, I doing a projection for the full year using the first six months of realized data (one column), six month of budget data (another column), summing both to have the full year projection (a third column) and comparing that to initial budget. 

Either way I'll drop you a line a bit later to confirm, and open another issue for something I want to discuss with you.

PR #272, opened 2020-03-24
[13.0][ADD] mis_builder_budget_tier validation

misern2:
##  Install

This module depends on ``base_tier_validation``. You can find it at  [OCA/server-ux](https://github.com/OCA/server-ux)

This module extends the functionality of Mis Budgets to support a tier validation process.

## Configuration
To configure this module, you need to:  

- Go to *Settings > Technical > Tier Validations > Tier Definition*.
-    Create as many tiers as you want for Mis Budget model.

## Usage
To use this module, you need to:

* Create a Mis Budget triggering at least one "Tier Definition".
* Click on *Request Validation* button.
* Under the tab *Reviews* have a look to pending reviews and their statuses.
* Once all reviews are validated button to *Confirm* will be shown.

Additional features:

* You can filter the Mis Budget requesting your review through the filter *Needs my
  Review*.
* User with rights to confirm the Mis Budget (validate all tiers that would
  be generated) can directly do the operation, this is, there is no need for
  her/him to request a validation.


HaraldPanten (approved):
Hi @misern2 , 

I tested it in a local environment and it seems to work well. Functional review 👍

sbidoul:
Interesting feature, thanks! Could you please move the PR to https://github.com/OCA/mis-builder-contrib ?

misern2:
@sbidoul done!.

PR #274, opened 2020-04-01
[12.0] fix currency rate in uninvoiced purchases

amcor:
cc @guadaltech
The currency rate  no apply in the result of the uninvoiced purchases


sbidoul:
@amcor thanks for this fix!

/ocabot merge patch

PR #279, opened 2020-04-20
[FIX] Inexisting security group linked to 'Show settings' function in MIS widget

MaxyMoos:
## Description

Super small change to enable the "Settings" button when previewing reports, which is displayed automatically if the user has the "Show full accounting features" right.

It seems like this is already fixed in the 10.0 branch as it already uses the `account.group_account_user` ref.

### DIFF
--- a/mis_builder/static/src/js/mis_report_widget.js
+++ b/mis_builder/static/src/js/mis_report_widget.js
@@ -114,7 +114,7 @@ odoo.define("mis_builder.widget", function(require) {
                 });
 
             var def2 = session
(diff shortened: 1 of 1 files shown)


sbidoul:
@MaxyMoos good catch. Would you mind doing the PR to 11.0 and add a file `named <pr number>.bugfix` with short changelog entry explaining the fix? I'll take care of forward porting.

MaxyMoos:
@sbidoul Sure thing, I'm closing this PR and will open the 11.0 one with the changelog addition

PR #281, opened 2020-04-21
[FIX] Non-existing security group linked to "Settings" button in MIS Widget

MaxyMoos:
## Description

Super small change to enable the "Settings" button when previewing reports, which is displayed automatically if the user has the "Show full accounting features" right.

It seems like this is already fixed in the 10.0 branch as it already uses the account.group_account_user ref.


sbidoul (approved):
Thanks! I'll merge and forward port later this week.

Issue #262, opened 2020-02-25
Support more n-to-date options


sbidoul:
We currently support Year-To-Date, which is from 1/1 of the year of the selected period until the last day of the selected period.

There is demand for more options: Month-To-Date, Fiscal Year-To-Date, etc.

So maybe we should generalize the ytd flag with some sort of start period? Not sure about the best UI.

svalaeys:
For consistency's sake, I propose that both start date and end date be configured in the same way, independently from one another, with the following two options:
- A determined date, possibly adjusted 
- The pivot date, possibly adjusted

I recently needed Fiscal-Year-To-Date periods. I went back and forth for a while about transforming the ytd_flag to a selection field that would allow the user to choose between calendar year-to-date and fiscal year-to-date. In the end, I couldn't think of a single use-case where some company which had a fiscal year that did not overlap with calendar year would also need dynamic date handling for a calendar year, or, at least, a use-case strong enough that would be worth complicating the UI.

I did however notice that I would eventually run into issues the day I would consolidate data between companies that didn't have the same fiscal year dates. We could of course have a dropdown to let the user choose the company fiscal year to consider, but it seems like again it would be complicating the UI. At the end of the day, the user knows which date he's looking for.

In my particular case, I was trying to configure a report so that the dates of its ten columns would adjust based only on the pivot date. Simplifying, columns where of two types : realized from fiscal year start to pivot date, and budget columns from pivot date + 1 to fiscal year end. So far, I have not found out how to do the second part with the current setup.

With my proposal, the configuration for the report I just described would therefore end up as follows, with real dates to simplify the example:

Column type 1
Source: Realized
Date From: Determined, 01/05/2029, not adjusted
Date To: Pivot Date, not adjusted

Column type 2:
Source: Budget
Date From: Pivot Date, Adjusted +1 Day, Offset 0
Date To : Determined, 30.04.2020, not adjusted

Every month, solely changing the pivot date would take care of everything. 

I don't know if my train of thought has covered all the possible scenarios, but to me it seems like the most consistent and simple UI to cover all the cases.

PR #292, opened 2020-04-23
Restore _intersect_days (as a model method)

### LINKED ISSUE #280: Restore _intersect_days model method
Hi Stephane, 
Thank you very much for all the latest changes in mis_builder and mis_builder_budget. Budgets by accounts, filters on analytics, everything I ever wanted. Thanks.
My question has to do with the method intersect_days in mis_kpi_data.py.

This method used to be the class method _intersect_days of MisKpiData. I had overwritten this method to achieve a different calculation of the pro-rata part of the budget item for a column. The reason is that there is no doubt that a great selling point of MIS Builder Budget is the automated pro-rated calculation of budget items. But for all my clients, they want a budget split equally in twelve months. This makes a lot of sense for rents, payroll, etc. 

Correct me if I am wrong, but I think the only way to achieve this in base is to create 12 different budget items each worth a month. But by doing so it takes away the clean, one-budget-item-per-year advantage of MIS Budget.

The override of _intersect_days took care of that for me. It was full of assumptions (30-day months) and inflexibility (a report column must fully enclose a period to take it into account) but it did its job for the most common cases I am encountering.

Now that it's not a class method anymore, it's a little less clean to override it. Therefore I have two questions:
- How do think about non pure day-based pro rata temporis needs? Do you systematically create multiple lines in those cases?
- Would you consider putting intersect_days back as a class method to ease its override?

Thanks for your thoughts.   

Stephane

sbidoul:
Semi supported to allow overrides.

fixes #280 

@svalaeys

### DIFF
--- a/mis_builder/models/mis_kpi_data.py
+++ b/mis_builder/models/mis_kpi_data.py
@@ -71,6 +71,10 @@ class MisKpiData(models.AbstractModel):
             )
 
     @api.model
+    def _intersect_days(self, item_dt_from, item_dt_to, dt_from, dt_to):
+        return intersect_days(item_dt_from, item_dt_to, dt_from, dt_to)
+
+    @api.model
     def _query_kpi_data(self, date_from, date_to, base_domain):
         """Query mis.kpi.data over a time period.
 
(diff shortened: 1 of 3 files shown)


svalaeys:
You are gentleman and a scholar. 

Let me know if you want to discuss further topics brought up in #280 and #262 to figure out how @MaxyMoos and I can help.

PR #296, opened 2020-04-24
 fix read_group fields arguments

2zx:
renamed the variable "flds" in read_group arguments to allow the execution of the method as per standard grouping on tree view.
The renaming involved the use of an alias for the fields module to guarantee the functionality of the method in the prorata.read_group.mixin

### DIFF
--- a/mis_builder/models/prorata_read_group_mixin.py
+++ b/mis_builder/models/prorata_read_group_mixin.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.fields import Date
 from odoo.exceptions import UserError
 
 from .mis_kpi_data import intersect_days
@@ -38,7 +39,7 @@ class ProRataReadGroupMixin(models.AbstractModel):
 
     @api.model
     def read_group(
-        self, domain, flds, groupby, offset=0, limit=None, orderby=False, lazy=True
+        self, domain, fields, groupby, offset=0, limit=None, orderby=False, lazy=True
     ):
         """Override read_group to perform pro-rata temporis adjustments.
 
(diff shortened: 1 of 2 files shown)


mis_builder/models/prorata_read_group_mixin.py
```diff
@@ -60,30 +60,30 @@ def read_group(
...
             for item in self.search(domain).read(read_fields):
                 key = tuple(item[k] for k in groupby)
                 if key not in res:
                     res[key] = {k: item[k] for k in groupby}
                     res[key].update({k: 0.0 for k in sum_fields})
                 res_item = res[key]
                 for sum_field in sum_fields:
-                    item_dt_from = fields.Date.from_string(item["date_from"])
-                    item_dt_to = fields.Date.from_string(item["date_to"])
+                    item_dt_from = ffields.Date.from_string(item["date_from"])
```
sbidoul: Maybe `from odoo.fields import Date` to avoid this `ffields`?

sbidoul (commented):
Ah, good catch. Could you resolve the merge conflict?

mis_builder/models/prorata_read_group_mixin.py
2zx: sure, it sounds better with an import

sbidoul:
Thanks! You still need to rebase to resolve the conflict. Can you also add a file named `296.bugfix` in `mis_builder/readme/newfragments` with a changelog entry? (something like "Fix stack trace when grouping budget by account items in the web client.")

sbidoul:
Almost there :) A newline in the changelog file, and if you don't mind a rebase and squash to have one commit.

2zx:
> Almost there :) A newline in the changelog file, and if you don't mind a rebase and squash to have one commit.

sorry: do you mean you want a single commit for the entire pull request?

sbidoul:
> sorry: do you mean you want a single commit for the entire pull request?

Yes please, since there are now 5 and it's a bit too much for such as simple change.

2zx:
> > sorry: do you mean you want a single commit for the entire pull request?
> 
> Yes please, since there are now 5 and it's a bit too much for such as simple change.

ok, i hope i did it right :-)

sbidoul:
I fixed a small regression in #311, so I close this one. Thanks for the fix!

PR #300, opened 2020-05-08
[FIX] mis_builder: Comparison percentages export format in XLSX

MaxyMoos:
When a divider/factor Style parameter is used on a KPI, it is also applied to the "Compare columns" when exporting to Excel, which leads to errors of magnitude for the percentages displayed in the resulting XLSX file.

Here is a simple report using a KPI with a style factor "1e3 - k" and with a compare column as it is displayed in Odoo :
Screenshot from 2020-05-08 09-27-33

And the resulting XLSX export:
Screenshot from 2020-05-08 09-27-47

Seen on 12.0 but it seems like 10.0 code is similar.

### DIFF
--- a/mis_builder/readme/newsfragments/300.bugfix
+++ b/mis_builder/readme/newsfragments/300.bugfix
@@ -1,0 +1,3 @@
+Having a "Compare columns" added on a KPI with an associated style using a
+Factor/Divider did lead to the said factor being applied on the percentages
+when exporting to XLSX.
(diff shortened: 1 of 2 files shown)


sbidoul (approved):
LGTM, thanks!

HaraldPanten (approved):
I think it should be OK too 👍

sergiocorato:
@sbidoul merge?

PR #304, opened 2020-07-08
Fix drilldown action name

ddufresne:
I click to drilldown on a report line.

image

The action title in the breadcrumb is the technical expression of the report line.

image

The problem is that this expression is not understandable for most users.
It is only understandable for technical users who parameterize the system.

After this PR, when possible, the name of the account is displayed instead of the technical expression.

image

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -697,6 +697,7 @@ class MisReport(models.Model):
                     if not drilldown_arg:
                         continue
                     drilldown_arg["period_id"] = col_key
+                    drilldown_arg["kpi_id"] = kpi.id
 
                 if name_error:
                     recompute_queue.append(kpi)
@@ -763,6 +764,7 @@ class MisReport(models.Model):
                         if not drilldown_arg:
                             continue
                         drilldown_arg["period_id"] = col_key
+                        drilldown_arg["kpi_id"] = kpi.id
                     kpi_matrix.set_values_detail_account(
                         kpi, col_key, account_id, vals, drilldown_args
                     )
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -838,7 +838,7 @@ class MisReportInstance(models.Model):
             )
             domain.extend(period._get_additional_move_line_filter())
             return {
-                "name": u"{} - {}".format(expr, period.name),
+                "name": self._get_drilldown_action_name(arg),
                 "domain": domain,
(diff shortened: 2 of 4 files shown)


sergiocorato (approved):
LGTM

astirpe (approved):

sbidoul:
Thanks for this PR! I have somewhat mixed feelings about it (which is why I did not comment earlier).

On the one hand I totally agree that the drill down action name is ugly. On the other hand, this only fixes it for account detail lines. It remains the same for kpi lines.

Perhaps we should round trip the kpi_id in addition to period_id and account_id, and display something like `{kpi.display_name} - {account.display_name} ({period.display_name})`?

ddufresne:
@sbidoul your suggestion makes sense.

I will work on it.

I might have to add/restructure some code in order to implement it. The reason is that the kpi_id is not passed to the method drilldown.

sbidoul:
@ddufresne try adding kpi_id next to period_id in drilldown_args [here](https://github.com/OCA/mis-builder/blob/d616aed378d155d73e41df5d6b9986f28a3c2eab/mis_builder/models/mis_report.py#L699) and [here](https://github.com/OCA/mis-builder/blob/d616aed378d155d73e41df5d6b9986f28a3c2eab/mis_builder/models/mis_report.py#L765).

ddufresne:
@sbidoul implemented your suggestion. Thanks.

sbidoul:
Cool, LGTM technically. Could you just add a unit test for that new `_get_drilldown_action_name` method?

ddufresne:
@sbidoul done

sbidoul:
Ah, and the last thing would be a file named `304.feature` in mis_builder/readme/newsfragments (a short changelog entry).

ddufresne:
@sbidoul done

marcelsavegnago (approved):
LGTM

qtiplb:
Hi @sbidoul would this be ready to merge?

Thanks!

sbidoul:
> would this be ready to merge?

Yes, absolutely. @ddufresne Cherry on the cake would be squashing the commits.

ddufresne:
@sbidoul squashed the commits. Thank you.

sbidoul:
@ddufresne thanks. Now travis is red, probably due to a recent pre-commit update. Could you re-run pre-commit on the branch and squash again?

ddufresne:
@sbidoul done

sbidoul:
Cherry-picked into https://github.com/OCA/mis-builder/pull/325 (14.0 migration)

PR #325, opened 2020-10-10
[14.0] migration

sbidoul:
dependencies

- [x] `date_range` https://github.com/OCA/server-ux/pull/217
- [x] `report_xlsx` https://github.com/OCA/reporting-engine/pull/443
- [x] `compute_fiscalyear_dates` and the `account.fiscal.year` model was moved to EE in https://github.com/odoo/odoo/pull/46110, annoying. Tracked in  https://github.com/OCA/account-financial-tools/issues/1069 - update: `compute_fiscalyear_dates` has now a default implementation in `account` so we can rely on that

todo

- [x] `mis_builder_demo`
- [ ] analytic tag filters are broken but miss test coverage (resolve_2many_commands is gone)
- [ ] search for `TODO 14` in the code
- [x] pdf report seems broken
- [ ] safe_eval issue (https://github.com/OCA/mis-builder/pull/325#issuecomment-760074430)
- [x] some failing AEP tests


sbidoul:
This kinda sorta works. Now I have talks to prepare...

tupaq:
report_xlsx is migrated. Check [OCA/reporting-engine#441](https://github.com/OCA/reporting-engine/pull/441)

HaraldPanten:
@sbidoul I have a question about mis. This one is the migration PR, right? so... What about the mis_builder that you've in 14.0 branch? THX for clarifying! 👍

sbidoul:
@HaraldPanten on the 14.0 branch it has `installable=False` (https://github.com/OCA/mis-builder/tree/14.0#unported-addons).

sbidoul:
@gurneyalex with the last commit (py 3.6 + py 3.8) builds, runbot says the build is skipped and the full log says 404. Do you know if this is related to the py 3.8 build (it normally should not be because travis2docker uses the first job IIUC) or otherwise give a hint on how I can diagnose this ?

HaraldPanten:
Hmmm I see. Thx!

kittiu:
@sbidoul may I know what is your plan for migration. This is PR still valid?
Thanks.

sbidoul:
This PR should work. I will rebase. IIRC there was an issue with PDF reports.

In any case all commits here are good for review (except the few that are CI experiments that I will remove) and I'll very much welcome reviews and tests :)

Saran440:
Functional Test.
Error when used Queries Domain
Step to error:
1. create template and query data from sale.order
Selection_001

2. sum query to KPI
Selection_002

3. view report, it throw error.
Selection_003



Maybe time is not input in safe_eval
https://github.com/OCA/mis-builder/pull/325/files#diff-919b98eeccf1b8c1b6c401a20043278ca730430263ae36b50fdecd38a38ac60eR586

sbidoul:
@Saran440 thanks for testing. Interesting find. We'll definitely need a new unit test that reveals that issue.

Reyes4711-S73:
Hi @sbidoul. I found one issue when I was generating one report. The function render_qweb_pdf changes in odoo to _render_qweb_pdf.

sbidoul:
@Reyes4711 thanks! I added a commit with that. PDF reports do work now.

sbidoul:
Interestingly the failing AEP test fixed itself automagically. Probably following a change in Odoo...

mis_builder/report/mis_report_instance_qweb.py
```diff
@@ -22,7 +22,7 @@ def render_qweb_pdf(self, res_ids=None, data=None):
             )
             # data=None, because it was there only to force Odoo
             # to propagate context
-            return super(Report, self.with_context(context)).render_qweb_pdf(
+            return super(Report, self.with_context(context))._render_qweb_pdf(
                 res_ids, data=None
             )
         return super(Report, self).render_qweb_pdf(res_ids, data)
```
hildickethan: Missed one!
```suggestion
            return super(Report, self.with_context(context))._render_qweb_pdf(
                res_ids, data=None
            )
        return super(Report, self)._render_qweb_pdf(res_ids, data)
```

sbidoul: Gah, done. Thanks!

dreispt:
@sbidoul IMO this is best merged, even if there are items to address.
The core functionality seems to work well.
You could downgrade to Beta maturity, if you like.

sbidoul:
Indeed, good suggestion @dreispt. I created #352 and #353 to track the two known bugs.

/ocabot merge patch

HaraldPanten:
> Indeed, good suggestion @dreispt. I created #352 and #353 to track the two known bugs.
> 
> /ocabot merge patch

Good job @sbidoul !

PR #327, opened 2020-10-27
[13.0][FIX] Compatibility with multicompany

JordiBForgeFlow:
- The mis.report.instance should allow to define as many
  companies as the user is allowed to operate on when logged.
- The company is only a required field if the flag 'multi_company' is not set.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -499,7 +499,7 @@ class MisReportInstance(models.Model):
         comodel_name="res.company",
         string="Company",
         default=lambda self: self.env.company,
-        required=True,
+        required=False,
     )
     multi_company = fields.Boolean(
         string="Multiple companies",
@@ -511,6 +511,7 @@ class MisReportInstance(models.Model):
         comodel_name="res.company",
         string="Companies",
         help="Select companies for which data will be searched.",
+        domain=lambda self: [("id", "in", self.env.companies.ids)],
     )
     query_company_ids = fields.Many2many(
         comodel_name="res.company",
@@ -544,12 +545,10 @@ class MisReportInstance(models.Model):
     )
     hide_analytic_filters = fields.Boolean(default=True)
 
-    @api.onchange("company_id", "multi_company")
+    @api.onchange("multi_company")
     def _onchange_company(self):
-        if self.company_id and self.multi_company:
-            self.company_ids = self.env["res.company"].search(
-                [("id", "child_of", self.company_id.id)]
-            )
+        if self.multi_company:
+            self.company_id = False
         else:
             self.company_ids = False
 
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -504,7 +504,7 @@ class MisReportInstance(models.Model):
     multi_company = fields.Boolean(
         string="Multiple companies",
         help="Check if you wish to specify "
-        "children companies to be searched for data.",
+        "several companies to be searched for data.",
         default=False,
     )
     company_ids = fields.Many2many(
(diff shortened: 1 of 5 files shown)


mis_builder/security/mis_builder_security.xml
```diff
@@ -4,7 +4,7 @@
         <field name="name">Mis Report Instance multi company</field>
         <field name="model_id" ref="model_mis_report_instance" />
         <field name="domain_force">
-            ['|',('company_id','=',False),('company_id','in',company_ids)]
+            ['|',('company_id','=',False),('company_id','in',company_ids), '|', ('company_ids', '=', False), ('company_ids', 'in', company_ids)]
```
MiquelRForgeFlow:
```suggestion
            ['&amp;', '|',('company_id','=',False),('company_id','in',company_ids), '|', ('company_ids', '=', False), ('company_ids', 'in', company_ids)]
```

and run pre-commit

LoisRForgeFlow:
To ease the review, I will showcase a bit the changes:

1. When the companies field is empty, the same report can be used for all companies, and you just need to activate more companies if you want to aggregate:
Base report instance:
image

If we open this report with only one company active we get the result for each one:
image
image
image

Then we can activate the second and third company to get the aggregated result:
image

2. Regarding `_onchange_company` the new behavior can be observer in the gif below:
Peek 2021-01-14 15-27

LoisRForgeFlow:
@sbidoul What do you think? does this make sense to you?

sbidoul:
> To ease the review, I will showcase a bit the changes:

Thanks a lot for this! I'll look into it as soon as possible.

LoisRForgeFlow:
@sbidoul Any chance you can consider this changes? Thanks!

sbidoul:
@LoisRForgeFlow yeah I absolutely need to find an hour to focus on this. I'll keep it on the topmost part of my stack.

LoisRForgeFlow:
@sbidoul Hi, just a friendly reminder. It would be great if you can give your opinion on this!

sbidoul:
I've reserved some time for MIS builder maintenance mid august. I'll get to it then. Sorry again for the delay.

mis_builder/models/mis_report_instance.py
```diff
@@ -544,20 +545,31 @@ def _compute_pivot_date(self):
...
+            self.company_id = company
             self.company_ids = False
 
     @api.depends("multi_company", "company_id", "company_ids")
     def _compute_query_company_ids(self):
         for rec in self:
             if rec.multi_company:
-                rec.query_company_ids = rec.company_ids or rec.company_id
+                # If no companies defined use active companies.
+                rec.query_company_ids = rec.company_ids or self.env.companies
```
sbidoul, line 572: Would it be better here to do the intersection of rec.company_ids and self.env.companies. Because in the report we print the list of companies used in the report and we should not print companies to which the user does not have access and therefore ignored by the record rules.

mis_builder/views/mis_report_instance.xml
```diff
@@ -163,6 +163,7 @@
                                 <field
                                     name="company_id"
                                     groups="base.group_multi_company"
+                                    attrs="{'required': [('multi_company', '=', False)]}"
```
sbidoul, line 166: Also, invisible if multi_company is True ?

mis_builder/models/mis_report_instance.py
```diff
@@ -544,20 +545,31 @@ def _compute_pivot_date(self):
     )
     hide_analytic_filters = fields.Boolean(default=True)
 
-    @api.onchange("company_id", "multi_company")
+    @api.onchange("multi_company")
```
sbidoul, line 548: Shall we add a constraint that 
- company_id is not set when multi_company is True and 
- company_id is set when multi_company if False

mis_builder/views/mis_report_instance.xml
sbidoul: For a better UX, let's take the opportunity to move the multi_company checkbox before the company_id field.

mis_builder/models/mis_report_instance.py
```diff
@@ -499,18 +499,19 @@ def _compute_pivot_date(self):
...
         string="Multiple companies",
         help="Check if you wish to specify "
-        "children companies to be searched for data.",
+        "several companies to be searched for data.",
         default=False,
     )
     company_ids = fields.Many2many(
         comodel_name="res.company",
         string="Companies",
         help="Select companies for which data will be searched.",
```
sbidoul, line 513:
```suggestion
        help="Select companies for which data will be searched (search all companies if empty).",
```

mis_builder/models/mis_report_instance.py
```diff
@@ -499,18 +499,19 @@ def _compute_pivot_date(self):
...
+        domain=lambda self: [("id", "in", self.env.companies.ids)],
```
sbidoul, line 514: I hesitate about this domain.

This domain hides companies that are not in the user context. In a way this is good because this is what the report will filter on. But OTOH it would be good to show the user that the report can work also on other companies if they select them in their context.

mis_builder/models/mis_report_instance.py
sbidoul: This constraint is not necessary if the company_id field is hidden in single company mode.

mis_builder/models/mis_report_instance.py
sbidoul: I'll remove this domain and add `query_company_ids` to the view as `Effective companies`.

sbidoul:
So I implemented my suggestions in #374  - apologies again for the late review.

sbidoul:
Superseded by #374

PR #331, opened 2020-10-31
[12.0][FIX] mis_builder: arrays in domain

alexey-pelykh:
To support expressions like `balp[][('something', 'in', ['a', 'b'])]`

cc @pedrobaeza @sbidoul

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -1,4 +1,5 @@
 # Copyright 2014 ACSONE SA/NV (<http://acsone.eu>)
+# Copyright 2020 CorporateHub (https://corporatehub.eu)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import datetime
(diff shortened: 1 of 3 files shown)


mis_builder/models/aep.py
```diff
@@ -88,7 +89,7 @@ class AccountingExpressionProcessor(object):
         r"\s*"
         r"(?P<account_sel>_[a-zA-Z0-9]+|\[.*?\])"
         r"\s*"
-        r"(?P<ml_domain>\[.*?\])?"
+        r"(?P<ml_domain>\[.*\])?"
```
sbidoul, line 92: One thing that might then be slightly annoying is that the allowed syntax is not the same for the account and move line domains. To make it fully correct we'd need something more powerful than a regex.

mis_builder/models/aep.py
```diff
@@ -169,7 +170,14 @@ def _parse_match_object(self, mo):
         # move line domain
         if ml_domain:
             assert ml_domain[0] == "[" and ml_domain[-1] == "]"
-            ml_domain = tuple(safe_eval(ml_domain, domain_eval_context))
+            components = safe_eval(ml_domain, domain_eval_context)
+            ml_domain = []
+            for component in components:
+                left, op, right = component
+                if isinstance(right, list):
+                    right = tuple(right)
```
sbidoul, line 178: Is there not an Odoo function somewhere that implement that kind of logic?

alexey-pelykh: I also thought I've seen one, but haven't found any actually 🤷‍♂️

mis_builder/models/aep.py
alexey-pelykh: True, but that would be much of a bigger change

PR #332, opened 2020-10-31
[IMP] mis_builder: remove field size limit

alexey-pelykh:
The size limits imposed on fields don't add any value

cc @pedrobaeza @sbidoul


sbidoul:
I guess this limit was made so the user does not create variable names that are really too long to use in expression.
But yeah, that probably does not add much value.

PR #339, opened 2020-11-17
[14.0][IMP] - Allow budget items overlap

sbejaoui:


mis_builder_budget/models/mis_budget_by_account_item.py
```diff
@@ -60,6 +60,8 @@ def _compute_balance(self):
 
     def _prepare_overlap_domain(self):
         """Prepare a domain to check for overlapping budget items."""
+        if self.budget_id.allow_items_overlap:
+            return [("id", "=", False)]
```
sbidoul, line 64: Is it not better to not call `_prepare_overlap_domain` in `_check_dates` when `allow_items_overlap` is set ?

sbidoul:
Superseded by #384 384

PR #342, opened 2021-01-14
[WIP][14.0][ADD] mis_builder_budget_activity_group (for government sector, experimental)

Saran440:
Normally, MIS Builder will point at account code but we implement module budget_activity for help users mapping with account code. then account code is not enough.
AG

On activity can select activity group (many : 1)
On activity can select account code (1 : 1) but Maybe 1 account code has activity more than 1 and difference with group
then budget report are wrong value.

This module budget reporting activity instead of account
i.e. 
Users select activity A2 (5200020) on Bills. Budgeting will increase commitment with Activity Group 1, 2, 3
it will check activity and input commitment to Activity Group 1 only.


Saran440:
we will move on to account-budgeting

PR #343, opened 2021-01-14
[13.0][FIX] mis_builder: copying reports with comparison columns.

LoisRForgeFlow:
While duplicating a MIS report instance, comparison columns are
ignored because they would raise an error, as they keep the old
`source_cmpcol_from_id` and `source_cmpcol_to_id` from the
original record.

Error that raises without this fix is the following:
image

@ForgeFlow

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -459,6 +459,17 @@ class MisReportInstancePeriod(models.Model):
                         % rec.name
                     )
 
+    def copy_data(self, default=None):
+        if self.source == SRC_CMPCOL:
+            # While duplicating a MIS report instance, comparison columns are
+            # ignored because they would raise an error, as they keep the old
+            # `source_cmpcol_from_id` and `source_cmpcol_to_id` from the
+            # original record.
+            return [
+                False,
+            ]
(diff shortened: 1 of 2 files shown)


mis_builder/models/mis_report_instance.py
```diff
@@ -459,6 +459,18 @@ def _check_source_cmpcol(self):
                         % rec.name
                     )
 
+    @api.returns(None, lambda value: value[0])
```
sbidoul: The decorator is not necessary, it is inherited automatically (according to its documentation and examples in the Odoo codebase at least).

sbidoul (commented):
Thanks! A minor remark. Can you also add a file `newsfragments/343.bugfix` with the changelog entry (the description you wrote in the PR is just fine for that).

mis_builder/models/mis_report_instance.py
LoisRForgeFlow: Let's trust Odoo this time :)

LoisRForgeFlow:
@sbidoul newsfragment added. Could you review it?

sbidoul:
/ocabot merge patch

Thanks!

PR #344, opened 2021-01-28
[13.0][FIX] mis_builder: always ignore cancel entries.

LoisRForgeFlow:
When you select all entries in "Target Moves", the expected is
that the report considers both posted and draft entries.

In v13 and due to the merge of invoice in `account.move` model,
a new state `cancel` has been introduced. This forces to always
filter entries in this state to keep the behavior of previous
versions (and expected).

@ForgeFlow

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -294,6 +294,8 @@ class AccountingExpressionProcessor(object):
             ]
         if target_move == "posted":
             domain.append(("move_id.state", "=", "posted"))
+        else:
+            domain.append(("move_id.state", "in", ("posted", "draft")))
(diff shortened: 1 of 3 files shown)


HaraldPanten (approved):

sbidoul:
@LoisRForgeFlow Travis is red, can you check? I think it is good opportunity to use `parent_state` instead of `move_id.state`.

sbidoul:
Also, the test failure shows this is going to be a breaking change because alternative move line views will now have to have the `parent_state` field... ah well, sigh, I can't wait until v17 when stock pickings will be account moves too.

sbidoul:
superseded by #377

PR #350, opened 2021-03-11
[10.0][FIX] Use account_model to get drill down name

hhgabelgaard:
In the same way as it is done in drilldown()

Then It will work with altermate source


sbidoul:
I added the changelog entry.

/ocabot merge patch

PR #361, opened 2021-05-11
[14.0][FIX] mis_builder: compute comparison domain correctly

joao-p-marques:
When on a MIS Report Instance, if you wanted to generate a new line of type comparison, you couldn't currently select any existing period to compare

This happened because the field domain was searching in a `NewId` context, thus not finding a correct period.

Changing the domain and making it use a computed field with a search for the `_origin` record solves the problem

@Tecnativa
TT29316

ping @pedrobaeza @victoralmau

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -274,6 +274,9 @@ class MisReportInstancePeriod(models.Model):
     source_cmpcol_to_id = fields.Many2one(
         comodel_name="mis.report.instance.period", string="Compare"
     )
+    allowed_cmpcol_ids = fields.Many2many(
+        comodel_name="mis.report.instance.period", compute="_compute_allowed_cmpcol_ids"
+    )
     # filters
     analytic_account_id = fields.Many2one(
         comodel_name="account.analytic.account",
@@ -318,6 +321,12 @@ class MisReportInstancePeriod(models.Model):
             "Period name should be unique by report",
         ),
     ]
+
+    @api.depends("report_instance_id")
+    def _compute_allowed_cmpcol_ids(self):
+        """Compute actual records while in NewId context"""
+        for record in self:
+            record.allowed_cmpcol_ids = record.report_instance_id.period_ids - record
 
     @api.constrains("source_aml_model_id")
     def _check_source_aml_model_id(self):
--- a/mis_builder/views/mis_report_instance.xml
+++ b/mis_builder/views/mis_report_instance.xml
@@ -335,16 +335,17 @@
                             name="source_sumcol_accdet"
                             attrs="{'invisible': [('source', '!=', 'sumcol')]}"
                         />
+                        <field name="allowed_cmpcol_ids" invisible="1" />
                         <field
                             name="source_cmpcol_to_id"
                             attrs="{'invisible': [('source', '!=', 'cmpcol')], 'required': [('source', '=', 'cmpcol')]}"
-                            domain="[('report_instance_id', '=', report_instance_id), ('id', '!=', id)]"
+                            domain="[('id', 'in', allowed_cmpcol_ids)]"
                             options="{'no_create': True, 'no_open': True}"
                         />
                         <field
                             name="source_cmpcol_from_id"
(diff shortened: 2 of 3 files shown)


victoralmau (approved):

mis_builder/models/mis_report_instance.py
```diff
@@ -319,6 +322,12 @@ def _compute_dates(self):
         ),
     ]
 
+    @api.depends("report_instance_id")
+    def _compute_allowed_cmpcol_ids(self):
+        """Compute actual records while in NewId context"""
+        for record in self:
+            record.allowed_cmpcol_ids = record.report_instance_id.period_ids._origin
```
pedrobaeza: `_origin` contains the values in `NewId` mode?

mis_builder/views/mis_report_instance.xml
```diff
@@ -335,16 +335,17 @@
                             name="source_sumcol_accdet"
                             attrs="{'invisible': [('source', '!=', 'sumcol')]}"
                         />
+                        <field name="allowed_cmpcol_ids" invisible="1" />
                         <field
                             name="source_cmpcol_to_id"
                             attrs="{'invisible': [('source', '!=', 'cmpcol')], 'required': [('source', '=', 'cmpcol')]}"
-                            domain="[('report_instance_id', '=', report_instance_id), ('id', '!=', id)]"
+                            domain="[('id', 'in', allowed_cmpcol_ids), ('id', '!=', id)]"
```
pedrobaeza: Remove current one in the compute instead.

mis_builder/models/mis_report_instance.py
joao-p-marques: At least that is what I understand by the documentation: https://github.com/odoo/odoo/blob/14.0/odoo/models.py#L5511
And other references: https://github.com/odoo/odoo/pull/49701#issue-405014807

In this case, `record.report_instance_id.period_ids` will have a set of NewId records like
```python
mis.report.instance.period(<NewId origin=2>, <NewId origin=3>, <NewId 0x7fdd275c5820>)
```
where the first 2 refer to existing columns, and the 3rd to the new one being created.
When we access `_origin`, we get:
```python
mis.report.instance.period(2, 3)
```
which points only to the existing records, skipping the one that is being created.

mis_builder/views/mis_report_instance.xml
joao-p-marques: Thanks, change done :+1:

mis_builder/models/mis_report_instance.py
pedrobaeza: And why skip them? You can do an assignation of `record.report_instance_id.period_ids` (removing itself), and it should work also for virtual ones.

joao-p-marques: Do you mean what I do now in the compute? https://github.com/OCA/mis-builder/pull/361/files#diff-0f911a0342a8e73e0643be069c2eef52c082d983131fc6fa055428b49e17b353R330
Or do you mean doing this directly on the field's domain as before?

pedrobaeza: I mean that the compute to be:
```python
record.allowed_cmpcol_ids = record.report_instance_id.period_ids - record
```

joao-p-marques: Thanks, you are correct :+1: This works, so I just left it as you suggested

pedrobaeza (approved):

pedrobaeza:
@sbidoul this is a v14 specific thing that works on previous versions the other way, but not for this one. Can you please check?

sbidoul:
LGTM, thanks! I've not tested yet but I'll try to do soon.

Is that something that needs to be reported to Odoo as a regression ? Or was it unsupported and working by chance before ?

It would be great if you could add a readme/newsfragment/361.bugfix file with the short description of the fix.

pedrobaeza:
I think that the domain trick working in previous versions was a side effect, as domains don't tend to be conceived for restricting data inside same record (ORM-inception, hehe), but maybe @rco-odoo can give us a hint about that.

joao-p-marques:
Thanks @sbidoul 

I added the bugfix message :+1:

sbidoul:
/ocabot merge patch

Thank you!

PR #377, opened 2021-08-14
[13.0] ignore moves in 'cancel' state

### LINKED ISSUE #376: Account moves in 'cancel' state must be ignored in Odoo 13+ when reporting on posted and draft moves

sbidoul:
supersedes #344 
closes #376

Use the `parent_state` field of `account.move.line` to filter entries in `posted` and `draft` state only. Before, when reporting in draft mode, all entries were used (i.e. there was no filter), and that started included the cancelled entries in Odoo 13.+.

This PR also contains a breaking change in the internal API. For quite a while the `target_move` argument of AEP and other methods was not used by MIS Builder itself and was kept for backward compatibility. To avoid rippling effects of the necessary change, we now remove this argument.

TODO
- [x] add a test for target_move=all

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -26,7 +26,7 @@ _DOMAIN_START_RE = re.compile(r"\(|(['\"])[!&|]\1")
 
 
 def _is_domain(s):
-    """ Test if a string looks like an Odoo domain """
+    """Test if a string looks like an Odoo domain"""
     return _DOMAIN_START_RE.match(s)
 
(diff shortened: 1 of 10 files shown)


sbidoul:
@LoisRForgeFlow you may want to look at this one too.

LoisRForgeFlow (approved):
Tested, it works as expected. Thanks!

sbidoul:
Thanks for testing @LoisRForgeFlow !

PR #388, opened 2021-10-04
[12.0] FIX mis_builder when user does not have time zone set.

eLBati:
Steps:

 - Install mis_builder
 - Empty time zone field of your user, if present
 - Create a MIS report template, following for example https://oca-mis-builder.readthedocs.io/en/latest/usage.html#reporting-on-non-accounting-data-queries

Get
`AttributeError: 'bool' object has no attribute 'upper'`

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -601,10 +601,9 @@ class MisReport(models.Model):
                     ]
                 )
             else:
-                datetime_from = _utc_midnight(date_from, self._context.get("tz", "UTC"))
-                datetime_to = _utc_midnight(
-                    date_to, self._context.get("tz", "UTC"), add_day=1
-                )
+                tz = str(self.env["ir.fields.converter"]._input_tz())
+                datetime_from = _utc_midnight(date_from, tz)
+                datetime_to = _utc_midnight(date_to, tz, add_day=1)
                 domain.extend(
                     [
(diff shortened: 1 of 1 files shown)


mis_builder/models/mis_report.py
```diff
@@ -601,10 +601,9 @@ def _fetch_queries(self, date_from, date_to, get_additional_query_filter=None):
                     ]
                 )
             else:
-                datetime_from = _utc_midnight(date_from, self._context.get("tz", "UTC"))
-                datetime_to = _utc_midnight(
-                    date_to, self._context.get("tz", "UTC"), add_day=1
-                )
+                tz = self._context.get("tz", "UTC") or "UTC"
```
sbidoul: Do you know where Odoo implies that no timezone means UTC ?

eLBati: This https://github.com/odoo/odoo/blob/91d6598221ac50c839df749a2020846d89056daa/odoo/addons/base/models/ir_fields.py#L231 is an example

sbidoul: Oh, thanks that's interesting. Maybe we could then use that `_input_tz` function here too, to be consistent with Odoo does ?

eLBati: Right, done, thanks

sbidoul:
I finished this with a newfragment in #450

PR #395, opened 2021-12-18
[13.0][FIX] mis_builder: Remove old migration scripts

pedrobaeza:
One of my migrations has crashed, and when I discover that it was due to this old migration scripts still being here, I was very astonished.

@Tecnativa


pedrobaeza:
Uhm, the pre-commit thing seems to be something outsider.

victoralmau (approved):

sbidoul:
Why does this script make your migration fail ?

/ocabot merge nobump

pedrobaeza:
Sorry, I didn't keep the log. I was surprised that the script was triggered, but I get a clean migration after removing these scripts, specially since I'm coming from version 8. My guess is that a renaming from old Spanish module `account_balance_reporting` with a lower version to `mis_builder` with current version has triggered these migration scripts when applying on v13.

Issue #392, opened 2021-12-05
Expand by AA


hitrosol:
Having MIS Budgets (by Accounts) is a great feature, so we can expand or collapse the Accounts Details by set the Disable account details expansion. 

It would be greater if we have the expansion feature by AA, so we don't have to make specific AA as KPI. This will help user if they have many KPI based on AA. 

I'm thinking to have the MIS Builder Analytic Items and put that as basic of for MIS Budgets (by AA) ?

TeoGoddet:
This seems to require a lot of work, what need to be done is discussed in #115

sbidoul:
Closing as duplicate of #115

Issue #115, opened 2018-08-23
Account groups expansion in reports


pedrobaeza:
I have to add groups expansion in reports, the same as accounts. My plans are:

* Add the option for showing groups and which depth level at report level.
* Stack and indent these groups accordingly.

And this is needed for v10! So I need to create 3 modules:

* account_group: Module for backporting account groups on v10
* l10n_es_account_group: Module with Spanish localization account group data.
* mis_builder_account_group: With this development

What do you think about this?

sbidoul:
Ouch, I'm not sure adding another expansion mechanism than accounts is feasible in an extension module for mis builder. This requires a refactoring first to generalize the expansion mechanism. I'll need to think about it.

pedrobaeza:
OK, let me know.

sbidoul:
@pedrobaeza I thought about this a bit more. This is indeed far from trivial. It requires 
1/ a generalization of the grouping/expansion mechanism which is currently hard coded to work on accounts; I've some sketchy ideas on how to do that (ie group/expand on other fields than account such as analytic dimensions or group_id), but it's a major surgery in MIS builder internals
2/ a second expansion mechanism to take into account the hierarchy of account groups

So if you are on a hurry or with restricted budget, it's better to seek alternatives.

For instance, a relatively easy approach to generate a trial balance based on account groups, is to generate KPIs from the groups (eg `bale[('group_id.id', '=', ...)]` or `bale[code_prefix%]`) from the account group definitions.

But I don't know the kind of report you want to produce. Do you have an example?

pedrobaeza:
I'm in a hurry I'm afraid, but I can develop fast also, hehe. This is going to be used with Spanish financial reports: https://github.com/OCA/l10n-spain/tree/11.0/l10n_es_mis_report, allowing to select on the moment they want to see the report, which account group level  (1, 2, 3, etc), and using account group definition from https://github.com/odoo/odoo/pull/23922

sbidoul:
Ah I see better, so you want to show the account details by account group instead of showing it by account?

That is quite easier to implement than the generalized grouping mechanism. Not in an extension module though. But if you backport account_group I see no problem to have mis_builder depend on it.

You need to start with a variant of [AEP.replace_exprs_by_account_id()](https://github.com/OCA/mis-builder/blob/34be8252f6d511cd6e2786a41d193e6935fdf912/mis_builder/models/aep.py#L407)

pedrobaeza:
[...]

PR #397, opened 2021-12-21
[mis_builder] generalize auto expands

TeoGoddet:
This is very far form mergability, 
it's purpose is to discuss an implementation of #115 (#392)

To do / to discuss : 

- how to choose which details lines to show (typicically for accounts, we need only the accounts included in the kpi expression)
We may have a domain field here : 
image

 - fix the drilldown (should not be a problem)

- optimisation issues 
Now the PR multiply the query number by the number of details line, to solve this we would need to add a group_by clause here : https://github.com/OCA/mis-builder/blob/90702af67af96d305b00eb11c5bf0bab2a42d6fe/mis_builder/models/aep.py#L355
but that would mean a details kind could not provide a domain for each line but a field for all line
that would also automatically solve the problem of wich details line we display
(We could get back wich lines to show from aep here instead having them from the report model here : https://github.com/OCA/mis-builder/blob/90702af67af96d305b00eb11c5bf0bab2a42d6fe/mis_builder/models/mis_report.py#L771)

- !! TEST !!

- genericity of the details line model
in the PR, i make account_account and account_analytic_account extends RowDetailIdentifierInterface
it prevent it to be compatible with account_move_line compatible model (budget, actuals, ..)
the right approach would be to encapsulate the right aml fields in the RowDetailIdentifierInterface

- make it easy to extend : 
split the _get_row_detail_identifiers in many _get_row_detail_identifiers_for_{}

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -14,6 +14,8 @@ from .accounting_none import AccountingNone
 
 _DOMAIN_START_RE = re.compile(r"\(|(['\"])[!&|]\1")
 
+UNCLASSIFIED_ROW_DETAIL = "other"
+
 
 def _is_domain(s):
     """Test if a string looks like an Odoo domain"""
@@ -300,16 +302,14 @@ class AccountingExpressionProcessor(object):
         date_to,
         additional_move_line_filter=None,
         aml_model=None,
+        auto_expand_col_name=None,
     ):
         """Query sums of debit and credit for all accounts and domains
         used in expressions.
 
         This method must be executed after done_parsing().
         """
-        if not aml_model:
-            aml_model = self.env["account.move.line"]
-        else:
-            aml_model = self.env[aml_model]
+        aml_model = self.env[aml_model or "account.move.line"]
         aml_model = aml_model.with_context(active_test=False)
         company_rates = self._get_company_rates(date_to)
         # {(domain, mode): {account_id: (debit, credit)}}
@@ -330,13 +330,16 @@ class AccountingExpressionProcessor(object):
             domain.append(("account_id", "in", self._map_account_ids[key]))
             if additional_move_line_filter:
                 domain.extend(additional_move_line_filter)
+
+            get_fields = ["debit", "credit", "account_id", "company_id"]
+            group_by_fields = ["account_id", "company_id"]
+            if auto_expand_col_name:
+                get_fields = [auto_expand_col_name] + get_fields
+                group_by_fields = [auto_expand_col_name] + group_by_fields
+
             # fetch sum of debit/credit, grouped by account_id
-            accs = aml_model.read_group(
-                domain,
-                ["debit", "credit", "account_id", "company_id"],
-                ["account_id", "company_id"],
-                lazy=False,
-            )
+            accs = aml_model.read_group(domain, get_fields, group_by_fields, lazy=False)
+
             for acc in accs:
                 rate, dp = company_rates[acc["company_id"][0]]
                 debit = acc["debit"] or 0.0
@@ -346,19 +349,45 @@ class AccountingExpressionProcessor(object):
                 ):
                     # in initial mode, ignore accounts with 0 balance
                     continue
-                self._data[key][acc["account_id"][0]] = (debit * rate, credit * rate)
+                if (
+                    auto_expand_col_name
+                    and auto_expand_col_name in acc
+                    and acc[auto_expand_col_name]
+                ):
+                    rdi_id = acc[auto_expand_col_name][0]
+                else:
+                    rdi_id = UNCLASSIFIED_ROW_DETAIL
+                if not self._data[key].get(rdi_id, False):
+                    self._data[key][rdi_id] = defaultdict(dict)
+                self._data[key][rdi_id][acc["account_id"][0]] = (
+                    debit * rate,
+                    credit * rate,
+                )
         # compute ending balances by summing initial and variation
         for key in ends:
             domain, mode = key
             initial_data = self._data[(domain, self.MODE_INITIAL)]
             variation_data = self._data[(domain, self.MODE_VARIATION)]
-            account_ids = set(initial_data.keys()) | set(variation_data.keys())
-            for account_id in account_ids:
-                di, ci = initial_data.get(account_id, (AccountingNone, AccountingNone))
-                dv, cv = variation_data.get(
-                    account_id, (AccountingNone, AccountingNone)
+            rdis = set(initial_data.keys()) | set(variation_data.keys())
+            for rdi in rdis:
+                if not initial_data.get(rdi, False):
+                    initial_data[rdi] = defaultdict(dict)
+                if not variation_data.get(rdi, False):
+                    variation_data[rdi] = defaultdict(dict)
+                if not self._data[key].get(rdi, False):
+                    self._data[key][rdi] = defaultdict(dict)
+
+                account_ids = set(initial_data[rdi].keys()) | set(
+                    variation_data[rdi].keys()
                 )
-                self._data[key][account_id] = (di + dv, ci + cv)
+                for account_id in account_ids:
+                    di, ci = initial_data[rdi].get(
+                        account_id, (AccountingNone, AccountingNone)
+                    )
+                    dv, cv = variation_data[rdi].get(
+                        account_id, (AccountingNone, AccountingNone)
+                    )
+                    self._data[key][rdi][account_id] = (di + dv, ci + cv)
 
     def replace_expr(self, expr):
         """Replace accounting variables in an expression by their amount.
@@ -371,23 +400,25 @@ class AccountingExpressionProcessor(object):
         def f(mo):
             field, mode, acc_domain, ml_domain = self._parse_match_object(mo)
             key = (ml_domain, mode)
-            account_ids_data = self._data[key]
+            rdi_ids_data = self._data[key]
             v = AccountingNone
             account_ids = self._account_ids_by_acc_domain[acc_domain]
-            for account_id in account_ids:
-                debit, credit = account_ids_data.get(
-                    account_id, (AccountingNone, AccountingNone)
-                )
-                if field == "bal":
-                    v += debit - credit
-                elif field == "pbal" and debit >= credit:
-                    v += debit - credit
-                elif field == "nbal" and debit < credit:
-                    v += debit - credit
-                elif field == "deb":
-                    v += debit
-                elif field == "crd":
-                    v += credit
+            for rdi in rdi_ids_data:
+                account_ids_data = self._data[key][rdi]
+                for account_id in account_ids:
+                    debit, credit = account_ids_data.get(
+                        account_id, (AccountingNone, AccountingNone)
+                    )
+                    if field == "bal":
+                        v += debit - credit
+                    elif field == "pbal" and debit >= credit:
+                        v += debit - credit
+                    elif field == "nbal" and debit < credit:
+                        v += debit - credit
+                    elif field == "deb":
+                        v += debit
+                    elif field == "crd":
+                        v += credit
             # in initial balance mode, assume 0 is None
             # as it does not make sense to distinguish 0 from "no data"
             if (
@@ -401,11 +432,11 @@ class AccountingExpressionProcessor(object):
         return self._ACC_RE.sub(f, expr)
 
     def replace_exprs_by_account_id(self, exprs):
-        """Replace accounting variables in a list of expression
+        """This method is depreciated and replaced by replace_exprs_by_row_detail.
+
+        Replace accounting variables in a list of expression
         by their amount, iterating by accounts involved in the expression.
-
         yields account_id, replaced_expr
-
         This method must be executed after do_queries().
         """
 
@@ -417,7 +448,7 @@ class AccountingExpressionProcessor(object):
             if account_id not in self._account_ids_by_acc_domain[acc_domain]:
                 return "(AccountingNone)"
             # here we know account_id is involved in acc_domain
-            account_ids_data = self._data[key]
+            account_ids_data = self._data[key][UNCLASSIFIED_ROW_DETAIL]
             debit, credit = account_ids_data.get(
                 account_id, (AccountingNone, AccountingNone)
             )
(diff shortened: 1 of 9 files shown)


mis_builder/models/kpimatrix.py
```diff
@@ -21,41 +21,70 @@
 _logger = logging.getLogger(__name__)
 
 
+class RowDetailIdentifierInterface(models.Model):
```
sbidoul: AbstractModel ?

sbidoul:
@TeoGoddet I don't have much time to look closely but this sounds promising!

> Now the PR multiply the query number by the number of details line, to solve this we would need to add a group_by clause here : 

This is probably the key thing to address as minimizing the number of db queries is a core design goal of MIS Builder (and what makes it complex - if we leave the many historical reason asides).

TeoGoddet:
Another big thing that need to be discussed is expanding by many2many fields
The problems is the same that exists with group_by and many2many, what to do when you have multiple tags, dupllicate the lines, split them, ... ? 
Maybe there is no generic solutions to this and we may have case by case solution (such as using analytic tag dimension for analytic tags)

sbidoul:
Yes, expanding by m2m is impossible in the general case. Expanding on many2one is the only thing that can be made generic, and, as you, say, users can use analytic dimensions to fallback on m2o.

TeoGoddet:
I just pushed a new version that is closer to the goal.
* It works, should be ok in performance optimal but I have not benchmark setup to test it.
* drilldown works !
* need to remove some legacy methods (or we should keep them for backward compatibility ?)
* still need to rewrite the test
* I renamed as least as possible field, maybe I should rename them and add a migration script ?
* Maybe reduce complexity :  the RowDetailIdentifier class is a pretty useless abstraction if we do not make it an odoo ihneritable class.

mis_builder/models/expression_evaluator.py
```diff
@@ -6,12 +6,7 @@
 
 class ExpressionEvaluator(object):
     def __init__(
-        self,
-        aep,
-        date_from,
-        date_to,
-        additional_move_line_filter=None,
-        aml_model=None,
```
sbidoul: Please keep the trailing commas as they are, so there are no needless diff. I recommend you install the latest version of black, and/or run `pre-commit run -a` locally to apply the same formatters and linters than the CI.

mis_builder/tests/test_mis_report_instance.py
```diff
@@ -376,22 +376,22 @@ def test_multi_company_compute(self):
...
+                    "%s [%s]" % (account.name_get()[0][1], account.company_id.name),
                 )
         self.report_instance.write({"multi_company": False})
         matrix = self.report_instance._compute_matrix()
         for row in matrix.iter_rows():
-            if row.account_id:
-                account = self.env["account.account"].browse(row.account_id)
-                self.assertEqual(row.label, "{} {}".format(account.code, account.name))
+            if row.row_detail_identifier:
+                account = self.env["account.account"].browse(row.row_detail_identifier)
```
sbidoul: Looking at this specific change, it might read better with something like this:

```suggestion
                account = self.env[row.row_detail_model].browse(row.row_detail_res_id)
```

Not sure.

mis_builder/tests/test_mis_report_instance.py
```diff
@@ -501,11 +497,7 @@ def test_query_company_ids(self):
         assert self.report_instance.query_company_ids == self.report_instance.company_id
         # create a second company
         c1 = self.report_instance.company_id
-        c2 = self.env["res.company"].create(
-            dict(
-                name="company 2",
```
sbidoul: Another example here where you should keep the trailing comma to reduce the diff (sorry to nitpick but it's important to me).

setup/_metapackage/setup.py
```diff
@@ -1,20 +1,20 @@
...
     classifiers=[
-        'Programming Language :: Python',
-        'Framework :: Odoo',
-        'Framework :: Odoo :: 14.0',
-    ]
+        "Programming Language :: Python",
+        "Framework :: Odoo",
+        "Framework :: Odoo :: 14.0",
+    ],
 )
```
sbidoul: You should not touch this `setup/` directory.

sbidoul (commented):
Looks like you are making good progress :) Let me know when you think it's ready for a first review.

TeoGoddet:
Hi ! Thanks for the review, sorry for the noise on the CI, i could not get a working local CI.
My progress so far are 
- everything is working (need extensive testing)
- backward-compatible aep (according to the ci tests may need real use tests)
- code almost clean

I'm unsure about the name of the variables introduced, they are not the result of a good reflection but of the long dev process.
You made good suggestions, big thanks !

setup/_metapackage/setup.py
TeoGoddet: Oupsi, had a problem running black and linting tools, I will check my diff

mis_builder/models/kpimatrix.py
TeoGoddet: Finally removed this as it was only more complexity, but indeed an abstractmodel is still an option
In this proposition, the id of the record is used to keep track and the record name_get function is used to get a name

TeoGoddet:
In another MR, i could propose to have the expand field to be choosen dynamicaly using a many2one to ir.field

TeoGoddet:
Ok, got the dev tools working better so the diff should be minimal
@sbidoul it's now really close to ready, i'd love to have a review
I'm affraid i need better names for `row_detail_identifier` and `row_detail_model`

Another thing have still in mind is to rewrite the expand field selection to be dynamic (maybe another MR)

TeoGoddet:
@sbidoul the tests are now fixed, I think it's ready for review

sbidoul:
Thanks a lot @TeoGoddet. I'll try to review and test this in depth soon.

AntoniRomera (approved):

hitrosol:
This PR works well in runboat. Thanks for @TeoGoddet and @sbidoul for this nice new feature.

hitrosol:
I try to combine this PR with https://github.com/OCA/mis-builder-contrib/pull/22 and MIS Budget (By Accounts), give us flexibility show the budget by AA, if we select the Auto Expand Detailed by AA. 
So we don't have to define line by line for each AA.

image

AntoniRomera:
@sbidoul You have been able to check them, from the tests carried out it seems to work without any problems.

sbidoul:
Yes I'll be reviewing this in details in the next days.

mis_builder/models/mis_report.py
```diff
@@ -748,21 +766,23 @@ def _declare_and_compute_col(  # noqa: C901 (TODO simplify this fnction)
...
-                for (
-                    account_id,
-                    vals,
-                    drilldown_args,
-                    _name_error,
-                ) in expression_evaluator.eval_expressions_by_account(
-                    expressions, locals_dict
-                ):
+                rdis = expression_evaluator.eval_expressions_by_row_detail(
+                    expressions, locals_dict  # , self.auto_expand_col_name
```
sbidoul: Remove this comment ? Or is there something left to do here ?

mis_builder/models/mis_report.py
```diff
@@ -1001,3 +1021,6 @@ def _evaluate(
             no_auto_expand_accounts=True,
         )
         return locals_dict
+
+    def _should_display_auto_expand(self, kpi, rdi, vals):
```
sbidoul, line 1024: I suppose this is to allow overriding ? Add a docstring to explain the use cases ?

mis_builder/models/expression_evaluator.py
```diff
@@ -50,6 +51,7 @@ def eval_expressions(self, expressions, locals_dict):
                 drilldown_args.append(None)
         return vals, drilldown_args, name_error
 
+    # we keep it for backward compatibility
```
sbidoul, line 54: Since it is not used anymore, better remove it now.

mis_builder/models/mis_report.py
```diff
@@ -91,11 +91,11 @@ class MisReportKpi(models.Model):
         copy=True,
         string="Expressions",
     )
-    auto_expand_accounts = fields.Boolean(string="Display details by account")
+
+    # TODO : this fields should be renamed to auto_expand_details or something like this
```
sbidoul: Hm, I think we should not rename them to avoid breaking report data files that are everywhere in locallisations.

sbidoul: The *label* of the `no_auto_expand_accounts` needs updating, though.

mis_builder/models/mis_report.py
```diff
@@ -465,6 +465,20 @@ def _default_move_lines_source(self):
     )
     account_model = fields.Char(compute="_compute_account_model")
 
+    auto_expand_col_name = fields.Selection(
+        [
+            ("account_id", _("Accounts")),
+            ("analytic_account_id", _("Analytic Accounts")),
+            ("partner_id", _("Parner")),
+        ],
```
sbidoul, line 472: Wondering if this should be a m2o to `ir.model.fields` instead of a selection.
In principle any m2o field should work.
Also, there is no guarantee there is a partner_id field in all move line sources.

mis_builder/models/kpimatrix.py
```diff
@@ -215,22 +216,23 @@ def set_values(self, kpi, col_key, vals, drilldown_args, tooltips=True):
             kpi, col_key, None, vals, drilldown_args, tooltips
         )
 
+    # TODO this could be renamed set_values_detail
```
sbidoul, line 219: I think we can do this renaming now.

sbidoul (commented):
Thanks for your patience and this great works.
I'll continue testing and reviewing and testing in the coming days, but here is a first set of (minor) comments.

Note to self: check impact on mis_builder_budget.

mis_builder/models/kpimatrix.py
```diff
@@ -462,23 +464,27 @@ def iter_subcols(self):
...
-        result = "{} {}".format(account.code, account.name)
-        if self._multi_company:
-            result = "{} [{}]".format(result, account.company_id.name)
+            rdi_ids.update(detail_rows.keys())
+        rdi_ids = list(rdi_ids)
+        if UNCLASSIFIED_ROW_DETAIL in rdi_ids:
+            rdi_ids.remove(UNCLASSIFIED_ROW_DETAIL)
+        rdis = self._rdi_model.search([("id", "in", rdi_ids)])
+        self._rdi_names = {rdi.id: self._get_rdi_name(rdi) for rdi in rdis}
+        self._rdi_names[UNCLASSIFIED_ROW_DETAIL] = _("Other")
```
sbidoul: When there is only one detail line with, say, no analytic account, it shows `Other` which is a little difficult to understand.
Would it be better to display `_("(None)")` or `_("(not set)")` in this case ?

pedrobaeza: or `_("Undefined")`

sbidoul: or `_("(not set)")` which is used in Custom filters. Not sure what's easier to understand for business folks.

pedrobaeza: Yes, it depends also on the translations, but I agree `Other` can't be.

pedrobaeza:
Note that I tried this in a branch mentioned in #115, but it's good to have this being considered finally. One of the problems I faced was the variables everywhere are mentioning `account`, and thus letting that names will lead to confusions, but that change made to conflict each time a change was introduced in main branch.

sbidoul:
> Note to self: check impact on mis_builder_budget.

Tested with budgets by account. This works well.

mis_builder/models/mis_report.py
sbidoul: Or perhaps we *should* rename the fields and keep the old field name with compute/inverse (with a deprecation log) for backward compatibility.

mis_builder/models/aep.py
```diff
@@ -401,11 +432,11 @@ def f(mo):
         return self._ACC_RE.sub(f, expr)
 
     def replace_exprs_by_account_id(self, exprs):
-        """Replace accounting variables in a list of expression
-        by their amount, iterating by accounts involved in the expression.
+        """This method is depreciated and replaced by replace_exprs_by_row_detail.
```
sbidoul: Since it is not used inside mis_builder anymore, let's remove it. It is unlikely that anyone is using that, and we'll do a major version bump anyway.

sbidoul:
A few minor comments, but overall, this looks very good to me :+1: Thanks a ton for this contrib.

mis_builder/models/kpimatrix.py
```diff
@@ -462,23 +464,27 @@ def iter_subcols(self):
...
         return result
 
-    def get_account_name(self, account_id):
-        if account_id not in self._account_names:
-            self._load_account_names()
-        return self._account_names[account_id]
+    def get_rdi_name(self, rdi_id):
+        if rdi_id not in self._rdi_names:
+            self._load_rdi_names()
+        return self._rdi_names[rdi_id]
```
FernandoRomera, line 487: I have done a local test and with this change we avoid possible errors.
```suggestion
        return self._rdi_names.get(rdi_id, _("Other"))
```

EDIT: This is necessary due to archived records, because apparently when the record is archived we don't get the id and name right.

sbidoul: @FernandoRomera can you provide an scenario to reproduce the problem ?

mis_builder/models/aep.py
```diff
@@ -346,19 +349,45 @@ def do_queries(
...
+                    auto_expand_col_name
+                    and auto_expand_col_name in acc
+                    and acc[auto_expand_col_name]
+                ):
+                    rdi_id = acc[auto_expand_col_name][0]
+                else:
+                    rdi_id = UNCLASSIFIED_ROW_DETAIL
+                if not self._data[key].get(rdi_id, False):
+                    self._data[key][rdi_id] = defaultdict(dict)
+                self._data[key][rdi_id][acc["account_id"][0]] = (
```
sbidoul, line 362: Note to self: understand why we group by rdi_id and then account_id and not the contrary.

mis_builder/models/kpimatrix.py
AntoniRomera: A possible scenario is when you add the option to group by "product_id" and you are looking at a previous year and as you don't use it anymore you have archived the product to avoid problems, but you still have to be able to see it in previous years' reports.

AntoniRomera: For the rest it works great, we have done a couple of tests and it makes it very easy to generate reports.

mis_builder/models/kpimatrix.py
```diff
@@ -462,23 +464,27 @@ def iter_subcols(self):
...
+        rdi_ids = list(rdi_ids)
+        if UNCLASSIFIED_ROW_DETAIL in rdi_ids:
+            rdi_ids.remove(UNCLASSIFIED_ROW_DETAIL)
+        rdis = self._rdi_model.search([("id", "in", rdi_ids)])
+        self._rdi_names = {rdi.id: self._get_rdi_name(rdi) for rdi in rdis}
+        self._rdi_names[UNCLASSIFIED_ROW_DETAIL] = _("Other")
+
+    def _get_rdi_name(self, rdi):
+        result = rdi.name_get()[0][1]
+        if self._multi_company and rdi.company_id:
```
sbidoul, line 480: What if `rdi` does not have a `company_id` field?

mis_builder/models/kpimatrix.py
```diff
@@ -462,23 +464,27 @@ def iter_subcols(self):
...
-
-    def _get_account_name(self, account):
-        result = "{} {}".format(account.code, account.name)
-        if self._multi_company:
-            result = "{} [{}]".format(result, account.company_id.name)
+            rdi_ids.update(detail_rows.keys())
+        rdi_ids = list(rdi_ids)
+        if UNCLASSIFIED_ROW_DETAIL in rdi_ids:
+            rdi_ids.remove(UNCLASSIFIED_ROW_DETAIL)
+        rdis = self._rdi_model.search([("id", "in", rdi_ids)])
```
sbidoul: search with active_test = False ?

mis_builder/models/kpimatrix.py
sbidoul: Ok, so maybe it's better to search for inactive records too in `_load_rdi_names()` ?

mis_builder/models/kpimatrix.py
TeoGoddet: I'm definitely not good as this kind of things + I'm not english native so i'm waiting for your best ideas :)
Would you like to keep Other when there are details line and  _("(not set)") or equivalent when there is none ? Or _("(not set)") always ?

mis_builder/models/mis_report.py
TeoGoddet: I imagined this as a second step, but it's pretty simple, so lets do it in this one

TeoGoddet:
@sbidoul I'm not forgetting about this but need some time to fix everything, thanks a lot for the review !

sbidoul:
No worries. Will you be at OCA days?

TeoGoddet:
Normally Yes :)

AntoniRomera:
@TeoGoddet If you agree, I can help you and make a PR in your repo with the proposed changes.

TeoGoddet:
@AntoniRomera would be very appreciated !!

AntoniRomera:
@sbidoul @TeoGoddet Is this at a standstill? Do you plan to do the merge?

rafaelbn:
Could be great to have this merged in 14 to port for 15 . @TeoGoddet did you got @sbidoul in OCA Days? Do you agree to get this merged? Thank you!

AntoniRomera:
@sbidoul @pedrobaeza @rafaelbn 

This PR has been ready for months, I still don't see why it can't be integrated, if it is to keep the history, can it be done in version 16.0 or how do you see it?

sbidoul:
@AntoniRomera I was waiting for answers to the questions I asked above.
I now notice there has been an additional commit since. So I'll try to find the time to review this again.
There were pre-commit and test issues, though so a rebase, and possibly some fixes, are necessary.

TeoGoddet:
@AntoniRomera @FernandoRomera i wont be able to work on this for at leat another 6 months, feel free to taleover the PR

PR #404, opened 2022-01-21
[DO NOT MERGE] ci: detect unreleased dependencies

sbidoul:


legalsylvain (approved):
Hi @sbidoul

Looks great to prevent merging PR that depends on unmerged extra PR.

**Naive question**
The syntax here :  
``git+https://github.com/oca/server-auth@14.0#subdirectory=setup/auth_jwt``

seems lighter that the syntax you proposed [here](https://github.com/OCA/oca-ci/issues/23#issuecomment-1016645879) : 
``odoo12-addon-product_print_category @ git+https://github.com/grap/grap-odoo-incubator#subdirectory=setup/product_print_category``

there is a difference between both ?

sbidoul:
> there is a difference between both ?

With recent pip version, both should work similarly. The branch (`@14.0`) is mandatory, though, otherwise you get the default branch which does not mean much in OCA.

.github/workflows/test.yml
```diff
@@ -9,6 +9,19 @@ on:
...
+  unreleased-deps:
+    runs-on: ubuntu-latest
+    name: Detect unreleased dependencies
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          if [ -f test-requirements.txt ] ; then
+            grep "://" test-requirements.txt
+            if [ $? -eq 0 ] ; then
+              echo "Unreleased dependencies found in test-requirement.txt."
```
joao-p-marques:
```suggestion
              echo "Unreleased dependencies found in test-requirements.txt."
```

joao-p-marques (commented):
Just found a small typo, otherwise looks good... 

Could there be any situation where we might use a path like `git+https://etc` to reference a **released** dependency?

sbidoul:
> Could there be any situation where we might use a path like git+https://etc to reference a released dependency?

I can't imagine one, in OCA at least.

legalsylvain:
I could see two limitation : 

1) this PR forbid to write a comment. isn'it ? I mean something like

```
# Note : blablabla, see https://github.com/OCA/XXX/issues/YY
name_of_package
```

maybe we could write a regex that skip line that begin by ``#``

2) in some cases, in ``(test-)requirements.txt`` we pin a specific version of a python library. See : 
https://github.com/OCA/server-tools/blob/14.0/test-requirements.txt

No idea if it should be forbidden.

sbidoul:
1. yes, we can probably improve the regex
2. openupgradelib is now on PyPI. and the odoorpc on PyPI is reasonably fresh, so we can remove such git references from test-requirements.txt

.github/workflows/test.yml
```diff
@@ -16,11 +16,12 @@ jobs:
       - uses: actions/checkout@v2
       - run: |
           if [ -f test-requirements.txt ] ; then
-            grep "://" test-requirements.txt
-            if [ $? -eq 0 ] ; then
-              echo "Unreleased dependencies found in test-requirement.txt."
-              exit 1
-            fi
+              # reject non-comment lines that contain a / (i.e. URLs, relative paths)
+              grep "^[^#].*/" test-requirements.txt
```
legalsylvain: :+1: !

legalsylvain:
This test should be done also on ``requirements.txt`` ? don't you think ?

sbidoul:
> This test should be done also on requirements.txt ? don't you think ?

Yes, I was thinking about that too, in case someone manages to put a local ref in `external_dependencies`. Done.

joao-p-marques (approved):

PR #406, opened 2022-01-26
[15.0][MIG] mis_builder_demo,mis_builder_budget: Migration to 15.0

yankinmax:

PR #408, opened 2022-01-28
[IMP] mis_builder: remove redundant super className in Python3

yankinmax:


yankinmax:
@sbidoul this PR is opened to respect your suggestion here https://github.com/OCA/mis-builder/pull/406#discussion_r794294384

PR #410, opened 2022-01-31
[14.0][FIX] mis_builder: fix access right issues when using actions

sebalix:


sbidoul:
The code looks better indeed, but do you care explaining what this actually fixes ?

Do you think this should work as is on other Odoo versions?

i-vyshnevska (approved):

i-vyshnevska:
@sbidoul direct access to "ir.actions.act_window" is allowed only for admin since v14, now access to action should be done  through the method "_for_xml_id"

sbidoul:
Thanks for the explanation, @i-vyshnevska. LGTM.
Could you add a readme/newsfragments/410.bugfix with a one liner end-user oriented changelog entry?

yankinmax (approved):

sebalix:
@sbidoul done, but regarding the `get_wizard_report_action` method I didn't find where this method is used so I didn't mention this part in the changelog entry.

yankinmax:
@sbidoul maybe we should also port these fixes to v15?

sbidoul:
Thanks for the fix!

Issue #415, opened 2022-02-16
[15.0+] mis_builder: non-admin user permission issues


sbidoul:
You are not allowed to access 'Models' (ir.model) records.

I found this while trying to test the forward port of #410

pedrobaeza:
I think we can provide a selection with a restricted list of model names, and the modules adding extra models may feed such selection.

rafaelbn:
This is a BUG. I can confirm it in production. We have to give all accountants admin right to work with MIS

**`Administration/Access Right`**

![imagen](https://user-images.githubusercontent.com/8736623/200616795-c62443e7-78e3-4c0f-98e8-aac15c6c089d.png)

Is this related @sbidoul about what you said?

> You are not allowed to access 'Models' (ir.model) records.


- [FIX] l10n_es_aeat: do not require administration permissions #[2583](https://github.com/OCA/l10n-spain/pull/2583)

@moduon MT-1527

sbidoul:
Normally the error I had in mind with this issue occurred only when creating *report templates*.
But if it occurs when creating reports, then we can probably split this issue in two that can be fixed independently:
1. make sure you don't need admin rights to view report templates and create reports
2. make sure you don't need admin rights to create report templates

(1) should be easier to solve than (2).

yajo:
Fix in https://github.com/OCA/mis-builder/pull/476

sbidoul:
Keeping open as users still need admin right to create report templates.

AnizR:
I think that adding groups to mis_builder might be a good idea:
 - A group to read reports
 - A group to edit reports
 - A group to create and edit templates
 
 An idea to tackle the problem of having access to `ir_model` and `ir_model_fields`:
  - Adding a configuration that only an admin (`group_erp_manager`) can set, that states the 'usable' models and fields
  - Having a Selection that displays those 'allowed models and fields' on templates

PR #416, opened 2022-02-16
[15.0] fix permission issues when running as non-admin user

sbidoul:
Towards #415

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -427,7 +427,7 @@ class MisReport(models.Model):
     _description = "MIS Report Template"
 
     def _default_move_lines_source(self):
-        return self.env["ir.model"].search([("model", "=", "account.move.line")])
+        return self.env["ir.model"].sudo().search([("model", "=", "account.move.line")])
 
     name = fields.Char(required=True, translate=True)
     description = fields.Char(required=False, translate=True)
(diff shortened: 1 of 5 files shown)


sbidoul:
This at least lets non-admin use and display mis reports.

Creating report templates still require admin rights in some circumstances, such as creating queries.
I'm not sure how to solve this since we need to display m2o on model and fields in the UI, and these now require admin rights.

Suggestions welcome.

pedrobaeza:
IIRC, that is only the advanced template options, isn't it? Why not putting them in separate views protected with the same settings permission and let usual templates to work with the same permission level?

sbidoul:
@rafaelbn would you like to test this one on runboat?

yajo (approved):

loida-vm:
The users can access to MIS Reporting if they are Billing Administrator, but not system administrator. 
LGTM :+1:

ghost (approved):
LGTM!

loida-vm (approved):

rafaelbn (commented):
My review with a question:

- Should have access to MIS REPORT a only "Billing" user?

https://www.loom.com/share/3f7fcc9634bf4d6ebe7b56ef4402a882

sbidoul:
/ocabot merge patch

Thanks for the review, folks.

@rafaelbn regarding access for billing users, well, it's debatable, but I think it is a reasonable default. It's been like this forever.
After all company P&L and Balance Sheet are typically public information.
And Billing Users have access to account moves anyway, so it's just an URL away from being visible.

We have done projects where access to account move was restricted by department and such so department managers can see the P&L of their department but not others.

pedrobaeza:
The menu visibility was modified in #460

PR #417, opened 2022-02-16
[15.0] remove u string prefixes

sbidoul:
Remove a bunch of unicode string prefixes that are useless in python 3.


pedrobaeza (approved):
Yeah, keeping in sync branches prior to 11.0 is now something not feasible IMO

PR #418, opened 2022-02-16
[FIX] pdf export issue introduced when backporting

sbidoul:
Fix issue introduced in #413


pedrobaeza (approved):
Thanks for the quick reaction

PR #421, opened 2022-03-07
[14.0][FIX] mis_builder: Be tolerant in tests with several installed companies

pedrobaeza:
If you are running tests in a integrated environment, you can have more than one company with the same account 200x, so we need to filter on the current active company.
    
We may have as well more companies as authorized ones, so let's assert the proper thing in `test_query_company_ids`.

@Tecnativa


pedrobaeza:
Hi @sbidoul we have found some problems in integration tests and this patch solves them. Can you please check?

flotho (approved):
Code review,
purpose is explicit and code conform
LGTM
Thanks @pedrobaeza

sbidoul:
LGTM. Thanks! 

/ocabot merge patch

PR #431, opened 2022-05-05
[FIX] mis_builder: Delete mis.report: delete dependency to subreport, restrict if other reports depend on the report

ghost:
I got an error when I tried to delete a report which had a subreport.
With this PR, the report's subreport record(s) will also be deleted.

### DIFF
--- a/mis_builder/models/mis_report_subreport.py
+++ b/mis_builder/models/mis_report_subreport.py
@@ -20,8 +20,16 @@ class MisReportSubReport(models.Model):
     _description = "MIS Report - Sub Reports Relation"
 
     name = fields.Char(required=True)
-    report_id = fields.Many2one(comodel_name="mis.report", required=True)
-    subreport_id = fields.Many2one(comodel_name="mis.report", required=True)
+    report_id = fields.Many2one(
+        comodel_name="mis.report",
+        required=True,
+        ondelete="cascade",
+    )
+    subreport_id = fields.Many2one(
+        comodel_name="mis.report",
+        required=True,
+        ondelete="restrict",
+    )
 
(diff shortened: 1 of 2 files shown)


sbidoul (changes requested):
Hi, I'd rather avoid cascading delete reports, as they are time consuming to build and deleting them without warning may surprise and upset users.

We could improve the error message, though.

ghost:
No mis.report will be deleted by cascading. But when a user chooses to delete a mis.report, the one2many mis.report.subreport records should be deleted. The only purpose of mis.report.subreport is to link a mis.report to another mis.report.

sbidoul:
Ah, yes, you are right. I read too quickly.

But still, this would silently break reports that depend on the one that is being deleted, correct ?

ghost:
No, there is no silently breaking. I tested with `mis_template_financial_report`, where the Balance Sheet depends on the Profit & Loss. I could not delete PL right away. But I could delete the BS and then the PL.

sbidoul:
If the BS depends on the PL and you delete the PL, the BS will not be functional anymore. And I think we should warn users about that.

ghost:
I could NOT delete PL right away.
Now I have set ondelete="restrict" for subreport_id, to ensure no silently breaking.

sbidoul:
`restrict` is good, thanks ! Could you fix the lint error and a a file named `431.feature` in mis_builder/readme/newsfragments with a short changelog entry ?

ghost:
`black` failed in pre-commit, I don't know why.

sbidoul:
I finalized this in #445 and ported to 12, 14, 15. Thanks for the fix!

PR #434, opened 2022-06-25
[IMP] mis_builder: child_ of for analytic account group filtering

TeoGoddet:
With this commit, the filter on analytic group also include lines with analytic account in childs group of the targeted group


sbidoul:
Hi, this can be seen as a backward incompatible change.
I think in some customer projects we solved this in an extension module that added a check box to control the operator.
PS: I'm not forgetting your #397

Issue #393, opened 2021-12-07
14.0: unable to add SUM COL at once 


petrus-v:
Issue is probably known ? I haven't found any similar referenced...

## Module
``mis_builder``

## Describe the bug

Configuring MIS Report, I'm unable to configure a *Sum columns* without saving the new column first. 

## To Reproduce
**Affected versions**: 14.0

1. connect to runbot and go to the `Demo Expenses vs Budget` MIS Report (Invoicing > Reporting > MIS Reports) 
2. Edit the report in form view and add a new column "Available 2"
3. select "sum columns" source (`sumcol`)
4. when you want to add lines of columns to sun the list of available column is empty

**Expected behavior**
available columns fulfilled while creating the sum column

**Additional context**

To be able to configure that kind of column we have to save the column first as regular column then edit again the report to change as sumcol then available column are listed.

alexis-via:
I confirm this bug.

alexis-via:
Bug https://github.com/OCA/mis-builder/issues/456 seems to be a duplicate of this one.

alexis-via:
The field period_to_sum_id has domain="[('report_instance_id', '=', parent.report_instance_id), ('id', '!=', parent.id)]" defined in the view (there is no domain in the code). If I remove ('report_instance_id', '=', parent.report_instance_id) from the domain, this list is not empty any more.

alexis-via:
Here is the exact scenario:
create the MIS report instance with 2 periods
SAVE
create a 3rd period with source = sumcol => in source_sumcol_ids, you cannot select period_to_sum_id (the list is empty).

Workaround:
create the MIS report instance with 2 periods
SAVE
create a 3rd period with source = sumcol and a name
SAVE
edit the 3rd period and the "source_sumcol_ids" will work without problems, you will be able to select period_to_sum_id

sbidoul:
This was fixed in #457

PR #457, opened 2022-08-05
[14.0][FIX] mis_builder: bug unable to add SUM COL at once

alexis-via:
Fixes bug #393. It's a hack, but it works well. I tried TONS of other solution (would take too much time to list them here), and it's the only solution I found.

Fix missing default value on company_id when creating a new report instance with default values


rapvalencia (approved):

petrus-v (approved):
It works like expected thanks for this fix.

I would love to see unit test to notice developer in case they remove it (by mistake) but won't lock as long issue is referenced !

Thanks a lot for your time

PR #458, opened 2022-08-11
[13.0][IMP] mis_builder: Add group to MIS report menu to prevent users without full accounting permissions can see it.

victoralmau:
Add group to MIS report menu to prevent users without full accounting permissions can see it.

This is required too in v14 and v15.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT38538

### DIFF
--- a/mis_builder/views/mis_report.xml
+++ b/mis_builder/views/mis_report.xml
@@ -310,6 +310,7 @@
         parent="account.menu_finance_configuration"
         name="MIS Reporting"
         sequence="90"
+        groups="account.group_account_user"
     />
     <menuitem
         id="mis_report_view_menu"


pedrobaeza (approved):

chienandalu (approved):

sbidoul:
It's not obvious to me that we should do this. Can you elaborate the reasons?

pedrobaeza:
Without this, a simple invoicing user (with no accounting access), can see complete P&L and balance sheets of the company. I don't think that's advisable. The same they don't see the rest of the accounting reports:

imagen

cesar-tecnativa (approved):

sbidoul:
Ok, thanks. This looks like a better default indeed.

/ocabot merge minor

pedrobaeza:
Thanks! Should we do the fw-ports?

sbidoul:
> Thanks! Should we do the fw-ports?

That would be nice, yes, thanks!

pedrobaeza:
Done at #459 and #460

Issue #455, opened 2022-07-27
[15.0] mis.committed.purchase not deleted when PO is reset to draft


fclementic2c:
## Module
MIS Builder / MIS Builder demo (because I use the committed amount alternative comparison)

## Describe the bug
Create a report comparing Actual Vs PO commited amounts
Create confirm a PO -> mis.committed.purchase are created
Cancel the PO, reset to Draft and re-confirm it -> mis.committed.purchase are duplicated

mis.committed.purchase should be deleted when PO is cancelled

see screenshots : https://docs.google.com/document/d/1dvTCCX51I9ZrgTEvpoFB8AVjLVlie8P5LpxaeUp3PQc/edit?usp=sharing

PR #466, opened 2022-08-31
[15.0][ADD] mis_builder_alternative_commitment: initial commit

Camille0907:
## Description
Add an addon to create a alternative source based on PO lines with MIS Builder, based on [MIS Builder Demo](https://github.com/OCA/mis-builder/tree/15.0/mis_builder_demo) addon.

The goal is, as in the demo module, to have a committed purchase view model, but here we include all purchases that are not in draft or cancelled state, even if they have been invoiced. 

Another difference is in the computation of the account_id, for the committed purchase : in addition of what's done in the demo, if the product_category's account_id is not set, the account_id tries to base itself on the first set account_id on five generations of parent_category, before searching for default value for new resources.

One fix, compared to the demo : the management of multicompany, to avoid the multiplication of the number of commitments by the number of companies : `AND ipd.company_id=po.company_id` is added [here](https://github.com/OCA/mis-builder/pull/466/files#diff-75cd70cf356375c6e5a8d945e62d905e59eb85b056239d2ca576c5f94a640a3aR60)
(Related to this issue : #455)

**Needed improvement :**
It would be better if the account_id tried to base itself on all the generations of parent_category, as long as it is set.


Camille0907:
I see the tests are red but don't understand how to make it turn to green. 
If I understand the error, it should probably be resolved by adding casts where there are ambiguities... But I don't see what is ambiguous :/

fclementic2c:
Works like a charm :)
Well done

fclementic2c (approved):
Functionel test ok

sbidoul:
LGTM at first sight. Could you move this PR to the mis-build-contrib repo, as we keep only the core modules here?

yankinmax:
Superseded by https://github.com/OCA/mis-builder-contrib/pull/25

yankinmax:
@Camille0907 can you pls close this PR as it's superseded?

Issue #183, opened 2019-04-30
Comparison by Companies


hitrosol:
Is it possible to compare the report by companies as follow using the MIS Builder ?

**Account ....... Grand Total ......Company A ......Company B.........etc**
Account 1 .................xxx................xxx....................xxx..................xxx
Account 2 .................xxx................xxx....................xxx..................xxx
Account 3.................xxx................xxx....................xxx..................xxx
etc....

and 

**Analytic Account ....... Grand Total ......Company A ......Company B.........etc**
AA 1 ...................................xxx................xxx....................xxx..................xxx
AA 2 ..................................xxx................xxx....................xxx..................xxx
AA 3...................................xxx................xxx....................xxx..................xxx
etc....

Thanks

sbidoul:
This is not supported today. It should be relatively easy to implement as a variant of #138 and #96.

sbidoul:
This is now supported via the Analytic Domain field on period/columns, which can be `[("company_id", "=", ...)]`. Not sure we need a specialized field to filter on companies.

PR #472, opened 2022-10-03
[MIG] Migrate mis_builder, mis_builder_budget, mis_builder_demo to 16.0

adrienpeiffer:
TODO :

- [x] port #416
- [x] port #474 
- ~port #481~ analytic tags are gone
- [x] port #484
- [x] Integrate analytic Apocalypse
- [x] Integrate account type rework (https://github.com/odoo/odoo/commit/26b2472f4977ccedbb0b5ed5f08be2c04313fd21)
- [x] Migrate JS to OWL 2.0 ?
- [ ] Functional testing

Depens on :
- [x] https://github.com/OCA/server-ux/pull/535
- [x] https://github.com/OCA/reporting-engine/pull/690


adrienpeiffer:
https://github.com/odoo/odoo/pull/101701

FernandoRomera:
@adrienpeiffer 
Merged:
https://github.com/OCA/server-ux/pull/535
https://github.com/OCA/reporting-engine/pull/690

Can you rebase, please?

ovnicraft:
@adrienpeiffer any news about this? I would like to know if more hands are needed here.

sbidoul:
I had a quick look at this.

The PDF and XLSX exports work.

There are a few UX issues (default company, preview/export buttons on the report form are not visible).

The preview widget does not work. Does it need to be rewritten in OWL?

The analytic filters... well I don't immediately see how we could implement them in 16.0.
When using the journal items as data source, we can't do much filtering with the analytic distribution json field.
Maybe we could promote analytic lines as an official data source (like in [mis_builder_analytic](https://github.com/OCA/mis-builder-contrib/blob/14.0/mis_builder_analytic/models/mis_account_analytic_line.py)), and enable analytic filters only when that data source is selected? Not sure. Suggestions welcome.

dzungtran89:
hi @sbidoul , about the widget `mis_report_widget`, it needs to be migrated to owl in order to work with odoo 16, so I think we can do step by step:
- Migrate the widget to owl to make the preview work at least -> I'll do this
- About the filter, for the moment I can add the filter of analytic lines `m2m`

sbidoul:
> Migrate the widget to owl to make the preview work at least -> I'll do this

Thanks! Please do a PR to acsone:16.0-mig-mis_builder-ape

> About the filter, for the moment I can add the filter of analytic lines m2m

I'm not sure how that would work?

dzungtran89:
> I'm not sure how that would work?

I think, because the analytic lines link to the `move_line_id (m2o)`, so with this filter, we could get relevant move lines for the report

sbidoul:
Hm, yes, but I'm afraid that will not work performance-wise.

rafaelbn:
Hello! Thank you! 😄 

Could you please rebase to check and test?

Thank you!

sbidoul:
@rafaelbn there is nothing to rebase. runboat is usable.

sbidoul:
A few more thoughts about analytic filters.

I'm going to removing everything that concerns analytic tags and analytic groups. These concepts are gone for good.

The remaining analytic filters would be analytic account and analytic plan. Selecting an analytic plan would be mandatory when filtering on analytic accounts. 

Maybe we can make the analytic filters visible only if analytic_account_id and analytic_plan_id are present in the move line source.

I'm still wondering what the analytic plan hierarchy is good for...

sbidoul:
@dzungtran89 little steps that can be done:

- [x] a PR with one commit to remove code related to analytic tags (across all 3 modules)
- [x] a PR with the forward ports mentioned in the original description (across all 3 modules)
- [ ] a PR with a SQL view model that emulates `account.move.line` based on the analytic lines with at least an additional column for the analytic plan

rafaelbn:
Hello @sbidoul ,

> I'm going to removing everything that concerns analytic tags and analytic groups. These concepts are gone for good.

For a migration from 15 to 16:
- To which object analytic tags will be migrated in order to maintain all reports working? Actually tags were dimension , important information for clients.
- To which object analytic groups will be migrated in order to maintain all report working? Less important but exists.

Do you have an idea actually or we have still to analyse?

I'm worried about transition. Let me know any help. I will try to record a analysis.

Thank you!

sbidoul:
Hi @rafaelbn if you have information about how the analytic concepts are migrated from 15 to 16 that would help. I very much doubt we can migrate reports that rely on analytic automatically, though.

dzungtran89:
Hi @sbidoul , thanks for your suggestions. The below are two forward ports
- https://github.com/OCA/mis-builder/pull/492
- https://github.com/OCA/mis-builder/pull/493

The reason of having two PRs is to avoid mixing up the commit history of two modules `mis_builder` and `mis_builder_widget`

sbidoul:
I have removed analytic tags and groups.

CasVissers-360ERP:
So I see a few ideas here for analytic account filtering:

1. Use analytic items like in mis_builder_analytic, not a big fan.
2. Fetch analytic accounts from analytic lines related to account.move.line, also a messy workaround
3. [Account Financial Report ](https://github.com/OCA/account-financial-reporting/blob/9d46cd275ecf809a4644ef269dbbf0c0fe07a33f/account_financial_report/models/account_move_line.py#L10) adds a stored field on account.move.line analytic_account_ids, I had to set to store=False because it's heavy to calculate this one a (semi-)big database. See discussion here: https://github.com/OCA/account-financial-reporting/pull/939#discussion_r1053313800 There might be a workaround possible to try to set analytic_account_ids using a query, I can look into this. Don't know exactly how to work with the JSON set-up here but in theory it should be possible.

I do like the idea of having analytic account ids stored on account.move.line but I think it might be good to do this in a generic OCA module, since more modules might be depending on it. cc @pedrobaeza

pedrobaeza:
Yes, that way seems interesting

CasVissers-360ERP:
"{""4"": 50.0, ""33"": 50.0}"
 For mis reporting this is also hell, there should be a debit/credit/balance stored per analytic account on account.move.line to make it feasible/fast enough. We might be able to use the JSON field type to store this and fetch it.

Starting to feel like using account.analytic.line is the only real way for mis reports to work in the short term.

_Other modules might have benefits of stored analytic accounts_

CasVissers-360ERP:
Last comment for today, for one of our customer we actually created a shadow table (account.move.line.reporting) of account.move.line to allow them to do easy, correct and editable multi-currency reporting with stored historic exchange rates (based on a global reporting currency). We fill the table with a cron which has given us a lot of flexibility, we linked to account.move.line with some depends. If we made an error we simply empty the table and let the cron re-populate the table.

Not a sexy set-up, but fits our KISS project model and would offer a solution for this issue as well.

sbidoul:
We need to create a move line source as a view like this:

```sql
SELECT
  aal.id,
  aal.company_id,
  aal.date,
  aal.name,
  aal.general_account_id as account_id,
  aal.account_id as analytic_account_id,
  aal.plan_id,
  aa.root_plan_id,
  CASE
    WHEN amount < 0 THEN -amount
    ELSE 0
  END AS debit,
  CASE
    WHEN amount >= 0 THEN amount
    ELSE 0
  END AS credit,
  aml.parent_state
FROM account_analytic_line aal
LEFT JOIN account_analytic_account aa on aa.id = aal.account_id
LEFT JOIN account_move_line aml on aml.id = aal.move_line_id
WHERE aal.move_line_id IS NOT NULL;
```

When reporting on this the user should be highly encouraged to filter on a root plan to avoid double counting.

CasVissers-360ERP:
Using analytic lines is the same way Odoo is doing analytic reporting but it's not how accountants and controller usually do their reporting. In addition I think any BI system connecting to Odoo will have troubles with this, not on-topic for this PR though.

In old Odoo installations but also new installations where the customer simply overwrites the analytic account on account move lines the reporting would be broken (have seen this multiple times in installations we take-over). The more I think about it the more I prefer a reporting table, which could provide additional features to improve multi-currency reporting.

But since the analytic line set-up is the way Odoo intended it I do agree that it would make most sense to go for the analytic line set-up.

sbidoul:
It's well known that Odoo is not very robust in guaranteeing the analytic balance matches the general balance, and 16.0 does not apparently improve that. But that should be solved elsewhere than in MIS Builder.

dzungtran89:
- btw if we rely to the `account.analytic.line`, which mean we just take the `posted` moves into account ([ref](https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L3329))

- In the `mis_builder_demo` module, I extracted all the analytic account from `analytic_distribution`, you can see [here](https://github.com/OCA/mis-builder/pull/489/commits/03e8de89ab007cfd00ba6fac781cf115ff9df7d5#diff-2760639afe2089745ded0e8303084e10bdc7e4d65c0601dc5c9193b4f198af2cR68) for reference

sbidoul:
> btw if we rely to the account.analytic.line, which mean we just take the posted moves into account ([ref](https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L3329))

That's a good point indeed.

> In the mis_builder_demo module, I extracted all the analytic account from analytic_distribution, you can see [here](https://github.com/OCA/mis-builder/pull/489/commits/03e8de89ab007cfd00ba6fac781cf115ff9df7d5#diff-2760639afe2089745ded0e8303084e10bdc7e4d65c0601dc5c9193b4f198af2cR68) for reference

This approach may be necessary in such cases yes. I worry about the performance, though.
But to be fully correct, the amounts (debit/credit) must also be adjusted according to the analytic distribution.

etobella:
@sbidoul Can we help you somehow?

sbidoul:
@etobella luminous ideas are always welcome :) Otherwise I'm fine, I guess...

image

etobella:
BTW, Odoo is not supporting the option to use analytic lines on non posted moves. So, why don't we do the same in order to simplify it? Or just ignore them when you don't have them posted. You can imagine, that for old database the JSON field will not be filled

sbidoul:
Yes that is a possibility. But I'd like to investigate if an alternative exist so that it could work for other move-line-like tables (such as budget items).

sbidoul:
Okay, I finally decided to remove analytic filters from this PR so it can be tested and merged.
The reason is that I can see no way to keep analytic filtering in a generic enough manner in the core module.

The filtering logic is still present and extensible so I propose to create two basic modules:
- `mis_builder_analytic_account_filter` which introduces a view on `account.analytic.line`, as well as the analytic filters on `mis.report.instance` (making sure such filters are enabled only when an `analytic_account_id` field is present in the move line source)
- `mis_builder_budget_analytic_account` which re-introduce the analytic account field on budget by account items, and depends on  `mis_builder_analytic_account_filter`

These modules will have an important caveat which is they can work with a single analytic plan, but at least we'll have re-introduced a similar feature level as before.

ovnicraft (approved):

rafaelbn:
Hello @sbidoul ,

> Hi @rafaelbn if you have information about how the analytic concepts are migrated from 15 to 16 that would help. I very much doubt we can migrate reports that rely on analytic automatically, though.

https://github.com/OCA/mis-builder/pull/472#issuecomment-1467677377

Odoo Enterprise migration from Odoo 15 to Odoo 16

`account.analytic.group` -> `account.analytic.plan`
`account.analytic.account` -> `account.analytic.account`
`account.analytic.tag` -> Deletes (All data is lost)
`account.analytic.default` -> Deletes (All data is lost)

Thread in twitter: https://twitter.com/rafaodoo/status/1641138242392031238

Really worried about migrate reports and preserver data.

I will edit this post to add a video with explanations

Regards

sbidoul:
I have replaced the analytic filters with a generic domain and search bar.

Here is the report-level filter domain:

image

The domain is available on columns too, so this open possibilities such as creating a single-period report with multiple companies or operating units in columns, and this closes https://github.com/OCA/mis-builder/issues/96 and https://github.com/OCA/mis-builder/issues/183.

image

And here is the search bar in the widget.

image

I still have some tweaking to do, but this is taking good shape, and this should offer a lot of flexibility to do analytic-style filtering, whatever people end up with (re-introduce tags, filter on analytic_distribution, re-introduce analytic dimensions, etc). This should also be resilient to future Odoo changes in this area :stuck_out_tongue:

ovnicraft:
@sbidoul amazing job! Thank you ! 

So drive some smoke test and LGTM

CasVissers-360ERP:
@sbidoul Are you planning on adding the filter widget for mis.report.kpi?

sbidoul:
> Are you planning on adding the filter widget for mis.report.kpi?

@CasVissers-360ERP not right now. It's much more complex because kpi expressions are generic python expressions, so it would require a python expression parser in javascript etc.

sbidoul:
With the look and feel improvement I thinks this is now ready.

image

sbidoul:
I think I'll wait before doing the analytic extension modules I mentioned in https://github.com/OCA/mis-builder/pull/472#issuecomment-1483243741  until I have a clearer view of the use cases.

pedrobaeza (approved):

pedrobaeza:
Should we merge it already?

ovnicraft:
Go head @pedrobaeza 👍🏽

CasVissers-360ERP:
Is there a way to warn users about analytic account not being an option anymore? I think it would avoid some possible bug reports.

`ValueError: Invalid field account.move.line.analytic_account_id in leaf ('analytic_account_id.code', '=', '03000')`

CasVissers-360ERP:
I would have liked some more testing time, but I think as an initial migration it is fine (not underestimating the amount of work that has been done here).

CasVissers-360ERP (approved):

pedrobaeza:
I let @sbidoul the honor of merging when he's ready.

sbidoul:
Let's wait until mid next week to let people test?

pedrobaeza:
OK for me.

etobella (approved):
Functional test locally and it worked like a charm!

sbidoul:
There is probably a bug when changing the move line source in a report template after reports have been created for it. I'll look into it.

sbidoul:
> Is there a way to warn users about analytic account not being an option anymore? I think it would avoid some possible bug reports.
> 
> `ValueError: Invalid field account.move.line.analytic_account_id in leaf ('analytic_account_id.code', '=', '03000')`

More broadly it would be good to show a nicer error message is such case [invalid domain]. This can happen for many reasons. I created https://github.com/OCA/mis-builder/issues/557 to track this.

sbidoul:
@CasVissers-360ERP a friendlier error message is now provided when invalid fields are used in expressions or filter domains. Let me know how it works for you.

sbidoul:
Since the widget is getting a life of its own (thanks OWL for enabling this!), I refactored it's configuration a little bit and moved it to its own tab in the report configuration form.

CasVissers-360ERP:
Functional review:

1. I love the filtering widget and the option to choose a custom search view. Thanks a lot.
2. The help view for KPI's looks a bit weird (nice2have)
image
3. Thanks for the error handling with incorrect domains. Hopefully saves some future inefficiencies.


Now on the analytic filtering:

>> I think I'll wait before doing the analytic extension modules I mentioned in https://github.com/OCA/mis-builder/pull/472#issuecomment-1483243741 until I have a clearer view of the use cases.

Not sure if this is what you mean but our use-case:
A P&L split by business line, for business lines analytic accounts are used. We need to filter in the P&L using analytic accounts. Currently we do this on KPI's using the analytic account filter. 

I was thinking https://github.com/odoo/odoo/pull/116496 might help here, what if we added a filter option on KPI's: "Filter Analytic Accounts" a m2m to account.analytic.account.

When computing amounts, if analytic distribution is set and available on the model, search analytic distributions to find if the analytic account is used. If this is the case we look at the distribution and correct the amount. I know it can be performance heavy, but this is something we can mention in the view? "Using analytic accounts in KPI's might have an impact on performance". Future ideas/changes might fix the performance? 

--- 

Anyhow from my side no blocking issues to merge this PR. Thanks

sbidoul:
@CasVissers-360ERP I fixed the help page. Thanks for the testing.

I'm following https://github.com/odoo/odoo/pull/116496 too. It would definitely help, in situations where each move line is allocated 100% in exactly one analytic account per plan. With proper indexing it may even be performant. I hope they land it.

For analytic distributions the only solution I see is a move-line-like SQL view on analytic lines.

sbidoul:
Let's go :)

/ocabot merge patch

PR #474, opened 2022-10-11
[PORT][15.0][FIX] mis_builder: bug unable to add SUM COL at once

leemannd:
Fixes bug #393 (it's a hack, but it works well)
Fix missing default value on company_id when creating a new report instance with default values

Forward port of https://github.com/OCA/mis-builder/pull/457


gurneyalex:
cc @alexis-via

fclementic2c:
Merci :)

PR #476, opened 2022-11-09
[FIX] mis_builder: let non-admins print reports

yajo:
## Description

Before this patch, normal users would get `AccessError` when trying to preview or print reports.

Towards #415.

@moduon MT-1527

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -475,9 +475,11 @@ class MisReport(models.Model):
     @api.depends("move_lines_source")
     def _compute_account_model(self):
         for record in self:
-            record.account_model = record.move_lines_source.field_id.filtered(
-                lambda r: r.name == "account_id"
-            ).relation
+            record.account_model = (
+                record.move_lines_source.sudo()
+                .field_id.filtered(lambda r: r.name == "account_id")
+                .relation
+            )
 
     @api.onchange("subkpi_ids")
     def _on_change_subkpi_ids(self):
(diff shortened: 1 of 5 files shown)


yajo:
No CI here? @sbidoul

sbidoul:
Dunno why there is no ci, no time to look today.
BTW, I had started this back then https://github.com/OCA/mis-builder/pull/416

yajo:
Yikes, I didn't find it :/. Well, this one includes a smaller change, changelog a test. Maybe you can merge and rebase yours on top of this one? Or how do you prefer to proceed?

mis_builder/readme/newsfragments/access-error.fix
```diff
@@ -0,0 +1 @@
+Fix access error when previewing or printing report.
```
sbidoul: Please rename this file `476.bugfix`, otherwise it will not be considered.

sbidoul:
I'll rebase mine. This looks good to me. Thanks for adding a test.

mis_builder/readme/newsfragments/access-error.fix
yajo: Done, thanks.

yajo:
This is weird. Locally, both pre-commit and odoo tests pass OK. Also in production you can open the mis report. Some ideas? Maybe you can merge this on #416 and continue there?

I rebased, just in case.

sbidoul:
I just pushed a copier update to 15.0 here. A rebase should fix pre-commit.
Regarding the test failure it looks like it's the new test failing?

yajo:
The pre-commit failure is new; must be python 3.11 kicking in.

Regarding the test, yes it's the one that's failing. But since your fix in #416 is better, that's why I think maybe it's a good moment for you to merge it. Probably the test will be OK in #416.

sbidoul:
Ok, cherry-picked in #416. So closing this one.

PR #478, opened 2023-01-03
[WIP][14.0][ADD] mis_builder_company_comparison

JoanSForgeFlow:
## Description

New module mis_builder_company_comparison. 
It allows the generation of reports with each company in each column in order to compare them.
Selection_999(734)


rafaelbn:
I'm not sure @JoanSForgeFlow , but I think this PR should be done in https://github.com/OCA/mis-builder-contrib

JoanSForgeFlow:
Yes, sorry.
I close the PR and I will open it there. 
Thank you @rafaelbn

sbidoul:
Alternatively, this could be in the core mis_builder module. See https://github.com/OCA/mis-builder/issues/183#issuecomment-487834227.

If this goes into the core module I'd prefer it to work without sudo, though.

PR #491, opened 2023-03-09
[15.0][FIX] mis_builder_budget: Change SavepointCase to TransactionCase to avoid the warning in logs

victoralmau:
Change SavepointCase to TransactionCase to avoid the warning in logs

@Tecnativa


pedrobaeza (approved):
Please always remember to indicate the reason (the why, not only the what), which is to avoid the warning in logs.

sbidoul:
Thanks. Merged manually at e2c07cd0.

PR #561, opened 2023-04-14
[14.0][IMP] use cell_style_props for val_rendered in kpi matrix

jonasbuchholz:
So far, the values set in set_values_detail_account were rendered using the row style properties, before evaluating the row's style expression. Thus, even though the correct style properties are passed as cell_style_props, some changes by the style_expression (e.g. prefix, suffix, ...) would not take effect.
With these changes, the style_expression gets evaluated before rendering the value, and the correct cell_style_props are passed to style_model.render.

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -236,21 +236,6 @@ class KpiMatrix(object):
         assert len(vals) == col.colspan
         assert len(drilldown_args) == col.colspan
         for val, drilldown_arg, subcol in zip(vals, drilldown_args, col.iter_subcols()):
-            if isinstance(val, DataError):
-                val_rendered = val.name
-                val_comment = val.msg
-            else:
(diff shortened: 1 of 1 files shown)


sbidoul:
Hi @jonasbuchholz thanks for the contrib.

Can you post a few screenshot with a sample config to help me understand what this fixes ?

jonasbuchholz:
Hi @sbidoul , thank you for your swift response!
Sure! One example would be a KPI that uses different rounding based on its value:
demo_kpi  
The style that is selected for the template in this case is "No Rounding". Without these changes, the rounding property is never applied to the values because the style_expression gets evaluated after they have been rendered:
without_fix  
If the rendering happens after evaluating the style_expression:
with_fix

I hope this helps to explain what these changes do.

sbidoul:
Thanks the change makes sense to me. I've not had time to look into it yet. Wondering how hard it would be to add a test for this.

PR #562, opened 2023-04-21
[16.0][FIX] _compute_source_aml_model_id: self to record

kmagusiak:
This change is necessary to fix an issue when running the migration, `self` can contain multiple records, so move_lines_source could also contain multiple instances and crash the migration.


sbidoul:
Thanks for the fix!

LGTM.

/ocabot merge patch

PR #563, opened 2023-04-25
[14.0][IMP] mis_builder: compute matrix: check if multiple effective companies

ghost:
This PR changes the report when **multi_company** is enabled in a report instance. 

Before, the report **always** included the company name.
`400000 Product Sales [YourCompany]`

Now, **if there is only one** effective company (query_company_ids), the report does not includes the company name.
`400000 Product Sales`

--

I have signed the OCA Contributor License Agreement.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -844,7 +844,8 @@ class MisReportInstance(models.Model):
         """
         self.ensure_one()
         aep = self.report_id._prepare_aep(self.query_company_ids, self.currency_id)
-        kpi_matrix = self.report_id.prepare_kpi_matrix(self.multi_company)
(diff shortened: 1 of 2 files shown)


sbidoul:
Hi, thanks for contributing.

Can you explain why this change is necessary and what the backward compatibility implications are?

ghost:
Hi @sbidoul 

I have a generic report instance which should be used by multiple companies. Therefore multi_company is enabled. Each company runs the report only for the active company. Then it is not necessary to show the company name on each account line. And the report looks bad / cluttered when the company name is long. Therefore I consider this change necessary.

I think there is no backward compatibility implications. And the report pdf has the company name on top, so I see no "compatibility" reason why the company name should be added to each account when there is only one effective company.

PR #567, opened 2023-06-12
[16.0][IMP] mis_builder: use odoo_test_helper in tests

marielejeune:
Make use of odoo-test-helper to correctly import fake model in tests 
See https://github.com/OCA/odoo-test-helper

This fixes Wrong Import warnings, such as 
odoo_test_helper.fake_model_loader: Wrong Import in module mis_builder, the class <class 'odoo.addons.mis_builder.tests.test_kpi_data.TestKpiData.MisKpiDataTestItem'> have been already imported.

### DIFF
--- a/mis_builder/tests/common.py
+++ b/mis_builder/tests/common.py
@@ -6,19 +6,6 @@ import doctest
 from odoo.tests import BaseCase, tagged
 
 
-def setup_test_model(env, model_cls):
-    model_cls._build_model(env.registry, env.cr)
-    env.registry.setup_models(env.cr)
-    env.registry.init_models(
-        env.cr, [model_cls._name], dict(env.context, update_custom_fields=True)
-    )
-
-
-def teardown_test_model(env, model_cls):
-    del env.registry.models[model_cls._name]
-    env.registry.setup_models(env.cr)
-
-
 def _zip(iter1, iter2):
     i = 0
     iter1 = iter(iter1)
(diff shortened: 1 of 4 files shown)


sbidoul:
Nice, thank you!

marielejeune:
@sbidoul tests are green but there is still a warning :
WARNING odoo odoo_test_helper.fake_model_loader: Wrong Import in module mis_builder, the class <class 'odoo.addons.mis_builder.tests.fake_models.MisKpiDataTestItem'> have been already imported.

and an error:
ERROR odoo odoo.modules.registry: Model mis.kpi.data.test.item has no table. 

Not sure how to fix them

mis_builder/tests/__init__.py
```diff
@@ -6,6 +6,7 @@
 from . import test_multi_company_aep
 from . import test_aggregate
 from . import test_data_sources
+from . import fake_models
```
lmignon: @marielejeune The fake models should not be imported otherwise it will pollute the current registry.

lmignon (changes requested):

mis_builder/tests/__init__.py
marielejeune: Obviously ... Thanks :-)

FrancoMaxime (approved):
Very interesting, LGTM

sbidoul:
@legalsylvain would you like to test an ocabot merge here (to check that it is correctly rejected)?

PR #569, opened 2023-06-12
[14.0][IMP] mis_builder: Company not required for report instance

norlinhenrik:
All companies may use - and edit - the same report instance.
The report will be made for the active company.

For this behavior, use these report instance filters:

- Multiple companies: False
- Allowed company: None


norlinhenrik:
Hi @sbidoul 
When all companies are allowed to use the same report instance, I need a new field (e.g. `edit_company_id`) to register which company is allowed to edit the report instance. Do you consider this a generic feature to include in `mis_builder`?

If not, do you think that this PR should be part of `mis_builder` or a custom module?

PR #571, opened 2023-07-27
[16.0][FIX]Do not raise UserError in inverse_expression when using multi

AnizR:
I think the raise of the UserError in `_inverse_expression` should be removed:

This is due to the fact that the inverse function is now (in Odoo16.0) played after `_compute_expression`. This results in not being able to set multiple expressions on kpi's. 

This behaviour is not expected and wasn't the one in previous versions.


sbidoul:
Looks good, thanks!

/ocabot merge patch

PR #573, opened 2023-10-20
[15.0][IMP] mis_builder: test performance improvement

josep-tecnativa:
CC @Tecnativa TT45458

### DIFF
--- a/mis_builder/tests/test_aep.py
+++ b/mis_builder/tests/test_aep.py
@@ -13,21 +13,32 @@ from ..models.aep import AccountingExpressionProcessor as AEP, _is_domain
 
 
 class TestAEP(common.TransactionCase):
-    def setUp(self):
-        super().setUp()
-        self.res_company = self.env["res.company"]
-        self.account_model = self.env["account.account"]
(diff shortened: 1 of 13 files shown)


mis_builder/tests/test_aep.py
```diff
@@ -13,67 +13,81 @@
...
+            date=datetime.date(cls.curr_year, 3, 1),
             amount=500,
-            debit_acc=self.account_ar,
-            credit_acc=self.account_in,
+            debit_acc=cls.account_ar,
+            credit_acc=cls.account_in,
         )
+
+    def setUp(self):
+        super().setUp()
```
pedrobaeza, line 93: As we have talked, put the docstring explaining WHY we do this. If not, the reasons are lost with the wind and coming again to this will make us to repeat the same mistakes.

josep-tecnativa: Done, thanks for the review.

josep-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):

giarve (approved):

PR #565, opened 2023-05-22
[14.0][IMP] Add a generic analytic_domain filter. Backport from v16

marielejeune:
Add a generic analytic_domain filter,
which can be set at the report level or at the column level. The context can contain a mis_analytic_domain key, this domain being always ANDed with the domains configured on the report and columns and in the other filters.

Backport from v16. See commit f2f82a547d6b5998cd60da12ad42a7636fe31cd6

Co-Authored-By: Stéphane Bidoul

PR #575, opened 2023-10-25
[14.0][IMP] mis_builder: Filter period by company

etobella:
Implements #183 inside the core

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -312,6 +312,7 @@ class MisReportInstancePeriod(models.Model):
         default="[]",
         help="A domain to additionally filter move lines considered in this column.",
     )
+    filter_company_ids = fields.Many2many("res.company")
 
     _order = "sequence, id"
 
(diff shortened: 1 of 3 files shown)


JoanSForgeFlow (approved):
Code review, LGMT :+1:

sbidoul:
LGTM. This must not be ported beyond 15 because in 16, the analytic domain covers the use case, is that correct?

etobella:
You are right @sbidoul It seems easier to configure, but it is not necessary

sbidoul:
BTW, we had backport part of the analytic domain feature to 14 in https://github.com/OCA/mis-builder/pull/565.

etobella:
In this case, we can remove this PR. WDYT?

sbidoul:
As you prefer, either we keep it (and forward port it) if you think an easier mechanism is necessary for company filters, or we remove it.

etobella:
With the new mechanism it is not needed, so I would recommend to refuse it

PR #584, opened 2023-11-15
[16.0] Custom fields in accounting expressions (e.g. `fldp.quantity`, `fld.carbon_balance`)

sbidoul:
This is a very rough experiment on the Accounting Expression Processor to support expressions such as `fldp.quantity[60%]`

This will read the quantity field in the same read_group as debit and credit so we can write efficient expressions on any numerical field that is available in the move-line-like table.

TODO
- [x] add tests
- [x] fix failing tests
- [x] update legend
- [x] update docs

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -169,15 +169,19 @@ Building your KPI
 Expressions can be any valid python expressions.
 
 The following special elements are recognized in the expressions to compute accounting
-data: {bal|crd|deb}{pieu}[account selector][journal items domain].
+data: ``{bal|crd|deb|pbal|nbal|fld}{pieu}(.fieldname)?[account selector][journal items domain]``.
 
-* bal, crd, deb: balance, debit, credit.
-* p, i, e: respectively variation over the period, initial balance, ending balance
-* The account selector is a like expression on the account code (eg 70%, etc).
+* ``bal``, ``crd``, ``deb``: balance, debit, credit.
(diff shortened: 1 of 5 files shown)


sbidoul:
This is ready to go. I added tests and documentation.

Very useful to combine quantities (say, hours) with accounting data in the same MIS report.

victoralmau:
I see that the changes added here have lost an interesting functionality (which existed in v16 and exists in v15). 
Before, in comparison mode (2023 vs 2024 for example) accounts with no data will not be shown, now they are all shown (even if they have no data in any of the periods). It seems to be related to https://github.com/OCA/mis-builder/commit/edaee9b3211e245d3c2a704c23d9fec8c9a653ad#diff-c78e2ca554b9359f0e967b6ab457d608b0030631f2b73e2e31f5942457cb11c0R401 but I am not clear about the proper fix, because there are other related changes, can you check it? Thanks.

pedrobaeza:
cc @sbidoul

sbidoul:
Ah, sorry if this caused a regression. I'll have a look in the coming days. Could you create a separate issue and if possible a little more explanation as I'm not sure I fully understand the issue.

victoralmau:
> Ah, sorry if this caused a regression. I'll have a look in the coming days. Could you create a separate issue and if possible a little more explanation as I'm not sure I fully understand the issue.

Of course, created and explained with a use case. https://github.com/OCA/mis-builder/issues/665

PR #585, opened 2023-11-15
[16.0][FIX] mis_builder: correctly compute context in mis_report_widget

IT-Ideas:
This PR ensures that the context used by the `mis_report_widget` widget RPC calls is correctly computed.

Previously using the `mis_analytic_domain` context key was useless as not taken into account in the widget.

### DIFF
--- a/mis_builder/static/src/components/mis_report_widget.esm.js
+++ b/mis_builder/static/src/components/mis_report_widget.esm.js
@@ -3,6 +3,7 @@
 import {Component, onWillStart, useState, useSubEnv} from "@odoo/owl";
 import {useBus, useService} from "@web/core/utils/hooks";
 import {DatePicker} from "@web/core/datepicker/datepicker";
+import {Domain} from "@web/core/domain";
 import {FilterMenu} from "@web/search/filter_menu/filter_menu";
 import {SearchBar} from "@web/search/search_bar/search_bar";
 import {SearchModel} from "@web/search/search_model";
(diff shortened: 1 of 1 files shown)


sbidoul:
This seems to break the filter. On runboat, I tried this filter with the demo report

image

And I got this error

image

The same works on 16.0

IT-Ideas:
> This seems to break the filter. On runboat, I tried this filter with the demo report
> 
> image
> 
> And I got this error
> 
> image
> 
> The same works on 16.0

Oops Indeed I forgot to return the Domain as a list and not as an ast Domain.

IT-Ideas:
@sbidoul it should be ok now 🙂

sbidoul:
Yes, I tested, this works fine now.

Wondering if the search bar widget is capable of loading and displaying the value from the context.

IT-Ideas:
> Yes, I tested, this works fine now.
> 
> Wondering if the search bar widget is capable of loading and displaying the value from the context.

I'll have a look and also about the fact that the filters would be returned to this view when modified on the drilldown one. I am wondering whether we could use the globalstate in order to achieve this but I need a bit more of research.
Or in another PR? Let me know 🙂

sbidoul (approved):

sbidoul:
This can be another PR, no urgency.

IT-Ideas:
@sbidoul Actually, the domain cannot be displayed as this does not correspond to a filter, but if filters are set in the context we should indeed use them. I do not know if it is the case. I'll have a look.

IT-Ideas:
Rebased

mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -104,11 +105,16 @@ export class MisReportWidget extends Component {
     }
 
     get context() {
-        var ctx = super.context;
-        if (this.showSearchBar) {
+        let ctx = this.props.record.context;
```
sbidoul, line 108: @IT-Ideas why is it not a problem to not call super here?

IT-Ideas: @sbidoul  Yes it would for sure. But from what I know, Component doesn't have a context property. Am I wrong?

sbidoul: > Component doesn't have a context property

Indeed :)

PR #587, opened 2023-11-16
[IMP] mis_builder: provide more views in drilldown

IT-Ideas:
Prior to this commit only the `tree` and `form` views where made available while drilldowning in the `mis.report.instance` preview.

After this commit if available, the `tree`, `form`, `pivot` and `chart` views will be provided.

The available views and orders can be easilly overriden.

Thanks for your pull request. Please take a moment to review if it meets the following
guidelines.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -877,6 +877,20 @@ class MisReportInstance(models.Model):
         kpi_matrix = self._compute_matrix()
         return kpi_matrix.as_dict()
 
+    @api.model
+    def _get_drilldown_views_and_orders(self):
+        return {"tree": 1, "form": 2, "pivot": 3, "graph": 4}
+
+    @api.model
+    def _get_drilldown_model_views(self, model_name):
+        self.ensure_one()
+        types = self.env["ir.ui.view"]._read_group(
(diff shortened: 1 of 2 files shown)


mis_builder/models/mis_report_instance.py
```diff
@@ -877,6 +877,21 @@ def compute(self):
...
         return kpi_matrix.as_dict()
 
+    @api.model
+    def _get_drilldown_views_and_orders(self):
+        return {"tree": 1, "form": 2, "pivot": 3, "graph": 4}
+
+    def _get_drilldown_model_views(self):
+        self.ensure_one()
+        types = self.env["ir.ui.view"]._read_group(
+            [("model", "=", self.source_aml_model_name)], ["type"], ["type"]
```
sbidoul: Can you make `_get_drilldown_model_views` a model method, and pass the model name as argument?
Because in the drilldown method, we use `period.source.aml_model_name`.

Can you also add a little unit test for `_get_drilldown_model_views`?

sbidoul (approved):
Tested on runboat with the Journal Items view.

IT-Ideas:
@sbidoul

rvalyi (approved):
LGTM (code review)

PR #589, opened 2023-11-23
[16.0][IMP] allow multi-company MIS budgets by account

sbidoul:
When doing multi-company consolidation with MIS Builder we may want to also display the corresponding budget.

It can therefore be desirable to create a MIS budget by account that applies to multiple companies.

It turns out that making the `company_id` field required for MIS Budgets by Accounts is overly restrictive, and that the really important thing is that the `company_id` on budget items matches the company of the account of the item.


andreampiovesana (approved):
LGTM

PR #591, opened 2023-12-22
[16.0][IMP] mis_builder: take field context into account

IT-Ideas:
[IMP] mis_builder: take field context into account
Prior to this commit only the action context was taken into account
when looking for the `mis_analytic_domain` key in the context.

This commit adds the support of the field context, where `self`
can be used in order to inject record data in the context.

Blocked by https://github.com/OCA/mis-builder/pull/585


IT-Ideas:
Blocked by https://github.com/OCA/mis-builder/pull/585

IT-Ideas:
Rebased

PR #592, opened 2023-12-26
[IMP] mis-builder: set analytic account constraint in the filter when displayed

IT-Ideas:
Up to now, the domain on analytic_account was not visible to the user. After this commit, the domain will be visible in the filters if they are displayed.

Blocked by https://github.com/OCA/mis-builder/pull/591

### DIFF
--- a/mis_builder/__manifest__.py
+++ b/mis_builder/__manifest__.py
@@ -29,6 +29,7 @@
     ],
     "assets": {
         "web.assets_backend": [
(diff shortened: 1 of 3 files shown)


IT-Ideas:
Blocked by #591

IT-Ideas:
This PR is actually specific to `analytic_account_id ` which is absolutely not suitable 😅

PR #593, opened 2023-12-27
[16.0][IMP] mis_builder: preserve the filters when navigating back from drill-down

IT-Ideas:
Prior to this commit the state of the searchModel was lost when navigating
back from a drill-down.

This commit uses the globalState in order to store and restore the state
of the searchModlel. The use of a SubEnv allows not having to pass props
up to the widget, which would have required overriding the renderer, the
compiler and field.

Blocked by #585


IT-Ideas:
@sbidoul

PR #596, opened 2024-02-07
[16.0][FIX] mis_builder: use compute_sudo on field source_aml_model_id to prevent errors due to missing permissions.

JordiBForgeFlow:
The field source_aml_model_id refers to the ir.model, which is a technical in nature, and users should not have naturally permission to read them directly.

## Test => PENDING
Don't forget to add unit tests. If your PR fixes a bug, prefer creating a separate
commit for the test so we one see that your test reproduces the bug and is fixed by the
PR.

## Target branch => PENDING TO CHECK IF PREVIOUS BRANCHES ARE AFFECTED

MIS Builder is actively maintained for Odoo versions 9, 10, 11 and 12.

If your feature is applicable with the same implementation to all these versions, please
target branch 10.0. Maintainers will port it to 9, 11 and 12 soon after merging.

In the rare cases your feature or implementation is specific to an Odoo version, then
target the corresponding branch.

### DIFF
--- a/mis_builder/readme/newsfragments/596.bugfix
+++ b/mis_builder/readme/newsfragments/596.bugfix
@@ -1,0 +1,1 @@
+Resolve a permission issue when creating report periods with a user without admin rights.
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -330,9 +330,11 @@ class MisReportInstancePeriod(models.Model):
     def _check_source_aml_model_id(self):
         for record in self:
             if record.source_aml_model_id:
-                record_model = record.source_aml_model_id.field_id.filtered(
-                    lambda r: r.name == "account_id"
-                ).relation
+                record_model = (
+                    record.source_aml_model_id.sudo()
+                    .field_id.filtered(lambda r: r.name == "account_id")
+                    .relation
+                )
                 report_account_model = record.report_id.account_model
                 if record_model != report_account_model:
                     raise ValidationError(


sbidoul:
Hi @JordiBForgeFlow. Sounds good but normally `compute_sudo` is true by default for stored computed fields: https://github.com/odoo/odoo/blame/60ee2dc2601c1fe7440d69d01844cc7eca4d39df/odoo/fields.py#L448

So I'm curious about the circumstances that trigger a problem related to this.

JordiBForgeFlow:
@sbidoul hmm good point. It failed in two different environments... will check further.

JordiBForgeFlow:
@sbidoul I reproduced the error in runboat:
https://github.com/OCA/mis-builder/assets/7683926/11e46b88-e90e-4a0f-a82c-f8ce3db5d5a5

JordiBForgeFlow:
The field is certainly compute_sudo = True

image

JordiBForgeFlow:
When I add compute_sudo explicity in the field definition, the error does not appear. It's a mistery! Perhaps @MiquelRForgeFlow , the master mistery solver, can help here...

MiquelRForgeFlow:
The problem was not in the compute, but in the `_check_source_aml_model_id` contraint, when accessing technical field `field_id`. Fixed.

mis_builder/readme/newsfragments/596.bugfix
```diff
@@ -0,0 +1,2 @@
+The user may not have permission to access to the field field_id in ir.model, which corresponds to an ir.model.fields record.
+It's a technical issue in nature, so we should use sudo to access it without error.
```
sbidoul:
```suggestion
Resolve a permission issue when creating report periods with a user without admin rights.
```

Let's have a user friendly change log.

MiquelRForgeFlow: Done

sbidoul (approved):

JordiBForgeFlow:
Thanks @sbidoul !

PR #601, opened 2024-02-08
[13.0][FIX] mis_builder: add sudo to access field_id due missing permissions (backport)

MiquelRForgeFlow:
Backport of https://github.com/OCA/mis-builder/pull/596.


MiquelRForgeFlow:
@sbidoul I don't know how to fix this error

pedrobaeza:
In this version, it's not needed (I would say only 15.0+ requires it), as Odoo removes the ir.model user access.

PR #605, opened 2024-04-21
[14.0][IMP] mis_builder_budget: drilldown: default budget & account

norlinhenrik:
Drilldown to "Budget By Account Items": Pass default budget and default account.

If the budget allows items overlap, one can create new budget items in the drilldown.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -944,6 +944,8 @@ class MisReportInstance(models.Model):
                 account_id,
             )
             domain.extend(period._get_additional_move_line_filter())
+            context = dict(self.env.context)
+            context["active_test"] = False
             return {
                 "name": self._get_drilldown_action_name(arg),
                 "domain": domain,
@@ -952,7 +954,7 @@ class MisReportInstance(models.Model):
                 "views": [[False, "list"], [False, "form"]],
                 "view_mode": "list",
                 "target": "current",
-                "context": {"active_test": False},
+                "context": context,
             }
         else:
             return False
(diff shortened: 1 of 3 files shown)


mis_builder/models/mis_report_instance.py
```diff
@@ -944,6 +944,8 @@ def drilldown(self, arg):
                 account_id,
             )
             domain.extend(period._get_additional_move_line_filter())
+            context = arg.get("context") or {}
```
sbidoul: This PR sounds interesting, but why looking for the context in `arg` instead of `self.env.context` ?

mis_builder_budget/models/mis_report_instance.py
```diff
@@ -118,4 +118,9 @@ def drilldown(self, arg):
                     "view_mode": "list",
                     "target": "current",
                 }
+            elif period.source == SRC_MIS_BUDGET_BY_ACCOUNT:
+                arg["context"] = {
+                    "default_budget_id": period.source_mis_budget_by_account_id.id,
+                    "default_account_id": arg.get("account_id"),
+                }
```
sbidoul: Mutating `arg` here does not sound right. Could this be propagated with `with_context` ?

mis_builder/models/mis_report_instance.py
norlinhenrik: Sorry for late reply. I get the context now from frozendict self.env.context.

mis_builder_budget/models/mis_report_instance.py
norlinhenrik: Done.

PR #607, opened 2024-04-30
[17.0][MIG] mis_builder

miquelpascual:
Module migrated to version 17.0

cc https://github.com/APSL 154544

@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 please review


javierobcn (approved):
Works well

etobella (approved):
Tested on runboat and seems to work well, @sbidoul Can we ask for your opinion?

sbidoul:
I'll have a look soon.

I've renamed the PR title :) Can you also reword the migration commit?

Also, can you rebase without 62359ca533ef4c00ed7352ad963c8963b6cab520? This was for python 3.9 support but Odoo 17 does not support it.

mis_builder/models/mis_report.py
```diff
@@ -147,18 +147,17 @@ class MisReportKpi(models.Model):
 
     _order = "sequence, id"
 
-    def name_get(self):
-        res = []
+    @api.depends("description")
```
sbidoul: This depends on `name` too.

mis_builder/models/mis_report.py
```diff
@@ -147,18 +147,17 @@ class MisReportKpi(models.Model):
...
+    def _compute_display_name(self):
         for rec in self:
-            name = f"{rec.description} ({rec.name})"
-            res.append((rec.id, name))
-        return res
+            rec.display_name = f"{rec.description} ({rec.name})"
+            return rec
```
sbidoul: There should be no need to return rec here.

mis_builder/models/mis_report.py
```diff
@@ -147,18 +147,17 @@ class MisReportKpi(models.Model):
...
 
     @api.model
     def name_search(self, name="", args=None, operator="ilike", limit=100):
         domain = args or []
         domain += ["|", ("name", operator, name), ("description", operator, name)]
-        return self.search(domain, limit=limit).name_get()
+        return self.search(domain, limit=limit)._compute_display_name()
```
sbidoul: In 17, the correct way to do this is to override `_name_search`.

mis_builder/models/mis_report.py
```diff
@@ -317,7 +323,8 @@ def name_get(self):
                 )
             else:
                 name = rec.kpi_id.display_name
-            res.append((rec.id, name))
+            rec.display_name = name
+            res.append(rec)
         return res
```
sbidoul: No need to return res here.

mis_builder/models/mis_report.py
```diff
@@ -354,7 +361,7 @@ def name_search(self, name="", args=None, operator="ilike", limit=100):
             ]
         )
         domain = osv_expression.AND([domain, name_search_domain])
-        return self.search(domain, limit=limit).name_get()
+        return self.search(domain, limit=limit)._compute_display_name()
```
sbidoul: This should redone with an override of `_name_search` instead.

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -13,8 +13,7 @@
                         </div>
                         <div class="oe_mis_builder_cp_right_bottom">
                             <div class="oe_mis_builder_filter_buttons">
-                                <FilterMenu t-if="showSearchBar" />
```
sbidoul, line 16: Is it ok to remove this (I have not investigated yet).

mis_builder/tests/test_aep.py
```diff
@@ -363,8 +363,19 @@ def test_inactive_tax(self):
...
-            dict(name="test tax", active=True, amount=0, company_id=self.company.id)
+            dict(
+                name="test tax",
+                active=True,
+                amount=0,
+                company_id=self.company.id,
+                tax_group_id=tax_group.id,
+                country_id=self.env["res.country"]
+                .search([("code", "=", "US")], limit=1)
+                .id,
```
sbidoul: Should this not be the country of `self.company`?

sbidoul (changes requested):
Hi, thanks for working on this migration. Here is a first round of comments.

mis_builder/models/mis_report.py
miquelpascual: Return is needed so we can compare the domain and the return of the function:
`return self.search(domain, limit=limit)._compute_display_name()`
Same in the other _compute_display_name method

mis_builder/static/src/components/mis_report_widget.xml
miquelpascual: [SearchBar](https://github.com/odoo/odoo/blob/e4f2adcf51a990e21bd113478b8cf9849c4f58b6/addons/web/static/src/search/search_bar/search_bar.js#L21) does already provide menu in 17.

mis_builder/tests/test_aep.py
miquelpascual: Company didnt have country set. Added it now.

peluko00 (approved):
Works good, tested in runboat

mis_builder/models/mis_report.py
sbidoul: This override of _name_search is not correct. Actually, there is no need to override in this case. Defining `_rec_name_search = ['name', 'description']` in the model should be sufficient.

mis_builder/models/mis_report.py
```diff
@@ -317,11 +323,12 @@ def name_get(self):
...
             else:
                 name = rec.kpi_id.display_name
-            res.append((rec.id, name))
+            rec.display_name = name
+            res.append(rec)
         return res
 
     @api.model
-    def name_search(self, name="", args=None, operator="ilike", limit=100):
+    def _name_search(self, name="", args=None, operator="ilike", limit=100):
```
sbidoul:
```suggestion
    def _name_search(self, name, domain=None, operator='ilike', limit=None, order=None):
```

This is the signature of `_name_search` in BaseModel.

mis_builder/models/mis_report.py
sbidoul:
```suggestion
        return self._search(domain, limit=limit, order=order)
```

This is what `_name_search` must return.

mis_builder/tests/test_mis_report_instance.py
```diff
@@ -484,25 +484,25 @@ def test_get_kpis_by_account_id(self):
...
             ["kpi 1 / subkpi 1 (k1.sk1)", "kpi 1 / subkpi 2 (k1.sk2)"],
         )
-        r = self.env["mis.report.kpi.expression"].name_search("k1.sk1")
-        self.assertEqual([i[1] for i in r], ["kpi 1 / subkpi 1 (k1.sk1)"])
-        r = self.env["mis.report.kpi.expression"].name_search("k4")
-        self.assertEqual([i[1] for i in r], ["kpi 4 (k4)"])
+        r = self.env["mis.report.kpi.expression"]._name_search("k1.sk1")
+        self.assertEqual([i.display_name for i in r], ["kpi 1 / subkpi 1 (k1.sk1)"])
+        r = self.env["mis.report.kpi.expression"]._name_search("k4")
+        self.assertEqual([i.display_name for i in r], ["kpi 4 (k4)"])
```
sbidoul: I don't think these tests need to change in any way.

mis_builder/models/mis_report.py
sbidoul: That does not sound correct. A compute method should not return a value. See my other comments.

mis_builder/models/mis_report.py
```diff
@@ -147,18 +147,17 @@ class MisReportKpi(models.Model):
...
-    def name_get(self):
-        res = []
+    @api.depends("description", "name")
+    def _compute_display_name(self):
         for rec in self:
-            name = f"{rec.description} ({rec.name})"
-            res.append((rec.id, name))
-        return res
+            rec.display_name = f"{rec.description} ({rec.name})"
+        return rec
```
sbidoul: No need to return in a compute method.

sbidoul:
Could you also cherry-pick https://github.com/OCA/mis-builder/pull/587 ?

mis_builder/models/mis_report.py
miquelpascual: Can't use _search, object is needed instead of only its id

mis_builder/tests/test_mis_report_instance.py
miquelpascual: These tests need to change cause previously `name_get` returned a tuple with both record id and name. Now its setted on display_name

ValentinVinagre:
@manuelregidor @Tisho99 could u review please?

mis_builder/models/mis_report.py
```diff
@@ -354,7 +357,7 @@ def name_search(self, name="", args=None, operator="ilike", limit=100):
             ]
         )
         domain = osv_expression.AND([domain, name_search_domain])
-        return self.search(domain, limit=limit).name_get()
+        return self.search(domain, limit=limit, order=order)
```
sbidoul:
```suggestion
        return self._search(domain, limit=limit, order=order)
```

This is how all `_name_search` are done in Odoo.

mis_builder/models/mis_report.py
```diff
@@ -147,18 +147,16 @@ class MisReportKpi(models.Model):
...
+            rec.display_name = f"{rec.description} ({rec.name})"
 
     @api.model
-    def name_search(self, name="", args=None, operator="ilike", limit=100):
-        domain = args or []
+    def _name_search(self, name, domain=None, operator="ilike", limit=None, order=None):
+        domain = domain or []
         domain += ["|", ("name", operator, name), ("description", operator, name)]
-        return self.search(domain, limit=limit).name_get()
+        return self.search(domain, limit=limit, order=order)
```
sbidoul: This _name_search can be removed and replaced with `_rec_names_search = ["name", "description"]`.

mis_builder/tests/test_mis_report_instance.py
sbidoul: Actually, I'm pretty sure of what I'm saying here. Because the return type of `name_search` has not changed (it still returns a list of (id, display_name) tuples.  So can you please try without changing these two tests at all?

mis_builder/models/mis_report.py
miquelpascual: Done

mis_builder/tests/test_mis_report_instance.py
miquelpascual: Ok, done.

mis_builder/models/mis_report.py
```diff
@@ -354,7 +353,8 @@ def name_search(self, name="", args=None, operator="ilike", limit=100):
             ]
         )
         domain = osv_expression.AND([domain, name_search_domain])
-        return self.search(domain, limit=limit).name_get()
+        records = self.search(domain, limit=limit, order=self._order)
+        return [(record.id, record.display_name) for record in records.sudo()]
```
sbidoul: Sorry but this is still not ok. You must override `_name_search`, not `name_search`, and return `self._search(...)`. There are plenty of examples in the Odoo source code.

mis_builder/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,30 @@
...
+- Arnaud Pineux \<<arnaud.pineux@acsone.eu>\>
+- Ernesto Tejeda \<<ernesto.tejeda@tecnativa.com>\>
+- Pedro M. Baeza \<<pedro.baeza@tecnativa.com>\>
+- Alexey Pelykh \<<alexey.pelykh@corphub.eu>\>
+- Jairo Llopis (<https://www.moduon.team/>)
+- Dzung Tran \<<dungtd@trobz.com>\>
+- Hoang Diep \<<hoang@trobz.com>\>
+- [APSL-Nagarro](<https://apsl.tech>):
+  - Miquel Pascual  \<<mpascual@apsl.net>\>
+  - Antoni Marroig  \<<amarroig@apsl.net>\>
```
sbidoul: Can you preserve the same style as other entries here? Also I have not seen contributions from Antoni Marroig here, why add it?

mis_builder/models/mis_report.py
miquelpascual: Done

mis_builder/readme/CONTRIBUTORS.md
miquelpascual: He's also working on it internally

peluko00:
Any update @sbidoul?

sbidoul:
There are still a couple of things I'd like to check. I'll try to do later this week or early next week.

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -13,8 +13,7 @@
...
-                                <DatePicker
+                                <DateTimePicker
```
sbidoul: This looks likes this, with the full calendar being displayed:

image

It should be a date field with the calendar drop down.

Also I don't think we should use DateTime here, a Date is better.

mis_builder/tests/test_aep.py
```diff
@@ -44,6 +44,10 @@ def setUp(self):
                 "account_type": "income",
             }
         )
+        # Set country to companys
+        self.company.country_id = (
+            self.env["res.country"].search([("code", "=", "US")], limit=1).id
+        )
```
sbidoul: Can you set `country_id` in the company creation statement at line 27 above instead?

mis_builder/tests/test_aep.py
```diff
@@ -363,8 +367,17 @@ def test_inactive_tax(self):
...
+
         tax = self.env["account.tax"].create(
-            dict(name="test tax", active=True, amount=0, company_id=self.company.id)
+            dict(
+                name="test tax",
+                active=True,
+                amount=0,
+                company_id=self.company.id,
+                tax_group_id=tax_group.id,
+                country_id=self.company.country_id.id,
```
sbidoul:
```suggestion
                country_id=self.company.country_id.id,
```

We don't need to set the country here if it is set on company_id.

mis_builder/tests/test_aep.py
miquelpascual: Done

mis_builder/static/src/components/mis_report_widget.xml
miquelpascual: Fixed

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -13,12 +13,11 @@
...
                             <div class="oe_mis_builder_filter_buttons">
-                                <FilterMenu t-if="showSearchBar" />
-                                <DatePicker
-                                    date="state.pivot_date"
-                                    onDateTimeChanged="onDateTimeChanged.bind(this)"
+                                <DateTimeInput
                                     placeholder="'Base date...'"
+                                    type="'date'"
                                     t-if="showPivotDate"
+                                    onChange="onDateTimeChanged.bind(this)"
```
sbidoul, line 21: Does this some sort of binding to `state.pivot_date`?

mis_builder/tests/test_aep.py
```diff
@@ -24,7 +24,14 @@ def setUp(self):
         self.curr_year = datetime.date.today().year
         self.prev_year = self.curr_year - 1
         # create company
-        self.company = self.res_company.create({"name": "AEP Company"})
+        self.company = self.res_company.create(
+            {
+                "name": "AEP Company",
+                "country_id": self.env["res.country"]
+                .search([("code", "=", "US")], limit=1)
+                .id,
```
sbidoul:
```suggestion
                "country_id": (
                    self.env["res.country"]
                    .search([("code", "=", "US")], limit=1)
                    .id
                ),
```

A little detail: this is nicer to read.

mis_builder/static/src/components/mis_report_widget.xml
miquelpascual: Yes, it works exactly as v16. The called function is this one:
`onDateTimeChanged(ev) {
        this.state.pivot_date = ev;
        this.refresh();
    }`

mis_builder/tests/test_aep.py
miquelpascual: pre-commit changes it

sbidoul: It looks better now, thanks.

mis_builder/static/src/components/mis_report_widget.xml
sbidoul: There is something missing, because in 16, you could set the Base Date on the report form, and it was preserved when clicking Preview. 

image

image

image

With this branch, it does not appear in preview. You can check on [runboat](http://oca-mis-builder-17-0-pr607-17b3f9776239.runboat.odoo-community.org/web#id=1&cids=1&menu_id=184&action=292&model=mis.report.instance&view_type=form).

sbidoul:
Ah, another thing, the button box on the report instance form does not look right:

image

image

Probably something with the classes on the button box div.

mis_builder/static/src/components/mis_report_widget.xml
miquelpascual: Fixed. In v16 the DatePicker date attribute was setted with state.pivot_date value. Since DateTimeInput doesnt have this attribute i setted it in value.

miquelpascual:
> Ah, another thing, the button box on the report instance form does not look right:
> 
> image
> 
> image
> 
> Probably something with the classes on the button box div.

Fixed

miquelpascual:
Any updates on this? @sbidoul

sbidoul:
@mpascuall  To me the buttons are still misplaced. In v16 they are on the right. I have not had time to look for a correct solution for this. But the `oe_right oe_button_box` classes we use here don't seem to work correctly in v17. Also it seems weird to add a `btn` class on `button` tags, but again I've not looked closely yet.

miquelpascual:
> @mpascuall To me the buttons are still misplaced. In v16 they are on the right. I have not had time to look for a correct solution for this. But the `oe_right oe_button_box` classes we use here don't seem to work correctly in v17. Also it seems weird to add a `btn` class on `button` tags, but again I've not looked closely yet.

Fixed, could you check it again please?

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
```
flachica, line 1:
```suggestion
```
This line launch warning of deprecation

flachica (changes requested):

HaraldPanten:
@sbidoul Can our team help you in anything with this module? I'm sure it's close to be merged 😉

sbidoul:
Yes, it's close. I'll review the latest changes after my holidays. Also check if it works in wizard mode.

HaraldPanten:
> Yes, it's close. I'll review the latest changes after my holidays. Also check if it works in wizard mode.

Nice. Enjoy your holidays 👍

sbidoul (approved):
Looks good now. Many thanks!

PR #588, opened 2023-11-17
[14.0][IMP] add: css sticky table header added for mis-builder-widget

dora-jurcevic:
Hello ! 
Here is the modification of the CSS of the table header on mis-builder preview, making the header stick on the top of the page. Only CSS has been modified. 
Thank you in advance for the review,
This is my first PR, so please be gentile :)

PR #610, opened 2024-05-28
[15.0][IMP] add: css sticky table header added for mis-builder-widget

hitrosol:
i forward the improvement in v14 from https://github.com/OCA/mis-builder/pull/588


sbidoul:
Merging this. Thanks for the port.

Unfortunately this approach does not work in 16+.

/ocabot merge patch

PR #611, opened 2024-05-30
[FIX] mis_builder: use sudo when calling ir.ui.view

cav-adhoc:
When you try to access to account move lines inside a MIS report with an user that has no admin rights, it returns an Access Error.
That is because the function _get_drilldown_model_views() is trying to get the types by calling ir.ui.view model without sudo permitions.

Here is a video replicating the error: https://app.screencastify.com/v3/watch/mw9vsXxOosnvNKLSfKtc
Regads

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -885,8 +885,10 @@ class MisReportInstance(models.Model):
     @api.model
     def _get_drilldown_model_views(self, model_name):
         self.ensure_one()
-        types = self.env["ir.ui.view"]._read_group(
-            [("model", "=", model_name)], ["type"], ["type"]
+        types = (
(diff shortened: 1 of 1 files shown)


jjscarafia (approved):
LGTM

jjscarafia:
@IT-Ideas Can you please check thise fix.
Seems it's related to this improvement https://github.com/OCA/mis-builder/pull/587

cc @sbidoul

sbidoul (approved):
Thanks for this fix!

IT-Ideas:
> @IT-Ideas Can you please check thise fix.
> 
> Seems it's related to this improvement https://github.com/OCA/mis-builder/pull/587
> 
> 
> 
> cc @sbidoul 

Thanks for the fix!

PR #621, opened 2024-08-07
[16.0][ADD] mis_builder: add option on report instance to drilldown in popup

AnizR:
When drilling down on a cell inside a report, open it in a new pop-up window instead of current window to avoid reloading the report when going back.

It is customizable on each report instance: 
image

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -580,6 +580,9 @@ class MisReportInstance(models.Model):
         string="Filter box search view",
         help="Search view to customize the filter box in the report widget.",
     )
+    drilldown_open_in_pop_up = fields.Boolean(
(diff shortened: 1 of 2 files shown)


acsonefho (approved):

sbidoul:
Let's add this to the core module, with a Popup drilldown option.

acsonefho (approved):

sbidoul (approved):

SAnnabelle (approved):

sbidoul:
Setting to draft. The popup only partially solves the original issue, which is the recomputation time when navigating back to the report after drilldown. Moreover the ergnomics of the popup is not optimal as a typical move lines tree view does not fit nicely in the popup.

sbidoul:
We could consider #633 instead

Issue #614, opened 2024-07-11
Set styles by sub-kpi


SAnnabelle:
It is possible for users to need to round a sub-kpi to 2 decimal and another sub-kpi to 3 decimal. 

The solution would be to be able to set styles per sub-kpi or to use a style expression on the sub-kpi.

PR #623, opened 2024-08-26
[IMP][16.0] mis_builder: allow definition of style expression on subkpi level

AnizR:
## Description

Solves issue: https://github.com/OCA/mis-builder/issues/614

The goal is to be able to define a style on a cell. Therefore, this PR allows to define a style on a subkpi:
image
On my row, I will have my cell with `KPI_1` in red and my cell with `KPI_2` without any style.

## Remark
There are now 3 ways to define a style:

1. `Style` on a KPI
2. `Style expression` on a KPI
3. `Style expression` in expressions of a KPI

The sytle defined in [3] has priority over [2] which has priority over [1].

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -235,32 +235,30 @@ class KpiMatrix:
         assert len(vals) == col.colspan
         assert len(drilldown_args) == col.colspan
         for val, drilldown_arg, subcol in zip(vals, drilldown_args, col.iter_subcols()):  # noqa: B905
-            if isinstance(val, DataError):
-                val_rendered = val.name
-                val_comment = val.msg
-            else:
-                val_rendered = self._style_model.render(
-                    self.lang, row.style_props, kpi.type, val
-                )
-                if row.kpi.multi and subcol.subkpi:
-                    val_comment = "{}.{} = {}".format(
-                        row.kpi.name,
-                        subcol.subkpi.name,
-                        row.kpi._get_expression_str_for_subkpi(subcol.subkpi),
-                    )
-                else:
-                    val_comment = f"{row.kpi.name} = {row.kpi.expression}"
             cell_style_props = row.style_props
-            if row.kpi.style_expression:
+
+            style_expression = ""
+            if (
+                row.kpi.multi
+                and subcol.subkpi
+                and row.kpi.expression_ids.filtered(
+                    lambda rec, subkpi=subcol.subkpi: rec.subkpi_id == subkpi
+                ).style_expression
+            ):
+                style_expression = row.kpi.expression_ids.filtered(
+                    lambda rec, subkpi=subcol.subkpi: rec.subkpi_id == subkpi
+                ).style_expression
+            elif row.kpi.style_expression:
+                style_expression = row.kpi.style_expression
+
+            if style_expression:
                 # evaluate style expression
                 try:
-                    style_name = mis_safe_eval(
-                        row.kpi.style_expression, col.locals_dict
-                    )
+                    style_name = mis_safe_eval(style_expression, col.locals_dict)
                 except Exception:
                     _logger.error(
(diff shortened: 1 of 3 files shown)


mis_builder/models/kpimatrix.py
```diff
@@ -251,16 +251,29 @@ def set_values_detail_account(
...
+
+            style_expression = ""
+            if (
+                row.kpi.multi
+                and subcol.subkpi
+                and row.kpi.expression_ids.filtered(
+                    lambda rec, subkpi=subcol.subkpi: rec.subkpi_id == subkpi
+                ).style_expression
+            ):
+                style_expression = row.kpi.expression_ids.filtered(
```
acsonefho, line 248: Sorry if my question is not relevant but how do you know that you'll have only one result after the `filtered(...)` ?

AnizR: We must have one expression per subkpi: https://github.com/OCA/mis-builder/blob/16.0/mis_builder/models/mis_report.py#L301

acsonefho (approved):

AnizR:
I've add a small fix: when a style expression was defined, it was only used to set style of cell but not style of number (it was ignoring rounding,prefix, etc.)

acsonefho (approved):

SAnnabelle (approved):
Functionally, it's correct. 

image

mis_builder/models/kpimatrix.py
```diff
@@ -235,32 +235,30 @@ def set_values_detail_account(
...
-                        row.kpi.name,
-                        subcol.subkpi.name,
-                        row.kpi._get_expression_str_for_subkpi(subcol.subkpi),
-                    )
-                else:
-                    val_comment = f"{row.kpi.name} = {row.kpi.expression}"
             cell_style_props = row.style_props
-            if row.kpi.style_expression:
+
+            style_expression = ""
```
sbidoul, line 240:
```suggestion
            style_expression = None
```

sbidoul (changes requested):
Can you also add a style on the `mis.report.kpi`, for completeness and symmetry with the styling of the kpi?

sbidoul:
> Can you also add a style on the `mis.report.kpi`, for completeness and symmetry with the styling of the kpi?

Or, before doing that can you explain why you chose to do a style expression instead of a style m2o. The latter seems easier to use.

AnizR:
> Can you also add a style on the `mis.report.kpi`, for completeness and symmetry with the styling of the kpi?

I suppose that you meant `mis.report.kpi.expression`

> Or, before doing that can you explain why you chose to do a style expression instead of a style m2o. The latter seems easier to use.

The style seems 'easier' to use but in my case, it wasn't enough: I want to be able to change my style depending on multiple conditions.

Nonetheless, I'll add it for completeness

sbidoul:
> > Can you also add a style on the `mis.report.kpi`, for completeness and symmetry with the styling of the kpi?
> 
> I suppose that you meant `mis.report.kpi.expression`

Yes. A style on `mis.report.kpi` might be useful, but it's another story.

AnizR:
> > > Can you also add a style on the `mis.report.kpi`, for completeness and symmetry with the styling of the kpi?
> > 
> > 
> > I suppose that you meant `mis.report.kpi.expression`
> 
> Yes. A style on `mis.report.kpi` might be useful, but it's another story.

It is already implemented on `mis.report.kpi`: https://github.com/OCA/mis-builder/blob/bc5b7d142afa5d660bc82d9e262aee9de13efb94/mis_builder/models/mis_report.py#L106

But indeed, adding it on `mis.report.kpi.expression` is complex

The main issue is the way it is injected: the style is set at the row level and not the cell level!

sbidoul:
I'm setting this as draft, as I think a more profound refactoring is needed to do this correctly. See #647

PR #625, opened 2024-08-27
[16.0][IMP] mis_builder: support multi company on non-accounting queries

AnizR:
## Description

Adds the possibility to specify a company field on queries.
This field will be used to filter records that are on the company of the instance report (using 'query_company_ids')

### DIFF
--- a/mis_builder/models/mis_report.py
+++ b/mis_builder/models/mis_report.py
@@ -392,12 +392,19 @@ class MisReportQuery(models.Model):
     date_field = fields.Many2one(
         comodel_name="ir.model.fields",
         required=True,
-        domain=[("ttype", "in", ("date", "datetime"))],
+        domain="[('ttype', 'in', ('date', 'datetime')),('model_id', '=', model_id)]",
         ondelete="cascade",
     )
     domain = fields.Char()
     report_id = fields.Many2one(
         comodel_name="mis.report", required=True, ondelete="cascade"
+    )
+    company_field_id = fields.Many2one(
+        comodel_name="ir.model.fields",
+        ondelete="set null",
+        domain="[('model_id', '=', model_id)]",
+        help="Field that defines company on related model."
+        "When set, it will be automatically be added in search domain of query.",
     )
 
     _order = "name"
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -407,7 +407,12 @@ class MisReportInstancePeriod(models.Model):
         Returns an Odoo domain expression (a python list)
         compatible with the model of the query."""
         self.ensure_one()
-        return []
+        domain = []
+        if company_field := query.sudo().company_field_id:
+            query_company_ids = self.report_instance_id.query_company_ids.ids
+            assert query_company_ids
+            domain = [(company_field.name, "in", query_company_ids)]
+        return domain
 
     @api.constrains("mode", "source")
     def _check_mode_source(self):
--- a/mis_builder/tests/test_mis_report_instance.py
+++ b/mis_builder/tests/test_mis_report_instance.py
@@ -633,3 +633,22 @@ class TestMisReportInstance(common.HttpCase):
             self.env, "mis_you", groups="base.group_user,account.group_account_readonly"
         )
         self.report_instance.with_user(test_user).compute()
+
+    def test_query_company(self):
+        c1 = self.report_instance.company_id
+
+        query = self.report.query_ids
+        self.assertEqual(len(query), 1)
+
+        period = self.report_instance.period_ids[0]
+        domain = period.with_company(c1)._get_additional_query_filter(query)
+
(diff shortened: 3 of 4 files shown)


acsonefho (approved):

SAnnabelle (approved):
Functionally it's correct

mis_builder/models/mis_report.py
```diff
@@ -399,6 +399,13 @@ def _compute_field_names(self):
     report_id = fields.Many2one(
         comodel_name="mis.report", required=True, ondelete="cascade"
     )
+    company_field = fields.Many2one(
+        comodel_name="ir.model.fields",
+        ondelete="cascade",
```
sbidoul: not sure about cascade here

AnizR: Indeed, set null seems more appropriate.

mis_builder/models/mis_report.py
```diff
@@ -399,6 +399,13 @@ def _compute_field_names(self):
     report_id = fields.Many2one(
         comodel_name="mis.report", required=True, ondelete="cascade"
     )
+    company_field = fields.Many2one(
```
acsonefho: `company_field_id` ?

acsonefho (changes requested):

acsonefho (approved):

mis_builder/models/mis_report_instance.py
```diff
@@ -406,7 +407,12 @@ def _get_additional_query_filter(self, query):
         Returns an Odoo domain expression (a python list)
         compatible with the model of the query."""
         self.ensure_one()
-        return []
+        domain = []
+        if (company_field := query.sudo().company_field_id) and (
+            instance_companies := self.report_instance_id.query_company_ids
+        ):
+            domain = AND([domain, [(company_field.name, "in", instance_companies.ids)]])
```
sbidoul: Why doing an AND with empty domain?

mis_builder/models/mis_report_instance.py
```diff
@@ -406,7 +407,12 @@ def _get_additional_query_filter(self, query):
         Returns an Odoo domain expression (a python list)
         compatible with the model of the query."""
         self.ensure_one()
-        return []
+        domain = []
+        if (company_field := query.sudo().company_field_id) and (
+            instance_companies := self.report_instance_id.query_company_ids
```
sbidoul: I think `query_company_ids` cannot be empty. And if it was, it would be surprising if the company field on the query was ignored. So may be we could simply `assert self.report_instance_id.query_company_ids` if `company_field_id` is set.

mis_builder/models/mis_report.py
```diff
@@ -399,6 +399,13 @@ def _compute_field_names(self):
     report_id = fields.Many2one(
         comodel_name="mis.report", required=True, ondelete="cascade"
     )
+    company_field_id = fields.Many2one(
+        comodel_name="ir.model.fields",
+        ondelete="set null",
+        domain="[('model_id', '=', model_id)]",
```
sbidoul, line 405: This domain seems correct, but I wonder why it is not on the `date_field` above.

sbidoul (commented):
Looks good. A couple of questions.

sbidoul:
@AnizR when you have a chance, could you check my review comments?

AnizR:
> @AnizR when you have a chance, could you check my review comments?

Yes, I'll free some time to do that this week

mis_builder/models/mis_report_instance.py
AnizR: Indeed, it is useless!

mis_builder/models/mis_report_instance.py
AnizR: Right, I'll adapt

mis_builder/models/mis_report.py
AnizR: Ok, I'll add it on `date_field`

mis_builder/models/mis_report_instance.py
```diff
@@ -407,7 +407,12 @@ def _get_additional_query_filter(self, query):
         Returns an Odoo domain expression (a python list)
         compatible with the model of the query."""
         self.ensure_one()
-        return []
+        domain = []
+        if company_field := query.sudo().company_field_id:
+            instance_companies = self.report_instance_id.query_company_ids
+            assert instance_companies
+            domain = [(company_field.name, "in", instance_companies.ids)]
```
```suggestion
            query_company_ids = self.report_instance_id.query_company_ids.ids
            assert query_company_ids
            domain = [(company_field.name, "in", query_company_ids)]
```

Just a nitpick, but in general I prefer to keep variable names so we have a uniform vocabulary.

AnizR: Done :+1:

sbidoul:
/ocabot merge minor

Nice improvement!

PR #628, opened 2024-08-27
[MIG] mis_builder_demo: migration to 17.0

dreispt:


mis_builder_budget/models/mis_budget_by_account_item.py
```diff
@@ -5,7 +5,11 @@
 
 
 class MisBudgetByAccountItem(models.Model):
-    _inherit = ["mis.budget.item.abstract", "prorata.read_group.mixin"]
+    _inherit = [
+        "mis.budget.item.abstract",
+        "prorata.read_group.mixin",
+        "analytic.mixin",
```
sbidoul: Adding the analytic mixin goes beyond the scope of a migration. Can we do this in another PR?

dreispt: Sure: https://github.com/OCA/mis-builder/pull/634

sbidoul (approved):

Issue #617, opened 2024-07-23
Empty data after the report reference date


SAnnabelle:
Generate a report that contains both columns based on fixed dates (e.g. the months of the year) and columns based on the report reference date (YTD). Generally, accounting reports are analysed from 01/01 to the month preceding the current month. For example, we're analysing from 1 January to 31 March, even though it's April. The problem is that the data for the fixed-date columns shows data up to and including April, but the YTD column show data from January to March because the report reference date is 31 March. This can lead to confusion when interpreting the report. 

A first solution that had been considered was to use filters on the desired dates. However, this only filters the data for accounting entries, not the data for queries. 

The proposed improvement is to add a configuration in the columns to determine whether we should have empty data after the report reference date. For example, if we do an analysis in April on the data from January to March, the April data for the fixed-date type columns would not be displayed for April because we don't display the data after the report reference date, which will be 31 March. 

![image](https://github.com/user-attachments/assets/bea32fc7-7e5d-4b79-a008-c82b0bef4c2b)

PR #629, opened 2024-08-28
[IMP][16.0] mis_builder: add possibility to dynamically hide a period depending on instance date

AnizR:
## Description

Related to issue: https://github.com/OCA/mis-builder/issues/617

This PR adds a boolean 'hide_period_based_on_instance_date' on periods of an instance.
When activated, the period will be shown only if the date of the instance is greater or equal to the end date of the period.

image

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -284,6 +284,17 @@ class MisReportInstancePeriod(models.Model):
         help="A domain to additionally filter move lines considered in this column.",
     )
 
+    hide_period_based_on_instance_date = fields.Boolean(
+        string="Dynamically Hide Column",
+        help="""Hide this column if the pivot date is before the end date.
+        In the case of a column computed from other columns (e.g. compare column),
+        the column is displayed if all subcolumns should be displayed.
+        """,
+        compute="_compute_hide_period_base_on_instance_date",
+        store=True,
+        readonly=False,
+    )
+
     _order = "sequence, id"
 
     _sql_constraints = [
@@ -299,6 +310,14 @@ class MisReportInstancePeriod(models.Model):
             "Period name should be unique by report",
         ),
     ]
+
+    @api.depends("source")
+    def _compute_hide_period_base_on_instance_date(self):
+        for rec in self:
+            # Coluwns with source 'Actuals' or 'Actuals (alternative)'
+            # will be displayed depending on their subcolumns
+            if rec.source in (SRC_CMPCOL, SRC_SUMCOL):
+                rec.hide_period_based_on_instance_date = True
 
     @api.depends("source", "report_instance_id.report_id.move_lines_source")
     def _compute_source_aml_model_id(self):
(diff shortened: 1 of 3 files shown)


acsonefho (approved):

mis_builder/models/mis_report_instance.py
```diff
@@ -284,6 +284,10 @@ def _compute_dates(self):
         help="A domain to additionally filter move lines considered in this column.",
     )
 
+    hide_period_based_on_instance_date = fields.Boolean(
+        help="Dynamically hide this period depending on the base date of the instance",
+    )
```
sbidoul, line 296: By just reading the field name and help text I don't understand what this does. Can you propose something more explanatory ?

Is it not better to talk about pivot date rather than base date? Not sure.

SAnnabelle (approved):
Functionally it's correct.

mis_builder/models/mis_report_instance.py
```diff
@@ -852,7 +869,10 @@ def _compute_matrix(self):
         self.ensure_one()
         aep = self.report_id._prepare_aep(self.query_company_ids, self.currency_id)
         kpi_matrix = self.report_id.prepare_kpi_matrix(self.multi_company)
-        for period in self.period_ids:
+
+        periods = self._get_periods()
+
+        for period in periods:
```
```suggestion

        for period in self._get_periods():
```

mis_builder/models/mis_report_instance.py
```diff
@@ -843,6 +847,19 @@ def _add_column(self, aep, kpi_matrix, period, label, description):
         elif period.source == SRC_CMPCOL:
             return self._add_column_cmpcol(aep, kpi_matrix, period, label, description)
 
+    def _get_periods(self):
+        periods = self.env["mis.report.instance.period"]
```
```suggestion
        periods = self.env["mis.report.instance.period"].browse()
```

or use `filter()` ?

mis_builder/views/mis_report_instance.xml
```diff
@@ -414,6 +414,7 @@
                         options="{'model': 'source_aml_model_name'}"
                         attrs="{'invisible': [('source_aml_model_name', '=', False)]}"
                     />
+                    <field name="hide_period_based_on_instance_date" />
```
sbidoul: I think this field should be hidden for some period types. For instance it does not make sense for comparison columns?

sbidoul (changes requested):
A few comments to handle.

mis_builder/models/mis_report_instance.py
AnizR: It tried to improve it, what do you think?

mis_builder/models/mis_report_instance.py
AnizR: I don't know why I used a loop, I think that a `.filtered()` is cleaner, indeed.

mis_builder/views/mis_report_instance.xml
AnizR: I've hide the field for comarison columns, is there any other case where I should hide this feature?

EDIT: I will do some changes https://github.com/OCA/mis-builder/pull/629#issuecomment-2415839583

AnizR:
I have to make some changes:
- If the field to dynamically hide period is not set to true, the col is always displayed
- If the mode is fix, the period is displayed if the pivot date is before the end
- A period with a source 'Acutals' or 'Actuals (alternative)' will always be displayed if the mode is not 'fix'
- A period that sums columnns will be displayed if its 'source columns' should be displayed
- A period that compares columns will be displayed if its 'source columns' should be displayed

mis_builder/models/mis_report_instance.py
```diff
@@ -458,6 +466,45 @@ def copy_data(self, default=None):
...
+         - A period that sums columnns will be displayed if its 'source
+           columns' should be displayed
+         - A period that compares columns will be displayed if its
+           'source columns' should be displayed
+        """
+        self.ensure_one()
+        if not self.hide_period_based_on_instance_date:
+            return True
+
+        if self.mode == "fix":
```
acsonefho: `MODE_FIX` ?

acsonefho (changes requested):

acsonefho (approved):

Issue #620, opened 2024-08-07
Multi-company budget for a multi-company report - add a company field to budget lines


SAnnabelle:
Issues : 

To have a mis report with a budget column available for companies A and B. The budget must be adapted according to the company in which you are working. We are therefore obliged to use a budget that is configured without a company. Because if you set a budget associated with company A, you will only see data in the budget column when you are connected to company A and not when you are connected to company B. 

Currently, there are no companies on the budget lines. So if you configure a budget without a company and put it in the budget column of the report, you will always see the same amount, whether you are logged in to company A or company B. This is not intended, as the budget is different depending on the company.

Solution: 
Add a company field to the budget lines. It must take the value of the company on the budget (if defined) and if it is not defined on the budget, then it must be defined manually on the lines.
In the report, if a column has a budget, the amount displayed in the budget column must be the sums of the budget lines associated with the query companies in the report, i.e. query_companys_ids.

sbidoul:
This is for Budgets by KPI, correct?

Related #589

SAnnabelle:
Yes it's for budgets by KPI

sbidoul:
This was done in #630

PR #630, opened 2024-08-28
[IMP] mis_builder_budget: allow to specify a company on budget by kpi

AnizR:
Fixes issue: https://github.com/OCA/mis-builder/issues/620

The goal here is to be able to have one 'budget by kpi' for multiple companies.
Therefore, we need to be able to specify to which company a 'budget item' belongs

### DIFF
--- a/mis_builder_budget/models/mis_budget_item.py
+++ b/mis_builder_budget/models/mis_budget_item.py
@@ -19,10 +19,30 @@ class MisBudgetItem(models.Model):
         )
     )
 
+    company_id = fields.Many2one(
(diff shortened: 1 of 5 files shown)


acsonefho (approved):

AnizR:
Small fix: I forgot to adapt the ir.rule to take into account company of `mis.budget.item` instead of using company of its related `mis.budget`

SAnnabelle (approved):
Functionally it's correct

sbidoul (approved):
LGTM

PR #632, opened 2024-09-04
[16.0][FIX] mis_builder: Add analytic_domain to report instance only if it is set

victoralmau:
Changes done:
- [x] Add `analytic_domain` to report instance only if it is set
- [x] Remove old migration (8.0.2.0.0.0)
- [x] Add migration script to set the value of the `analytic_domain` field

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT50696


pilarvargas-tecnativa (approved):

mis_builder/migrations/16.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,33 @@
...
+    env.cr.execute(
+        """
+        SELECT id, analytic_account_id
+        FROM mis_report_instance_period
+        WHERE analytic_account_id IS NOT NULL
+        """
+    )
+    for item_id, analytic_account_id in env.cr.fetchall():
+        domain = [("analytic_account_id", "=", analytic_account_id)]
+        period_model.browse(item_id).write({"analytic_domain": domain})
```
pedrobaeza: Do it everything in SQL:

```suggestion
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE mis_report_instance_period
        SET analytic_domain = CONCAT('[("analytic_account_id", "=", ', analytic_account_id::VARCHAR, ')]')
        WHERE analytic_account_id IS NOT NULL
        """
    )
```

pedrobaeza (approved):

mis_builder/migrations/16.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+        UPDATE mis_report_instance_period
+        SET analytic_domain = CONCAT('[("analytic_account_id", "=", ', analytic_account_id::VARCHAR, ')]')
+        WHERE analytic_account_id IS NOT NULL
+        """,  # noqa: E501
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE mis_report_instance
+        SET analytic_domain = CONCAT('[("analytic_account_id", "=", ', analytic_account_id::VARCHAR, ')]')
```
sbidoul: Does this domain work with out of the box v16 accounting?

pedrobaeza: This is applied over `account.move.line`, isn't it? Indeed, I don't find a field `analytic_account_id` on such model.

victoralmau: You are right, I change it to `analytic_line_ids.account_id` to make it correct, ok?

pedrobaeza: Well, I see no other option. Good for you, @sbidoul ?

sbidoul: I see there is an `analytic_distribution_search` field on `analytic.mixin`. That was added after 16.0 was released, I think. Maybe it's best to use that?

sbidoul:
/ocabot merge patch

Thanks for this. Do you plan to port it to 17?

PR #634, opened 2024-09-11
[IMP] mis_builder_budget: allow Analytic Accounts on account budget

dreispt:


sbidoul:
This makes sense to me.  

Out of curiosity, how do you use that in practice in actual reports? Or does that require additional custom modules to be of practical use?

dreispt:
No additional modules needed.
With this you can add filters to your MIS Reports, that then can apply to both Account Moves and Budget Items, to have reports by Analytic Account.

sbidoul:
Ah yes, the analytic mixin now provides a search function. Nice!

sbidoul (approved):

PR #636, opened 2024-10-01
[17.0][FIX] mis_builder: Add analytic_domain to report instance only if it is set

victoralmau:
FWP from 16.0: https://github.com/OCA/mis-builder/pull/632

Changes done:
- [x] Add `analytic_domain` to report instance only if it is set

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT50696

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -719,7 +719,8 @@ class MisReportInstance(models.Model):
             )
         )
         # report-level analytic domain filter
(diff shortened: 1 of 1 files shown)


sbidoul:
Since the migration commit is in 16, does it need to stay in 17?

yvaucher (approved):
LGTM

Code review

pilarvargas-tecnativa (approved):

PR #644, opened 2024-10-18
[16.0][IMP] mis_builder: add read access to model and fields to group account manager

AnizR:
## Description

Editing a query in a template can throw an access error.

This happens when a 'Billing administrator' (group_account_manager) hasn't the group 'Access Rights' (group_erp_manager).
In fact, only the group 'Access Rights' gives a read access to `ir_model` and `ir_model_fields`.
But, on a `mis.report.query`, the user needs those access to be able to set `model_id` `and field_ids`


pedrobaeza (changes requested):
This is not correct. We shouldn't grant that permission, but bypass the restriction when applicable. I know it's a direct m2o, but we should find another way. Giving permissions is not the way.

AnizR:
> This is not correct. We shouldn't grant that permission, but bypass the restriction when applicable. I know it's a direct m2o, but we should find another way. Giving permissions is not the way.

Ok, how would you do such thing then?

I had the idea to 'dynamically build a selection containing the name of models/fields'. This selection would be computed in sudo and would set the m2o. What do you think?

Do you have any other idea?

pedrobaeza:
Yes, that's what we have done in other places.

sbidoul:
I'm not sure we can solve that in mis_builder, because any solution will weaken the security baseline by exposing the list of models and fields.

AnizR:
This wasn't a good idea.
Let's discuss the best way to tackle this issue: https://github.com/OCA/mis-builder/issues/415

PR #646, opened 2024-11-08
[17.0][IMP] mis_builder: add support for branch companies

### LINKED ISSUE #645: Issue with branches on report matrix
## Module
mis_builder

## Describe the bug
When using branches, the total debit/credit on the report line is not computed correctly.
Drill down works fine though.

This is due to branches using the same chart of account as the mother company (same account_id).
I'll propose a PR soon.

mis_report_matrix
mis_report_drill_down

## To Reproduce
**Affected versions**: 17

1. Create a company with branches and some invoices for each company/branches
2. Create a simple mis report template with income kpi -balp[4%]
3. Create a multi company/branch mis report (everything in allowed companies)

**Expected behavior**
The report income line should take into account all company/branche income.

jguenat:
Attempt to fix #645

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -12,6 +12,7 @@ from odoo.tools.float_utils import float_is_zero
 from odoo.tools.safe_eval import datetime, dateutil, safe_eval, time
 
 from .accounting_none import AccountingNone
+from .simple_array import SimpleArray
 
 _logger = logging.getLogger(__name__)
 
@@ -316,7 +317,7 @@ class AccountingExpressionProcessor:
         aml_model = aml_model.with_context(active_test=False)
         company_rates = self._get_company_rates(date_to)
         # {(domain, mode): {account_id: (debit, credit)}}
-        self._data = defaultdict(dict)
+        self._data = defaultdict(lambda: defaultdict(lambda: SimpleArray((0.0, 0.0))))
         domain_by_mode = {}
         ends = []
         for key in self._map_account_ids:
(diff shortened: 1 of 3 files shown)


mis_builder/models/aep.py
```diff
@@ -364,7 +364,18 @@ def do_queries(
                 ):
                     # in initial mode, ignore accounts with 0 balance
                     continue
-                self._data[key][acc["account_id"][0]] = (debit * rate, credit * rate)
```
sbidoul, line 367: If self._data becomes `defaultdict(lambda: defaultdict(lambda: SimpleArray((0.0, 0.0))))`, this can become `self._data[key][acc["account_id"][0]] += (debit * rate, credit * rate)`

sbidoul (commented):
Interesting, I thought we'd need this for 18 only. The fix looks good to me.

Could you look at adding a test, probably in `test_aep.py` ?

mis_builder/models/aep.py
jguenat: Not sure I understand. change the tuple to an array possibly containing multiple tuples?
won't this cause issue in other parts of the code?

sbidoul: SimpleArray behaves like a tuple, but with support for member-wise arithmetic operations.

I have pushed a commit here to illustrate.

(Note I actually wanted to do another PR, but I accidentally pushed to your branch. I hope you don't mind).

sbidoul (approved):
Thanks for adding a test!

mis_builder/models/aep.py
jguenat: Very handy this SimpleArray, thank you !

sbidoul:
Removing the needs port label as this is included in #652

PR #652, opened 2024-12-02
[18.0][MIG] mis_builder: Migration to 18.0

chaule97:
- When searching for accounts by code, we only [search](https://github.com/odoo/odoo/blob/74d4503a45c9d4ca741d349c3fce673358d2cc42/addons/account/models/account_account.py#L148) for accounts by code within a single company. I will add a [TODO note](https://github.com/chaule97/mis-builder/blob/6947a817867bb3c98065315682e82862ba4fe646/mis_builder/models/aep.py#L197) to improve this query in future updates.
- In Odoo 18, [name_search](https://github.com/odoo/odoo/blob/74d4503a45c9d4ca741d349c3fce673358d2cc42/odoo/models.py#L1817) already searches by `display_name`, so I override the `_search_display_name` method in `mis.report.kpi.expression`.
- in the `reporting-engine` module,  the `report_type` field in `report_type` model has a new value [xlsx](https://github.com/OCA/reporting-engine/blob/a3b5fc95a5f82ee2d1079f30c6a0b7625a5173f2/report_xlsx/models/ir_report.py#L15). However, the [test_report](https://github.com/odoo/odoo/blob/fa375ffa4467efe99d53a7546f4dcf0553f51190/odoo/tools/test_reports.py#L65) in odoo 18  can't examine it. I have to add `with self.assertLogs("odoo.tools.test_reports", level="WARNING"):` to ignore the warning.


mis_builder/models/aep.py
```diff
@@ -192,11 +192,20 @@ def done_parsing(self):
...
-                account_ids = self._account_model.search(acc_domain_with_company).ids
+                # TODO we only search account by code with only one company
+                # we need to optimize it
+                account_ids = []
+                for company in self.companies:
+                    account_ids += (
+                        self._account_model.with_company(company)
+                        .search(acc_domain_with_company)
+                        .ids
+                    )
```
sbidoul, line 207: Why is this necessary?

mis_builder/models/aep.py
```diff
@@ -341,21 +350,21 @@ def do_queries(
             # fetch sum of debit/credit, grouped by account_id
             _logger.debug("read_group domain: %s", domain)
             try:
-                accs = aml_model.read_group(
+                accs = aml_model.with_context(
+                    allowed_company_ids=self.companies.ids
```
sbidoul, line 355: Not sure we want to use the context for this. Maybe we should and `company_id in self.companies.ids` in the domain instead.

mis_builder/models/mis_report_instance.py
```diff
@@ -896,7 +911,15 @@ def drilldown(self, arg):
         expr = arg.get("expr")
         account_id = arg.get("account_id")
         if period_id and expr and AEP.has_account_var(expr):
-            period = self.env["mis.report.instance.period"].browse(period_id)
+            period = self.env["mis.report.instance.period"].search_fetch(
```
sbidoul: I see the performance benefit of search_fetch but we should balance it's use against maintainability. For small tables such as this one, the performance benefit is not worth the reduced maintainability and readability. Can you please revert these?

mis_builder/models/mis_report_style.py
```diff
@@ -133,6 +132,10 @@ def check_positive_val(self):
     hide_always_inherit = fields.Boolean(default=True)
     hide_always = fields.Boolean(default=False)
 
+    _sql_constraints = [
+        ("style_name_event_uniq", "unique(name)", "Style name should be unique")
+    ]
```
sbidoul: This constraint seems unrelated to the migration? Better have it in a separate PR.

mis_builder/readme/CREDITS.md
```diff
@@ -0,0 +1 @@
+The migration of this module from 17.0 to 18.0 was financially supported by Camptocamp.
```
sbidoul: I'm not fan of this kind of marketing.

henrybackman (approved):

sbidoul (commented):
Hello, thanks for this migration.

Lots of nice things! I need some time to look closer at the multi company part.

My main worry is the search_fetch which I think make the code less readable for little performance benefit.

As a side note (no need to change), in general I prefer to see mechanical changes such as replacing `_` by `self.env._` in separate commits to make review a bit easier.

mis_builder/models/aep.py
chaule97: > - When searching for accounts by code, we only [search](https://github.com/odoo/odoo/blob/74d4503a45c9d4ca741d349c3fce673358d2cc42/addons/account/models/account_account.py#L148) for accounts by code within a single company

Example: We have two accounts in different companies with the same code. When we search by code, even though we add `[("company_ids", "in", self.companies.ids)]`, the result includes only one account per company when with_company is used.

mis_builder/models/aep.py
chaule97: I have tried it, but its testcase failed:
```
domain = list(domain) + domain_by_mode[mode]
domain.append(("account_id", "in", self._map_account_ids[key]))
domain.append(("company_id", "in", self.companies.ids))

if additional_move_line_filter:
    domain.extend(additional_move_line_filter)
     # fetch sum of debit/credit, grouped by account_id
     _logger.debug("read_group domain: %s", domain)
      try:
          accs = aml_model.read_group(
              domain,
              ["debit", "credit", "account_id", "company_id"],
              ["account_id", "company_id"],
              lazy=False,
           )
```

> 2024-12-10 04:32:45,052 42839 ERROR odoo odoo.addons.mis_builder.tests.test_aep: FAIL: TestAEP.test_aep_basic
> Traceback (most recent call last):
>   File "/home/chau/code/oca/oca-18.0/mis-builder/mis_builder/tests/test_aep.py", line 166, in test_aep_basic
>     self.assertEqual(self._eval("balp[][('account_id.code', '=', '400AR')]"), 100)
> AssertionError: AccountingNone != 100

chaule97:
`search _fetch` is a new feature, sometime we are not not familiar with it. but you are right when we don't need to over optimize. I think if code that can't be overridden or queries that can only be used once then it is still good for modules.

image

sbidoul:
I'm aware of the optimization that search_fetch brings, but it comes at a cost of code that is harder to read, and also the optimization benefit is easily lost when later modifying the code to read another field as it's easy to forget to add the new field to the search_fetch call. So I think we should reserve these for places where it really matters.

It is anyway unrelated to the migration. Could you remove these search_fetch changes, and possibly open another PR if you think it is important in some places, so we can discuss it separately from the migration?

Could you also move the constraint on style name to a separate PR? It is also unrelated to the migration and it will be easier for me to track the backports if it is separated.

Thanks for moving the self.env._ part to a separate commit!

I'll look closer at the multicompany parts soon.

pedrobaeza:
I think `self.env._ ` is part of the migration itself, as it's something of 18.0 version that comes with the migration (we have included in fact in the migration guide). It's not a different feature. And it can't be backported.

sbidoul:
> I think `self.env._ ` is part of the migration itself, as it's something of 18.0 version that comes with the migration (we have included in fact in the migration guide). It's not a different feature. And it can't be backported.

Yes, it is in a separate commit, that is fine. I'm not asking to have it in a separate PR.

pedrobaeza:
I don't think it should go in a separate commit as said.

chaule97:
> Could you also move the constraint on style name to a separate PR? It is also unrelated to the migration and it will be easier for me to track the backports if it is separated.

Should I create PR for 17.0 or 18.0?

sbidoul:
@chaule97 as you prefer. Even 16.0 if you like.

sbidoul:
> I don't think it should go in a separate commit as said.

My reasoning is that it's similar to the changes done by pre-commit. Mechanical changes done by odoo-module-migrator are best in a separate commit and keep the changes done by a human in a distinct commit.

mis_builder/readme/CREDITS.md
pedrobaeza: I think it's legit in this case to promote that funding, and we created that section for this goal. C2C is contributing in this case this way. It's true that we need to be aware when migrating to another version to remove it, but it's fair enough to put it.

pedrobaeza:
> My reasoning is that it's similar to the changes done by pre-commit. Mechanical changes done by odoo-module-migrator are best in a separate commit and keep the changes done by a human in a distinct commit.

Well, I do that changes by hand, not using odoo-module-migrator, but then let's normalize this into a commit `[MIG] <module>: odoo-module-migrator auto-changes`, but not specific commits for each of the changes.

pedrobaeza:
Anyway, thinking twice, at the end, odoo-module-migrator is doing the same job as you as human, so we differentiate them? What we want is to have all the changes needed for a migration in one place (or is it only me that this is what I want to check while reviewing a migration - among other things, of course -?)

sbidoul:
> odoo-module-migrator is doing the same job as you as human, so we differentiate them

By that reasaoning, why then requiring a separate commit for pre-commit changes?
When there are massive mechanical changes like that it makes the review much easier when they are isolated from the rest.

pedrobaeza:
pre-commit changes are only syntactic or tool-related ones. They can be mostly ignored. These other changes involves framework changes and other modifications that may affect with more probability with the features/functioning of the module.

mis_builder/models/aep.py
sbidoul: Thanks for the explanation.

I started investigating this. And oh my... the SQL query that is now emitted by this read_group is very complex and I have serious doubts about its performance. Need to find time to dig deeper.

ryanc-me (approved):

mis_builder/models/aep.py
sbidoul: I'm investigating this at the Brussels sprint. Using `_read_group` instead of `read_group` looks promising.

sbidoul:
@chaule97 I pushed a commit to your branch with the `_read_group` improvement. 

There is still another one I'm working on right now.

mis_builder/tests/test_mis_report_instance.py
```diff
@@ -504,21 +502,22 @@ def test_qweb(self):
...
-            report_type="xlsx",
-        )
+        with self.assertLogs("odoo.tools.test_reports", level="WARNING"):
+            test_reports.try_report(
+                self.env.cr,
+                self.env.uid,
+                "mis_builder.mis_report_instance_xlsx",
+                [self.report_instance.id],
+                report_type="xlsx",
+            )
```
sbidoul, line 512: We should test that the warning is the one we expect and nothing else.

vdewulf:
This is the test I just did, not successful unfortunately:

- created a report template with 2 lines, crd[60%] and dbt[60%]
image

- created a report for 2025 period

- the database had the following entries on this account: 

image

- MIS report shows: 
30 for the credit line (not ok, should be 0 based on the entries)
30 for the debit line (ok)
image

When clicking on the 30 figures (on the 3 lines, same effect), the following error is shown:
image

sbidoul:
Added a couple commits with the refactoring of `prorata.read_group.mixin` to use `_read_group` instead of `read_group`, and a basic test. Other tests for that are in the `mis_builder_budget` module.

sbidoul:
I removed a jquery remnant in the js code.

@vdewulf the drilldown issue should be fixed. The debit/credit calculation seems correct.

mis_builder/tests/test_mis_report_instance.py
chaule97: Do I have to rename this function name to `test_xlsx_with_warning`?

sbidoul: > Do I have to rename this function name to `test_xlsx_with_warning`?

@chaule97 not necessarily. My point is that since we capture de logs, we should check that no other warning or error is logged than the one we expect.

chaule97: I have updated it

Saran440 (approved):
Functional Test :+1:

sbidoul:
I'll get back to this soon to finalize and merge.

andreampiovesana (approved):
merge?

BhaveshHeliconia:
Functional Test 👍

BhaveshHeliconia:
@sbidoul: Would it be possible to kindly merge this PR at your earliest convenience? We’re hoping to start using it in the production environment. Appreciate your support!

Thank you! 😊

andrius-preimantas:
We are also looking forward to this module. Any plans on merging any time soon?

sbidoul:
Yes, it will be merged next week latest. In the meantime the data model should not change, so it is safe to use.

andrius-preimantas:
Thanks for the response @sbidoul , highly appreciated

rov-adhoc (approved):

AnizR (approved):

sbidoul:
Alright, this one looks good now. We'll finish `mis_builder_budget` and `mis_builder_demo` separately.

/ocabot merge nobump

sbidoul:
I realised I did lose @chaule97 's authorship on two commit during my work here.

So since the merge is still very young I exceptionally did a rebase and force push to 18.0 to restore authorship on 39efa86fb1ff22563fa7dcc68095a4979a289699 and df5bbd74777fca57a95bd6af324ed31819704aa3.

chaule97:
> I realised I did lose @chaule97 's authorship on two commit during my work here.
> 
> So since the merge is still very young I exceptionally did a rebase and force push to 18.0 to restore authorship on [39efa86](https://github.com/OCA/mis-builder/commit/39efa86fb1ff22563fa7dcc68095a4979a289699) and [df5bbd7](https://github.com/OCA/mis-builder/commit/df5bbd74777fca57a95bd6af324ed31819704aa3).

No problem, just stay it

PR #654, opened 2024-12-10
[ADD] mis_builder_budget: active field for `mis.budget.by.account`

acsonefho:
Add `active` field on `mis.budget.by.account` and related views.


sbidoul:
Oops, clicked the wrong button. Sorry, I reverted.

sbidoul:
Could you open another PR, and also do the same on `mis.budget` for completeness?

pedrobaeza:
And it's not `[ADD]`, but `[IMP]`, as this is not adding any new module.

PR #655, opened 2024-12-10
[16.0][IMP] mis_builder_budget: add active field

acsonefho:
Add `active` field on `mis.budget.by.account` and `mis.budget` model and related views.


sbidoul:
The commit message must have more content on the first line, because that is what we read when looking at the list of commit, so the first line should give a decent hint of what it is about.

sbejaoui (approved):

sbidoul:
I tested on runboat but I don't see the Archived ribbon on the form. Can you add it?

sbidoul (approved):

PR #656, opened 2024-12-10
[17.0][IMP] mis_builder: enforce unique style name

chaule97:


sbidoul (approved):

mis_builder/models/mis_report_style.py
```diff
@@ -133,6 +132,10 @@ def check_positive_val(self):
     hide_always_inherit = fields.Boolean(default=True)
     hide_always = fields.Boolean(default=False)
 
+    _sql_constraints = [
+        ("style_name_event_uniq", "unique(name)", "Style name should be unique")
```
sbidoul:
```suggestion
        ("style_name_uniq", "unique(name)", "Style name should be unique")
```

A little detail.

PR #660, opened 2024-12-11
[CI][DON'T MERGE]

pedrobaeza:
We have a weird CI error in this module. Let's check if it's a general problem due to a change upstream.


pedrobaeza:
OK, no problem here.

PR #661, opened 2024-12-12
[15.0][IMP] mis_builder: change analytic account filter to many2many

ThiagoMForgeFlow:
Changing the analytic account filter in mis report instance to Man2many adds the possibility to show values that has different analytic accounts. The aggregation works as an OR.

@ForgeFlow


sbidoul:
Hello, thanks for this contrib. This could makes sense, but I don't think we can merge this as this would be a big breaking change. Also, this part has completely changed with the v16 analytic mechanisms so this could not be forward ported either.

sbidoul (changes requested):
As mentioned, I don't think we can merge this.

PR #667, opened 2025-01-08
[16.0][FIX] mis_builder: don't show account details when there is no data

### LINKED ISSUE #665: [16.0] mis_report: Added rows of accounts with no data in the comparison periods
It seems to be a side effect of the changes added in https://github.com/OCA/mis-builder/pull/584 (it was working correctly before).

Example use case:
- Create a mis report and use the template “Pérdidas y ganancias completo (PGCE 2008)” added by `l10n_es_mis_report` from OCA/l10n-spain.
- Define the comparison mode (2023 vs 2024).
- Click on preview/print/export
- Rows of all accounts will be displayed (even if there is no data in 2023 and 2024 for that account).

Before the change of the initially indicated PR (or in v15) the rows of the accounts without data in the periods were not added.

Ping @sbidoul

sbidoul:
fixes #665

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -56,6 +56,13 @@ class Accumulator:
             custom_field: AccountingNone for custom_field in custom_field_names
         }
 
+    def has_data(self):
+        return (
+            self.debit is not AccountingNone
+            or self.credit is not AccountingNone
+            or any(v is not AccountingNone for v in self.custom_fields.values())
+        )
+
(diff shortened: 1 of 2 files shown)


sbidoul:
cc/ @victoralmau

victoralmau (approved):
Functional review OK.
Can you change the PR title to: _[16.0][FIX] mis_builder: Do not show account details when there is no data_? Thanks!

PR #674, opened 2025-03-26
[14.0][IMP] mis_builder: pivot view

Chionne27:
Thanks for your pull request. Please take a moment to review if it meets the following
guidelines.

## Description

Before coding, it is recommended to create an issue to discuss the problem or feature
you want to add. If you did not, don't worry, we can discuss on the PR too.

If there is a pre-existing issue that your PR implements or fixes, add a pointer to it
in the PR description.

If your PR is simple enough that it does not require a preliminary discussion, then make
sure to explain what it does (i.e. _why_ the change is necessary).

## Test

Don't forget to add unit tests. If your PR fixes a bug, prefer creating a separate
commit for the test so we one see that your test reproduces the bug and is fixed by the
PR.

## Target branch

MIS Builder is actively maintained for Odoo versions 9, 10, 11 and 12.

If your feature is applicable with the same implementation to all these versions, please
target branch 10.0. Maintainers will port it to 9, 11 and 12 soon after merging.

In the rare cases your feature or implementation is specific to an Odoo version, then
target the corresponding branch.

## CLA

Have you signed the OCA Contributor License Agreement? If not, please visit
https://odoo-community.org/page/cla to learn how.

## Changelog entry

This projects uses [towncrier](https://pypi.org/project/towncrier/) to generate it's
changelog. Make sure your PR includes a changelog entry in
`<addon>/readme/newsfragments/`. It must have the issue or PR number as name, and one of
`.feature`, `.bugfix`, `.doc` (for documentation improvements), `.misc` (if a ticket has
been closed, but it is not of interest to users). The changelog entry must be reasonably
short and phrased in a way that is understandable by end users.

## Documentation

Consider improving the documention in `docs/`.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -949,7 +949,7 @@ class MisReportInstance(models.Model):
                 "domain": domain,
                 "type": "ir.actions.act_window",
                 "res_model": period._get_aml_model_name(),
-                "views": [[False, "list"], [False, "form"]],
+                "views": [[False, "list"], [False, "form"], [False, "pivot"]],
                 "view_mode": "list",
                 "target": "current",
                 "context": {"active_test": False},


Chionne27:
This PR adds a pivot view to the MIS report section

sbidoul:
Thanks! Can you elaborate the use case, as it's not obvious to my why this is useful.

Chionne27:
> Thanks! Can you elaborate the use case, as it's not obvious to my why this is useful.

Hi @sbidoul. This improvement could be useful better analysis of the Mis reports

mis_builder/models/mis_report_instance.py
```diff
@@ -552,7 +552,7 @@ def _compute_pivot_date(self):
     date = fields.Date(
         string="Base date", help="Report base date " "(leave empty to use current date)"
     )
-    pivot_date = fields.Date(compute="_compute_pivot_date")
+    pivot_date = fields.Date(compute="_compute_pivot_date", store=True)
```
sbidoul: Storing the pivot date is problematic, because it depends on the context and can be `today`.

Chionne27: Thanks for the review. I removed it

mis_builder/models/mis_report_instance.py
```diff
@@ -949,7 +949,7 @@ def drilldown(self, arg):
                 "domain": domain,
                 "type": "ir.actions.act_window",
                 "res_model": period._get_aml_model_name(),
-                "views": [[False, "list"], [False, "form"]],
+                "views": [[False, "list"], [False, "form"], [False, "pivot"]],
```
SirPyTech, line 952: I believe this one is the only change you need, this allows to show the pivot view during the drill down: 
image

mis_builder/views/mis_report_instance.xml
```diff
@@ -49,6 +49,15 @@
             </tree>
         </field>
     </record>
+    <record model="ir.ui.view" id="mis_report_instance_view_pivot">
```
SirPyTech: This change and the following show a pivot view of the MIS Reports: 
image

and I think it's not what you want, could you please remove them?

SirPyTech (changes requested):
Thanks for the PR!
I've done a code review and tried it in runboat, see the comments below.

mis_builder/views/mis_report_instance.xml
Chionne27: Yes I agree, I removed the pivot view in mis report and let only in the drill down view

SirPyTech (approved):

SirPyTech:
@OCA/mis-builder-maintainers can someone re-open and have a look?
It's just a one-line change to add the pivot view during the drilldown.

sbidoul:
/ocabot merge minor

Thanks for your patience.

sbidoul:
Will you port this to 18?

Chionne27:
This implementation is not necessary in v18 because the pivot view is already set for the drilldown. https://github.com/OCA/mis-builder/blob/052622a28642ee970fd59e21c180245f8d593960/mis_builder/models/mis_report_instance.py#L947

sbidoul:
Nice. Thanks for the feedback.

Issue #159, opened 2019-02-11
Footnotes / Annotations


alyngsoe:
Would it be possible in the report view to add a column for adding comments (in the report preview form), and also showing the column on the pdf report and xls export ?
Why: When the accountant is preparing reports for e.g. board meetings, it would be valuable if he/she could add short comments to specific report line. This could be "High costs from new machine impl." or "decreased sale on asia market" etc...

sbidoul:
That would be a valuable feature.

I see it rather as footnotes than a comment column, what do you think?

From the top of my head, this requires:
* a new model (`mis.report.instance.footnote`) that associates the footnote text to a report cell, ie the couple `(mis.report.subkpi, mis.report.instance.period)`.
* an extension to the preview widget to allow entering the footnotes
* updating the XLS and PDF exports to display the footnotes

alyngsoe:
@sbidoul , 
I actually discussed this with a current customer, and here we clearly are in the need of comments on report line level. Although footnotes would also be (another) nice feature ;-)

I cannot contribute directly with development ressources, but would be happy to help in anyway by testing, feedback etc..

sbidoul:
@alyngsoe can you elaborate on which use case line-level comments resolve that cannot be resolved with footnotes?

alyngsoe:
@sbidoul 
Use case 1: Tax control report.
We have a MIS tax report to show columns with "theoretical" and "actual" tax for each finansial account. When the company has an external financial audit, the internal accountant will look through this report and analyse differences. To explain the differences to the external accountant, he will need to put notes on all the report lines with differences.

alyngsoe:
@sbidoul If your suggestion of a footnote model could give us the option to display a footnote number on the report lines ( as you usually do in regular "footnote" usage), then the footnote solution would be sufficient for the feature request.

sbidoul:
@alyngsoe Yes that was my idea indeed.

sbidoul:
A simplified design could start with linking footnotes to the couple `(mis.report.kpi, mis.report.instance)`, ie report rows instead of cells. Attaching footnotes to cells could be optional and/or added later.

A further simplification is to let user enter footnootes in the `mis.report.instance` form instead of using a dedicated javascript widget. This is a bit less comfortable but much easier to implement, and the result (preview, pdf, excel) is the same.

alyngsoe:
@sbidoul your simplificated solutions sounds good to me :-)

lili94:
[...]

PR #676, opened 2025-04-03
[16.0][IMP] mis_builder: introduce annotation in reports

AnizR:
This PR introduces 'annotation' in mis_builder reports.
An annotation is a comment or a remark that a user can write on a cell.

It tries to solve issue: #159 (opened 6 yeas ago!)

# Description
The goal is to write a remark on a cell (independently of the pivot date).
These annotations are visible directly within the view but also at export in .pdf and in .xlsx.

**Edit:** 2 security groups have been introduced to allow edit and read of annotaitons
**Edit 2:** It is not possible to add a note on a line that comes from the option "details by account" 

### UI in mis_builder_widget
On cells, we have now a drop down menu to drill down or annotate:
image
It opens a window to write an annotation:
image
image

### Print .pdf

The annotation is also present in the .pdf:
image

### Export .xls

In the .xls, the annotation are added as 'comment' on cells:
image


lmarion-source (approved):

hitrosol:
Long wait feature finally come. 

I tried on runboat using Chrome, Edge and Safari. The drop down menu is behind the other row.

Screen Shot 2025-04-03 at 11 19 37 PM


It looks okay on Firefox, but no sequence number is added in the footnotes. If I refresh the browser the notes is gone.
Screen Shot 2025-04-03 at 11 21 59 PM

Thanks

sbidoul:
> Long wait feature finally come.

It was just waiting for someone to fund it.

AnizR:
> Long wait feature finally come.
> 
> I tried on runboat using Chrome, Edge and Safari. The drop down menu is behind the other row.
> 
> Screen Shot 2025-04-03 at 11 19 37 PM
> 
> It looks okay on Firefox, but no sequence number is added in the footnotes. If I refresh the browser the notes is gone. Screen Shot 2025-04-03 at 11 21 59 PM
> 
> Thanks

Thanks for your feedback.
I am using firefox but I'll try to figure out why it doesn't work with other browsers.
Regarding the annotations, I opened your instance and indeed, it doesn't work with your column '3 M Budget' but it works with others. I'll fix it:
image

AnizR:
> Long wait feature finally come.
> 
> I tried on runboat using Chrome, Edge and Safari. The drop down menu is behind the other row.
> 
> Screen Shot 2025-04-03 at 11 19 37 PM
> 
> It looks okay on Firefox, but no sequence number is added in the footnotes. If I refresh the browser the notes is gone. Screen Shot 2025-04-03 at 11 21 59 PM
> 
> Thanks

It should be fixed now.
Thanks for your feedback!

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,56 @@
...
+    kpi_id = fields.Many2one(
+        comodel_name="mis.report.kpi",
+    )
+    subkpi_id = fields.Many2one(
+        comodel_name="mis.report.subkpi",
+    )
+    note = fields.Char()
+
+    @api.model
+    def edit_annotation(self, period_id, kpi_id, subkpi_id, note):
```
sbidoul:
```suggestion
    def set_annotation(self, period_id, kpi_id, subkpi_id, note):
```

mis_builder/models/mis_report_instance.py
```diff
@@ -801,6 +801,17 @@ def _add_column_move_lines(self, aep, kpi_matrix, period, label, description):
...
+        annotations = self.env["mis.report.instance.annotation"].search(
+            [
+                ("period_id", "=", period.id),
+            ]
+        )
+
+        note_by_subkpi_id = {
+            (note.kpi_id.id, note.subkpi_id.id): note.note for note in annotations
+        }
+        locals_dict = {"notes": note_by_subkpi_id}
```
sbidoul: I need to look closer, but locals_dict is used as the evaluation context for expressions. I don't think it is useful to add notes in that context

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,56 @@
...
+
+    period_id = fields.Many2one(
+        comodel_name="mis.report.instance.period",
+    )
+    kpi_id = fields.Many2one(
+        comodel_name="mis.report.kpi",
+    )
+    subkpi_id = fields.Many2one(
+        comodel_name="mis.report.subkpi",
+    )
```
sbidoul: Add a unique constraint on these 3 fields?

mis_builder/security/ir.model.access.csv
```diff
@@ -20,3 +20,4 @@ access_mis_report_subreport,access_mis_report_subreport,model_mis_report_subrepo
 manage_mis_report_style,access_mis_report_style,model_mis_report_style,account.group_account_manager,1,1,1,1
 access_mis_report_style,access_mis_report_style,model_mis_report_style,base.group_user,1,0,0,0
 access_add_to_dashboard_wizard,access_add_to_dashboard_wizard,model_add_mis_report_instance_dashboard_wizard,base.group_user,1,1,1,0
+access_mis_report_annotation, access_mis_report_annotation,model_mis_report_instance_annotation,base.group_user,1,1,1,1
```
sbidoul: That is too broad? Annotation can possibly contain sensitive data so I would not have that readable by all employees by default.

sbidoul (commented):
Thanks @AnizR. Here is a first round of comment after a quick look.

mis_builder/models/mis_report_instance.py
AnizR: Indeed, I refactored a bit to avoid using the locals_dict.
Instead, I directly added the notes when adding a column which enables adding annotation on budget and comparison columns.

mis_builder/security/ir.model.access.csv
AnizR: To be as flexible as possible, I introduced 2 new groups to edit and read annotations

SAnnabelle (changes requested):
It works well! 

Would it be possible to make the annotations company-dependent? This would allow each company to have its own comments if the report is shared between several companies.

AnizR:
> It works well!
> 
> Would it be possible to make the annotations company-dependent? This would allow each company to have its own comments if the report is shared between several companies.

Thanks for your feedback!

Annotations are now linked to a company.
This company is the 'main' company of the user when creating the annotation.

sbidoul:
> Annotations are now linked to a company.
> This company is the 'main' company of the user when creating the annotation.

Is that not potentially surprising for the user?

For instance, what if it is a multi-company consolidation report?

sbidoul:
I wonder to what extent the annotation should be linked to (some of) the filter values of the report (effective companies, pivot date, ...).

AnizR:
> > Annotations are now linked to a company.
> > This company is the 'main' company of the user when creating the annotation.
> 
> Is that not potentially surprising for the user?
> 
> For instance, what if it is a multi-company consolidation report?

I figured that the 'final user' would always use the same 'main company' when watching his report. Therefore, in my case, it makes sense to attach annotations to the 'main company'. What would you prefer? Storing every companies selected when watching the report?


> I wonder to what extent the annotation should be linked to (some of) the filter values of the report (effective companies, pivot date, ...).

I thought about adding the pivot date but realized that in odoo enterprise, they also have an 'annotation' feature but they don't store the pivot date. The annotation is linked to the 'cell'.

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,71 @@
...
```
sbidoul: `ondelete="cascade"` on period_id, kpi_id, subkpi_id ?

Some `required=True`?

Also add a composite index on `period_id, kpi_id, subkpi_id` to anticipate when many old notes accumulate?

AnizR: done :+1:

AnizR:
> I wonder to what extent the annotation should be linked to (some of) the filter values of the report (effective companies, pivot date, ...).

I changed a bit my code:

- I removed the field 'company_id' on annotations
- I introduced a 'annotation_context' that represents the context in which an annotation has been added and in which we should display an annotation.

For now, I only use the 'query_company_ids' from the mis report instance.

So, let's say that we have a report that can be used on company A and B. I am on company A and I add a note. This note will be visible only if I display the report from company A (without company B). If I add a note while having company A and B selected, this note will be displayed on the report only if I have company A and B selected.

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,106 @@
...
+        if annotation:
+            annotation.note = note
+        else:
+            self.env["mis.report.instance.annotation"].create(
+                {
+                    "period_id": period_id,
+                    "kpi_id": kpi_id,
+                    "subkpi_id": subkpi_id,
+                    "note": note,
+                    "annotation_context": self._get_annotation_context(instance_id),
```
sbidoul:
```suggestion
                    "annotation_context": json.dumps(instance_id._get_annotation_context()),
```

It's simpler.

But is the `json.dumps` necessary? I'd think we can provide the dict to the json field.

mis_builder/models/mis_report_instance.py
```diff
@@ -940,3 +963,15 @@ def _get_drilldown_action_name(self, arg):
             return f"{kpi.description} - {account.display_name} - {period.display_name}"
         else:
             return f"{kpi.description} - {period.display_name}"
+
+    def _get_annotation_context(self):
+        """
+        Return context of annotation linked to this instance
+        """
```
```suggestion
        """Return the context used to filter annotation linked to this instance."""
```

Also, nit: https://peps.python.org/pep-0257/

mis_builder/security/res_groups.xml
```diff
@@ -0,0 +1,13 @@
...
+        <field name="name">Mis report: read annotation</field>
+    </record>
+    <record model="res.groups" id="group_edit_annotation">
+        <field name="name">Mis report: edit annotation</field>
+        <field
+            name="implied_ids"
+            eval="[Command.link(ref('mis_builder.group_read_annotation'))]"
+        />
+    </record>
+</odoo>
```
sbidoul, line 17: Automatically assign these new groups to the admin user? I think there is a pattern for that throughout Odoo.

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -85,21 +85,49 @@
...
+                                        >
+                                            <div
+                                                href="javascript:void(0)"
+                                                class="dropdown-item js_tag"
+                                                t-on-click="drilldown"
+                                                t-att-data-drilldown="JSON.stringify(cell.drilldown_arg)"
+                                                t-if="cell.drilldown_arg"
+                                                role="menuitem"
+                                            >
+                                                Explore
```
sbidoul: Drill down instead of Explore ? not sure.

But can we keep the immediate drill down when we click on the value, and have the menu appear only when we click on the little down arrow ?

mis_builder/security/res_groups.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record model="res.groups" id="group_read_annotation">
+        <field name="name">Mis report: read annotation</field>
```
sbidoul:
```suggestion
        <field name="name">MIS Report: view annotations</field>
```

mis_builder/security/res_groups.xml
```diff
@@ -0,0 +1,13 @@
...
+    </record>
+    <record model="res.groups" id="group_edit_annotation">
+        <field name="name">Mis report: edit annotation</field>
```
sbidoul:
```suggestion
        <field name="name">MIS report: add annotations</field>
```

sbidoul (commented):
Here are a few remarks about ergonomy:

- Now the cells where there is no drill down are underlined on hover. It is important that this underline appears only as visual cue that a drilldown is possible, as before.
- After adding an annotation I had to refresh to see it. Is that intended?
- Removing an annotation leaves a weird display, with the little call out still present until I refresh

image

I think it's important that the annotation changes appear correctly immediately, and without requiring a full recompute.

Tested with chrome on the demo report on runboat.

mis_builder/models/mis_report_instance_annotation.py
AnizR: Indeed the `json.dumps` can be safely removed

mis_builder/static/src/components/mis_report_widget.xml
AnizR: Ok, I'll remove the 'Explore' in the menu and set back the immediate drill down when clicking on cell's value

AnizR:
> Here are a few remarks about ergonomy:
> 
>     * Now the cells where there is no drill down are underlined on hover. It is important that this underline appears only as visual cue that a drilldown is possible, as before.
> 
>     * After adding an annotation I had to refresh to see it. Is that intended?
> 
>     * Removing an annotation leaves a weird display, with the little call out still present until I refresh
> 
> 
> image
> 
> I think it's important that the annotation changes appear correctly immediately, and without requiring a full recompute.
> 
> Tested with chrome on the demo report on runboat.

Thanks for your feedback.
I have done some changes, those ergonomic issues are fixed.
Now, the loading of notes can be triggered independently from the computation of the matrix of a report!

mis_builder/models/kpimatrix.py
```diff
@@ -524,16 +523,47 @@ def as_dict(self):
...
                     col_data = {
+                        "cell_id": cell.cell_id,
                         "val": val,
                         "val_r": cell.val_rendered,
                         "val_c": cell.val_comment,
                         "style": self._style_model.to_css_style(
                             cell.style_props, no_indent=True
                         ),
+                        # notes can not be added on 'details by account' lines
+                        "can_be_annotated": not bool(cell.row.account_id),
```
sbidoul:
```suggestion
                        "can_be_annotated": not cell.row.account_id,
```

mis_builder/models/kpimatrix.py
```diff
@@ -524,16 +523,47 @@ def as_dict(self):
...
                         col_data["drilldown_arg"] = cell.drilldown_arg
                     row_data["cells"].append(col_data)
             body.append(row_data)
 
         return {"header": header, "body": body}
+
+    # logic to convert semantic coordinates (period, kpi, subkpi)
+    # to visual coordinates (cell id)
+    @classmethod
+    def _make_row_id(cls, kpi_id, account_id) -> str:
```
sbidoul:
```suggestion
    def _make_row_id(cls, kpi_id: int, account_id: int | None) -> str:
```

mis_builder/models/kpimatrix.py
```diff
@@ -524,16 +523,47 @@ def as_dict(self):
...
                     }
                     if cell.drilldown_arg:
                         col_data["drilldown_arg"] = cell.drilldown_arg
                     row_data["cells"].append(col_data)
             body.append(row_data)
 
         return {"header": header, "body": body}
+
+    # logic to convert semantic coordinates (period, kpi, subkpi)
+    # to visual coordinates (cell id)
```
sbidoul:
```suggestion
    # Logic to convert semantic coordinates (period, kpi, subkpi)
    # to visual coordinates (cell id) and back. The rendering logic must no know
    # about semantic concepts such as periods and kpis. Having these well identified
    # methods allow us to easily spot where the conversion between the rendering and
    # semantic domain occur.

```

mis_builder/models/kpimatrix.py
```diff
@@ -524,16 +523,47 @@ def as_dict(self):
...
+    @classmethod
+    def _make_row_id(cls, kpi_id, account_id) -> str:
+        return f"{kpi_id}:{account_id or ''}"
+
+    @classmethod
+    def _make_cell_id(cls, kpi_id, account_id, period_id, subkpi_id) -> str:
+        return f"{kpi_id}#{account_id or ''}#{period_id}#{subkpi_id or ''}"
+
+    @classmethod
+    def _pack_cell_id(cls, cell):
```
sbidoul:
```suggestion
    def _pack_cell_id(cls, cell: KpiMatrixCell) -> str:
```

mis_builder/models/kpimatrix.py
```diff
@@ -524,16 +523,47 @@ def as_dict(self):
...
+        return cls._make_cell_id(
+            cell.row.kpi.id,
+            cell.row.account_id,
+            cell.subcol.col.key,
+            cell.subcol.subkpi and cell.subcol.subkpi.id,
+        )
+
+    @classmethod
+    def _unpack_cell_id(cls, cell_id: str):
```
sbidoul:
```suggestion
    def _unpack_cell_id(cls, cell_id: str) -> Tuple[int, int | None, int, int | None:
         """Convert cell id to kpi_id, account_id, period_id, subkpi_id."""
```

mis_builder/models/mis_report_instance.py
```diff
@@ -877,7 +878,35 @@ def _compute_matrix(self):
...
+
+        return {
+            KpiMatrix._make_cell_id(
+                annotation.kpi_id.id,
+                False,
+                annotation.period_id.id,
+                annotation.subkpi_id and annotation.subkpi_id.id,
+            ): {"note_text": annotation.note, "note_sequence": sequence}
+            for sequence, annotation in enumerate(annotations, 1)
+        }
```
sbidoul, line 924: Computing the sequence without sorting here may lead to surprising numbering. I see two possibilities
1. assign sequence numbers in the rendering phase (i.e. in de widget, and pdf - in xls we don't need it)
2. or sort the annotations here, probably by (period.sequence, subkpi.sequence, kpi.sequence), altough this sorting algo would gain quite a bit of knowledge of the rendering algorithm

Granted, (2) my be hard to do in a qweb template.

mis_builder/models/mis_report_instance.py
```diff
@@ -877,7 +878,35 @@ def _compute_matrix(self):
     def compute(self):
         self.ensure_one()
         kpi_matrix = self._compute_matrix()
-        return kpi_matrix.as_dict()
+        ret = kpi_matrix.as_dict()
+
+        ret["notes"] = self.get_notes()
+        return ret
+
+    def get_notes(self) -> dict:
```
```suggestion
    def get_notes_by_cell_id(self) -> dict:
        """Return annotations by cell id."""
```

mis_builder/models/mis_report_instance.py
```diff
@@ -940,3 +969,13 @@ def _get_drilldown_action_name(self, arg):
             return f"{kpi.description} - {account.display_name} - {period.display_name}"
         else:
             return f"{kpi.description} - {period.display_name}"
+
+    def _get_annotation_context(self):
+        """Return the context used to filter annotation linked to this instance."""
+        self.ensure_one()
+        annotation_context = {}
+        if companies := self.query_company_ids.ids:
```
```suggestion
        if query_company_ids := self.query_company_ids.ids:
```

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,105 @@
...
+        comodel_name="mis.report.kpi",
+        ondelete="cascade",
+        required=True,
+    )
+    subkpi_id = fields.Many2one(
+        comodel_name="mis.report.subkpi",
+        ondelete="cascade",
+    )
+    note = fields.Char()
+
```
sbidoul:
```suggestion
```

mis_builder/models/kpimatrix.py
```diff
@@ -144,6 +141,8 @@ def __init__(
         self.drilldown_arg = drilldown_arg
         self.val_type = val_type
```
sbidoul:
```suggestion
```

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,105 @@
...
+        help="""
+        Context used when adding annotation
+        """
+    )
+
+    _sql_constraints = [
+        (
+            "annotation_unique",
+            "unique(period_id, kpi_id,subkpi_id)",
+            "Only one annotation per cell(defined using period,kpi,sub kpi)",
```
sbidoul:
```suggestion
            "Only one annotation per cell is allowed (defined using period, kpi, sub kpi)",
```

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,105 @@
...
+        self.env.cr.execute(
+            """
+                CREATE INDEX IF NOT EXISTS
+                mis_report_instance_annotation_period_id_kpi_id_subkpi_id_idx
+                ON mis_report_instance_annotation(period_id,kpi_id,subkpi_id);
+            """
+        )
+
+    @api.model
+    def get_first_matching_annotation(self, cell_id, instance_id):
```
sbidoul:
```suggestion
    def _get_first_matching_annotation(self, cell_id, instance_id):
```

mis_builder/models/mis_report_instance.py
```diff
@@ -877,7 +878,35 @@ def _compute_matrix(self):
...
+            lambda rec: rec.annotation_context == annotation_context
+        )
+
+        return {
+            KpiMatrix._make_cell_id(
+                annotation.kpi_id.id,
+                False,
+                annotation.period_id.id,
+                annotation.subkpi_id and annotation.subkpi_id.id,
+            ): {"note_text": annotation.note, "note_sequence": sequence}
```
```suggestion
            ): {"text": annotation.note, "sequence": sequence}
```

In the contexts where we use that we know we are talking about notes, so this would like make things lighter.

mis_builder/static/src/components/mis_report_widget.css
```diff
@@ -65,3 +65,14 @@
...
     justify-content: flex-end;
 }
+
+.dropdown-mis-builder {
+    overflow: visible !important;
+}
+
+.footnote-mis-builder {
+    font-size: 80%;
+    color: #fd0000;
```
sbidoul: I think there are color variables we can and should use.

mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -138,6 +142,21 @@ export class MisReportWidget extends Component {
             [this._instanceId()],
             {context: this.context}
         );
+        this.state.can_edit_annotation = await this.user.hasGroup(
+            "mis_builder.group_edit_annotation"
+        );
+        this.state.can_read_annotation = await this.user.hasGroup(
+            "mis_builder.group_read_annotation"
+        );
```
sbidoul: Let's add fields on mis.report.instance so the logic to decide if annotations can be read or written is easier to override.

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -85,21 +86,52 @@
...
+                                                />
+                                            </span>
+                                        </div>
+
+                                        <div
+                                            class="dropdown-menu o_filter_menu"
+                                            role="menu"
+                                        >
+                                            <a
+                                                t-if="state.can_edit_annotation and cell.can_be_annotated"
```
sbidoul, line 119: Always show the Annotate menu, but gray it out / disable it if annotations are not supported.

Actually, if `state.can_edit_annotation` is False, we can hide the dropdown triangle completely. And if `cell.can_be_annotated` is False, show it with a diabled Annotate menu entry.

sbidoul (approved):
The code is much easier to read, thanks!

mis_builder/models/mis_report_instance.py
AnizR: I prefer the second option with a small tweak, I'll sort the annotation by (kpi.sequence,period.sequence,subkpi.sequence) to follow the 'reading order' (left to right and up to bottom)

AnizR:
I have done some changes and improvements that were asked.

I've done multiple commits in order to make the review easier. Once everything will be "approved", I'll squash all my commits.

mis_builder/models/mis_report_instance.py
```diff
@@ -877,7 +884,44 @@ def _compute_matrix(self):
...
         kpi_matrix = self._compute_matrix()
-        return kpi_matrix.as_dict()
+        ret = kpi_matrix.as_dict()
+
+        ret["notes"] = self.get_notes_by_cell_id()
+        return ret
+
+    def get_notes_by_cell_id(self) -> dict:
+        self.ensure_one()
+        if not self.env.user.has_group("mis_builder.group_read_annotation"):
```
```suggestion
        if not self.user_can_read_annotation:
```

mis_builder/models/mis_report_instance_annotation.py
```diff
@@ -0,0 +1,104 @@
...
+                    .browse(instance_id)
+                    ._get_annotation_context(),
+                }
+            )
+
+    @api.model
+    def remove_annotation(self, cell_id, instance_id):
+        annotation = self._get_first_matching_annotation(cell_id, instance_id)
+        if annotation:
+            annotation.unlink()
```
sbidoul, line 113: Should we test `instance.user_can_edit_annotation` here?

mis_builder/static/src/components/mis_report_widget.css
```diff
@@ -65,3 +65,18 @@
...
+}
+
+.footnote-mis-builder {
+    font-size: 80%;
+    color: red;
+    position: relative;
+    bottom: 1ex;
+}
+
+.annotation_disabled {
```
sbidoul: not used?

mis_builder/tests/test_mis_report_instance_annotation.py
```diff
@@ -0,0 +1,165 @@
...
+                    ),
+                    (
+                        0,
+                        0,
+                        dict(
+                            name="subkpi2_report2",
+                            description="subkpi 2, report 2",
+                            sequence=2,
+                        ),
+                    ),
```
sbidoul, line 28: Use `Command.create`?

mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -140,6 +148,15 @@ export class MisReportWidget extends Component {
         );
     }
 
+    async refresh_annotation() {
+        this.state.mis_report_data.notes = await this.orm.call(
+            "mis.report.instance",
+            "get_notes",
```
sbidoul:
```suggestion
            "get_notes_by_cell_id",
```

mis_builder/report/mis_report_instance_qweb.xml
```diff
@@ -97,12 +98,30 @@
                                                 <t
                                                     t-out="cell and cell.val_rendered or ''"
                                                 />
+                                                <span class="footnote-mis-builder">
+                                                    <t
+                                                        t-out="notes.get(cell.cell_id,{}).get('sequence','')"
```
sbidoul, line 106: I got a `AttributeError: 'NoneType' object has no attribute 'cell_id'` when generating the pdf with the default demo report on runboat, probably here.

sbidoul (commented):
A few more details.

If disabled menu items don't work I really prefer to have the little triangle with an empty menu so figures stay aligned. Can yo also put a little space between the cell content and the arrow?

image


Otherwise I think its almost done. Thanks a lot!

mis_builder/report/mis_report_instance_qweb.xml
AnizR: In deed, I'll check that, thanks!

AnizR:
> A few more details.
> 
> If disabled menu items don't work I really prefer to have the little triangle with an empty menu so figures stay aligned. Can yo also put a little space between the cell content and the arrow?
> 
> image
> 
> Otherwise I think its almost done. Thanks a lot!

Thanks again for your review.
Here is how it looks now with some padding and the drop-down button greyed out:
image

sbidoul:
There is still an alignment problem: 
image

sbidoul:
Also, can we tweak the pdf a little bit?

- In cells, make a reserved space for the footnoote callout so figures remain aligned? Add this reserved space in all cells only if there is at least one note?
- Remove the bullets in the bottom footnotes list (in books or documents we usually see the footnote number, it's enough and nicer).
- Add a some padding between the bottom footnotes list and the table.

image

sbidoul:
> In cells, make a reserved space for the footnoote callout so figures remain aligned? Add this reserved space in all cells only if there is at least one note?

In the widget too?

mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -170,6 +187,69 @@ export class MisReportWidget extends Component {
...
+                qweb.render("mis_builder.annotation_dialog", {
+                    text: note_text,
+                })
+            ),
+            buttons: buttons,
+        }).open();
+    }
+
+    async remove_annotation(event) {
+        const cell_id = $(event.target).data("cell-id");
```
sbidoul: Do we really need these $ ? In v18 there is no jquery anymore, so let's not add more dependency on it, that will make migration harder.

mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -170,6 +187,69 @@ export class MisReportWidget extends Component {
...
+        new Dialog(this, {
+            title: "Annotate",
+            size: "medium",
+            $content: $(
+                qweb.render("mis_builder.annotation_dialog", {
+                    text: note_text,
+                })
+            ),
+            buttons: buttons,
+        }).open();
```
sbidoul, line 245: A little detail if not hard to do: put the focus on the content field when opening the dialog box.

AnizR: I tried but didn't manage do it easily

AnizR:
> There is still an alignment problem: image

Finally :+1: 
image

sbidoul (approved):

SAnnabelle:
1. The structure of the note with line breaks is not retained in footnotes.

image

image

2. Is it possible that when we click on the footnote number, we are redirected to the footnote in the report?

sbidoul:
Rich text annotations might be a can of worms :) And it's not really a footnote anymore (if we compare to books or documents). Is that an important use case?

sbidoul:
> Is it possible that when we click on the footnote number, we are redirected to the footnote in the report?

Which makes me think. In the widget is it not better to display footnotes on hover? In large reports where the footnotes be far below the view it would make them easier to view.

AnizR:
> > Is it possible that when we click on the footnote number, we are redirected to the footnote in the report?
> 
> Which makes me think. In the widget is it not better to display footnotes on hover? In large reports where the footnotes be far below the view it would make them easier to view.

I'll add it on hover but it is possible to directly see the annotation by opening the wizard to edit the annotation

sbidoul:
> It is possible to directly see the annotation by opening the wizard to edit the annotation

Except if you are a readonly user?

AnizR:
> Rich text annotations might be a can of worms :) And it's not really a footnote anymore (if we compare to books or documents). Is that an important use case?

What I can do is allowing multi line note and display as:
image



> > It is possible to directly see the annotation by opening the wizard to edit the annotation
> 
> Except if you are a readonly user?

In deed!

AnizR:
I've added:
1. Displaying note when hovering the sequence of annotation
2. Clicking on annotations triggers a scroll down to the section of note
3. In the footnotes, when clicking on the number of a note, it will scroll up to the corresponding note
4. Multi line notes are displayed in multi line instead of forcing it into one line

mis_builder/static/src/components/mis_report_widget.css
```diff
@@ -80,8 +80,14 @@
     padding-right: 1px;
 }
 
-.footnote_mis_builder_list {
+.footnote_mis_builder_table {
```
sbidoul: I had not noticed this, but let's prefix all css classes with `.oe_mis_builder` ?

SAnnabelle (approved):
Functionally, it's very good. 

- Access rights work
- The multi-company context is taken into account
- Notes are also included in the pdf file and excel export
- Several things have been put in place to make it easier for users to create and view notes

AnizR (commented):
The 'drill down' seems to be broken, I'll fix it.

AnizR:
> The 'drill down' seems to be broken, I'll fix it.

Fixed!

sbidoul:
If there are no additional comments I'll merge next week.

hbrunn:
note this breaks with python 3.9

sbidoul:
Ah, yes, we don't test nor officially support Python 3.9 in OCA 16.0.
What's breaking exactly?

hbrunn:
the type hints, `sometype | someothertype` was introduced in 3.10

sbidoul:
> the type hints, `sometype | someothertype` was introduced in 3.10

Right. We'll fix that.

hbrunn:
I'll be happy to PR removing the `| None` parts, the rest happily works with py3.9. just wanted to be sure you actually care about 3.9

PR #679, opened 2025-05-14
[17.0][FIX] mis_builder: Wrong expression assignment in evaluator

juancarlosonate-tecnativa:
If self.aep is not present, the original expression should not be evaluated, as it might be invalid or cause an error.

TT56278 @tecnativa @pedrobaeza
image


juancarlosonate-tecnativa:
this doesn't seem to be the issue entirely (still investigating)

PR #680, opened 2025-05-16
[17.0][WIP] mis_builder_budget: Define the appropriate value if not budgetable

victoralmau:
Define the appropriate value if not budgetable

Example use case:
- Uncheck in the KPI of a template (Other for example) the Budgetable option.
- Preview the MIS report.
- The data in this column must be empty.
- The data in the sum column (if any) must be appropriate.

**Before**
antes

**After**
despues

@Tecnativa TT56278

### DIFF
--- a/mis_builder_budget/models/mis_report_instance.py
+++ b/mis_builder_budget/models/mis_report_instance.py
@@ -11,7 +11,7 @@ from .mis_report_instance_period import SRC_MIS_BUDGET, SRC_MIS_BUDGET_BY_ACCOUN
 
 
 class MisBudgetAwareExpressionEvaluator(ExpressionEvaluator):
-    def __init__(self, date_from, date_to, kpi_data, additional_move_line_filter):
+    def __init__(self, aep, date_from, date_to, kpi_data, additional_move_line_filter):
         super().__init__(
             aep=None,
             date_from=date_from,
@@ -20,6 +20,7 @@ class MisBudgetAwareExpressionEvaluator(ExpressionEvaluator):
             aml_model=None,
         )
         self.kpi_data = kpi_data
+        self._origin_aep = aep
 
     @api.model
     def _get_kpi_for_expressions(self, expressions):
@@ -44,6 +45,10 @@ class MisBudgetAwareExpressionEvaluator(ExpressionEvaluator):
                 vals.append(self.kpi_data.get(expression, AccountingNone))
                 drilldown_args.append({"expr_id": expression.id})
             return vals, drilldown_args, False
+        elif kpi and not kpi.budgetable:
+            self.aep = self._origin_aep
+            self._aep_queries_done = False
+            self.aep_do_queries()
         return super().eval_expressions(expressions, locals_dict)
 
 
@@ -63,6 +68,7 @@ class MisReportInstance(models.Model):
         )
 
         expression_evaluator = MisBudgetAwareExpressionEvaluator(
+            aep,
             period.date_from,
             period.date_to,
             kpi_data,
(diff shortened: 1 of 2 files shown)


sbidoul:
Note this behaviour is by design. Non budgetable KPIs are for KPIs that are computed from other KPIs (sums etc).

pedrobaeza:
Isn't a bit weird to have that ERR expressions if you haven't marked everything as budgetable? Can that check be computable somehow for avoiding this error? Or to silent that errors?

sbidoul:
The errors show that such KPI should be budgetable as they are not computed from other KPIs.

What's the problem with marking them budgetable?

juancarlosonate-tecnativa:
It's not that there's a problem, it's just illogical. Why do i have to mark all the KPIs as budgetable if we only want to budget one of them? Logically, for the non-budgetable KPIs, the column you're assigning as the budget source should either show no data or a zero. Otherwise, for example, column totals could be incorrect.

sbidoul:
> Logically, for the non-budgetable KPIs, the column you're assigning as the budget source should either show no data or a zero.

The current logic is that a non budgetable kpi is computed from other budgetable kpis. To give an example, if you have two budgetable KPIs `revenue` and `cost` and a third one that is not budgetable named `profit = revenue - cost`, we expect the budgeted profit to be computed. So showing an empty value for non budgetable kpis won't do what we expect.

juancarlosonate-tecnativa:
> The current logic is that a non budgetable kpi is computed from other budgetable kpis.

I'm sorry, but i don't understand this it doesn't make sense to me, especially if the logic of the mis_builder itself is that all kpis are non-budgetable.

marcfargas:
Hi all,

Barging in here (we are users of this module).

If I am understanding correctly, what @sbidoul is saying is that non-budgetable does not exactly mean (1) "the user cannot budget this" but more (2) "this cannot be budgeted because it is a computed value" (ie, has an expression). 

That would mean that those KPIs would in fact be budgetable, it just happens that the budget is computed not manually entered, but a budget it is. One may expect non-budgetable to mean that no matter what, that KPI will be blank on the reports.

Is that so? In such a case, there would be no way to inform what can be or cannot be budgeted by the user (meaning 1) while meaning 2 is redundant (if there's an expression/computation it is implicit the non-budgetable status).  

It is my understanding that at this moment I would have to mark as budgetable: what the user can budget (1) and any computed dependencies of that KPI (2).

sbidoul:
> Is that so? In such a case, there would be no way to inform what can be or cannot be budgeted by the user (meaning 1) while meaning 2 is redundant (if there's an expression/computation it is implicit the non-budgetable status).

More or less. There is indeed no way to inform that a KPI cannot/should not be budgeted by the user *if that KPI contains an accounting expression such as `bal[...]`*.

So currently, all KPIs that have accounting expressions or expressions on Queries *must* be marked budgetable (or else an error will be displayed), and KPIs that are expressions depending on other KPIs *should not* be marked budgetable so they can be computed from other budgeted values and avoid redundant entries in the budget.

juancarlosonate-tecnativa:
> More or less. There is indeed no way to inform that a KPI cannot/should not be budgeted by the user _if that KPI contains an accounting expression such as `bal[...]`_.

Well, i still stand by my opinion that this behavior is counterintuitive, from the moment all KPIs are considered non-budgetable by default module, the use of the budgetable boolean becomes confusing and counterintuitive for most users.

In my opinion the expected behavior should be to display 0 or no data in the budget source column in case the KPI is not budgetable.

sbidoul:
> this behavior is counterintuitive

I can agree with that.

But unless a more complete solution is proposed, I would suggest improving the documentation.

Displaying 0 or no data is not acceptable for the reasons I explained above.

pedrobaeza:
What about detecting expressions with one of the reserved words (bal, balp,etc), and discard that expressions automatically in the budget column?

pedrobaeza:
@sbidoul please check my previous comment https://github.com/OCA/mis-builder/pull/680#issuecomment-2915812449

PR #684, opened 2025-05-23
[16.0][IMP] mis_builder: remove py3.10-style type hints

hbrunn:
[DO NOT FORWARD PORT]

cf https://github.com/OCA/mis-builder/pull/676#issuecomment-2904616126

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -1,6 +1,8 @@
 # Copyright 2014 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from __future__ import annotations
+
 import logging
 from collections import OrderedDict, defaultdict
 


mis_builder/models/kpimatrix.py
```diff
@@ -546,12 +546,12 @@ def as_dict(self):
     # semantic domain occur.
 
     @classmethod
-    def _make_row_id(cls, kpi_id: int, account_id: int | None) -> str:
+    def _make_row_id(cls, kpi_id: int, account_id: int) -> str:
```
sbidoul: Could you do like this?

```suggestion
    def _make_row_id(cls, kpi_id: int, account_id: Optional[int]) -> str:
```

with `from typing import Optional` at the top.

hbrunn: ruff doesn't allow that: https://github.com/OCA/mis-builder/actions/runs/15213332995/job/42792688070?pr=684#step:7:51

I can however use that as last straw to upgrade my customer's python version, as it will be EOL anyways at the end of the year and we forget about 3.9 compatibility

sbidoul: As you prefer. Else just remove the type annotations.

hbrunn: reverted to the initial version. Or did you mean remove all annotations?

sbidoul: Wait, could you try with `from __future__ import annotations` at the top of the file, and keep `int | None` ?

That should work with 3.9 too.

hbrunn: ah right, better

sbidoul:
Thanks!

/ocabot merge patch

PR #683, opened 2025-05-23
[17.0][IMP] mis_builder: Avoid error if a user does not have the Administration/Settings permission

victoralmau:
Avoid error if a user does not have the Administration/Settings permission

**Before**
antes

**After**
despues

Please @pedrobaeza can you review it?

@Tecnativa TT56480

PR #685, opened 2025-05-27
[18.0][IMP] mis_builder: Avoid error if a user does not have the Administration/Settings permission

victoralmau:
FWP from 17.0: https://github.com/OCA/mis-builder/pull/683

Avoid error if a user does not have the Administration/Settings permission

**Before**
antes

**After**
despues

imagen

@Tecnativa TT56480


sbidoul:
I cherry-picked this in #652

PR #689, opened 2025-06-03
[18.0][PORT] mis_builder: introduce annotation in reports

AnizR:
Forward port of #676


vdewulf:
live from Cartagena, Spain, during the Spanish OCA Days

Tested MIS Builder Annotation in v18 with runboat, selected the database with all modules installed.

There was no demo data because the demo data module was not migrated yet.

Steps:
- created a super basic template with one line with the expression: crd[4%]
- created a MIS report with the date range 2025 that I created on the fly, using the template from previous step
- created the 2025 date range from the MIS Report view, which opened a pop up showing "archived" on the right. Why archived? When selecting a value on the "type" field, the archived word disappeared. ==> this is a bug on Date range, nothing to do with this PR
- displayed the report. On the right side of the report, next to the data cell, there is now an arrow showing. When clicking the arrow, there is "Annotate" that appears. 
image

- Clicking there opens a popup to add text.
- Confirm => a little "1" appears next to the data, and the text of the annotation is displayed at the bottom of the report.
- it is possible to delete the note in one click from the footnote zone. It works.
image


==> CCL: this works as expected.

Side note: this feature could be added in the documentation of the module!

vdewulf:
Tested the export in PDF and Excel.
In excel, it works as expected but in PDF the annotation number is too big:
image

AnizR:
@vdewulf thanks for your review! 

I've made changes to address your remarks.

vdewulf:
Export in pdf tested again, now it is nice!
image

sbidoul (approved):

sbidoul:
@etobella if you have a moment, would you like to review the js part here?

mis_builder/static/src/annotation_dialog/annotation_dialog.esm.js
```diff
@@ -0,0 +1,31 @@
+import {Component, useState} from "@odoo/owl";
+import {Dialog} from "@web/core/dialog/dialog";
+
+export class AnnotationDialog extends Component {
+    static components = {Dialog};
+    static props = {
+        close: Function,
+        annotationText: {type: String, optional: true},
+        confirm: {type: Function, optional: true},
```
etobella: If you are making functions optional, you should define the defaultProps, as you are using it later. Otherwise, make it required

mis_builder/static/src/components/mis_report_widget.xml
```diff
@@ -81,21 +82,65 @@
...
-                                    </t>
-                                    <t t-if="!cell.drilldown_arg">
-                                        <t t-out="cell.val_r" />
-                                    </t>
+                                    <div>
+                                        <t t-if="cell.drilldown_arg">
+                                            <a
+                                                href="javascript:void(0)"
+                                                class="mis_builder_drilldown"
+                                                t-on-click="drilldown"
```
etobella: set href="#" and t-on-click that prevents default `t-on-click.prevent="drilldown"`

https://github.com/odoo/owl/blob/master/doc/reference/event_handling.md#modifiers

etobella (approved):
I made some small non blocking comments on the code review.

BTW, if the cell is empty, you cannot anotate. is it expected?

pedrobaeza:
This has conflicts to resolve.

FrancoMaxime (approved):
LGTM: code review

PR #653, opened 2024-12-03
[18.0][MIG] mis_builder_budget: Migration to 18.0

chaule97:
- Depends on: https://github.com/OCA/mis-builder/pull/652

PR #690, opened 2025-06-03
[18.0][MIG] mis_builder_budget

sbidoul:
Continuing from #653


vdewulf:
live from Cartagena, Spain, during the Spanish OCA Days

Tested MIS Builder Budget on v18 runboat in database with all modules installed.

There was no demo data because the demo data module was not migrated yet.

Steps:
- created a super basic template with one line with the expression: crd[4%]
- created a MIS report with the date range 2025 that I created on the fly, using the template from previous step
- created the 2025 date range from the MIS Report view, which opened a pop up showing "archived" on the right. Why archived? When selecting a value on the "type" field, the archived word disappeared. ==> this is a bug on Date range, nothing to do with mis builder budget.
- displayed the report: the report shows the right amount based on the accounting items
- created a new budget (based on accounts)
- [solved compared to PR653] created a new item for this budget. Here, the Date range field correctly shows the already existing 2025 date range.
- adapted the mis report to have 2 columns in comparison mode, one with actuals, the other with the budget created previously. It works (actuals shows 127k, budget shows 500k) for the whole 2025 year.
- adapted the report to get a view on half a year only for the budget column by changing the date ranges to 31st May. The report displays now 206 849€, which is different from the 5/12 of 500 000€, I suppose the prorata temporis is based on days of the year.

==> CCL: bug with access to date range data is now solved.

sbidoul:
Thanks @vdewulf !

Since this was technically reviewed in #653, going ahead with the merge.

/ocabot merge nobump

PR #678, opened 2025-05-08
[16][IMP] mis_builder: improve ui by adding button to resize report instance

AnizR:
## Description

Small UI improvement: add possibility to widen the "sheet" of a report instance.
This feature has been developed mainly for report with multiple columns.

### Display setting

On instance report, there's a new setting 'wide_display_by_default' (page 'layout') that sets the default view mode of the report. 

### Screenshots

image
image
image

PR #693, opened 2025-06-04
[18.0][PORT] mis_builder: improve ui by adding button to resize report instance

AnizR:
port forward of #678 

## Small modification

When widening, I now set the 'sheet wide' to be 100% (previously 90%).

## Description

Small UI improvement: add possibility to widen the "sheet" of a report instance.
This feature has been developed mainly for report with multiple columns.

### Display setting

On instance report, there's a new setting 'wide_display_by_default' (page 'layout') that sets the default view mode of the report.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -592,6 +592,10 @@ class MisReportInstance(models.Model):
         help="Search view to customize the filter box in the report widget.",
     )
 
+    wide_display_by_default = fields.Boolean(
+        string="Open report in wide mode by default",
+    )
+
     @api.depends("report_id.move_lines_source")
     def _compute_widget_search_view_id(self):
         for rec in self:
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -16,6 +16,7 @@ const config = [{
             openerp: "readonly",
             owl: "readonly",
             luxon: "readonly",
+            document: "readonly",
         },
 
         ecmaVersion: 2024,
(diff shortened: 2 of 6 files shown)


sbidoul:
Would it possible to have a white background on the expand icon ? I think it would look nicer.

AnizR:
> Would it possible to have a white background on the expand icon ? I think it would look nicer.


Thanks for the review
Done: 
image

mis_builder/models/mis_report_instance.py
```diff
@@ -592,6 +592,10 @@ def _compute_pivot_date(self):
         help="Search view to customize the filter box in the report widget.",
     )
 
+    wide_display_by_default = fields.Boolean(
```
pedrobaeza, line 595: Wasn't the previous behavior to be expanded by default and now we are setting the opposite one?

AnizR: Hello, 
No, by default it is (and was) not expanded.

But, if you would like to have it 'by default expanded' on a report, you can set the field 'Open report in wide mode by default' to true on the form 'instance.report', page 'layout'

pedrobaeza: Hi @AnizR I think I know what is happening: `web_responsive` is affecting it to expand it... I will check how both this new feature and the module interact together.

AnizR: Hi @pedrobaeza, okay thanks, let me know if I need to do any change

PR #700, opened 2025-06-08
[18.0][FIX] mis_builder_budget: reintroduce missing fields

sbidoul:
These fields were erroneously removed during migration in #653 and #690


sbidoul:
This fixes a bug with date range selection similar to what was reported in https://github.com/OCA/mis-builder/pull/690#issuecomment-2935325020

PR #708, opened 2025-07-28
[IMP] mis_builder: Sort reports by sequence

u0f:
When there are many different reports for various use cases and companies, it's difficult to quickly find the desired report because the list view is ordered by ID, showing the oldest entries first.

The solution is adding a sequence field to the `mis_report_instance` list view to allow custom ordering and improve usability.

https://www.loom.com/share/52ba1531f4254f3c9823422c3bd7b42b?sid=d232990a-0a55-4847-aa8e-97f7783f9032

MT-10677 

cc @moduon @yajo @EmilioPascual @Andrii9090

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -476,8 +476,10 @@ class MisReportInstance(models.Model):
 
     _name = "mis.report.instance"
     _description = "MIS Report Instance"
+    _order = "sequence, id"
 
     name = fields.Char(required=True, translate=True)
+    sequence = fields.Integer(default=10)
     description = fields.Char(related="report_id.description", readonly=True)
     date = fields.Date(
         string="Base date", help="Report base date " "(leave empty to use current date)"
--- a/mis_builder/views/mis_report_instance.xml
+++ b/mis_builder/views/mis_report_instance.xml
@@ -27,6 +27,7 @@
         <field name="model">mis.report.instance</field>
         <field name="arch" type="xml">
             <tree>
+                <field name="sequence" widget="handle" />
                 <button
                     type="object"
                     name="preview"


mis_builder/models/mis_report_instance.py
```diff
@@ -478,6 +478,7 @@ def _compute_pivot_date(self):
     _description = "MIS Report Instance"
 
     name = fields.Char(required=True, translate=True)
+    sequence = fields.Integer()
```
EmilioPascual:
```suggestion
    sequence = fields.Integer(default=10)
```

The field sequence should have a value by default.
Also it should be in `_order` of the the model.

mis_builder/views/mis_report_instance.xml
```diff
@@ -40,6 +40,7 @@
                     string="Export"
                     icon="fa-download"
                 />
+                <field name="sequence" widget="handle" />
```
EmilioPascual: That is unusual. The sequence is usually the first one for better usability.
IMHO: I would add the sequence to the form view. WDYT?

EmilioPascual (changes requested):
Good job @u0f. Thank you for the improve.
However, there are some changes.

mis_builder/views/mis_report_instance.xml
sbidoul: Here we have the buttons on the left, so it might make sense. On runboat it looks ok.

u0f: Now
Captura desde 2025-07-28 10-55-48

With the proposed change
Captura desde 2025-07-28 10-55-51

yajo: IMHO I agree with @EmilioPascual. I think that having the handle as the leftmost element looks better.

Could you change it please @u0f?

mis_builder/models/mis_report_instance.py
```diff
@@ -476,8 +476,10 @@ def _compute_pivot_date(self):
 
     _name = "mis.report.instance"
     _description = "MIS Report Instance"
+    _order = "sequence"
```
```suggestion
    _order = "sequence,id"
```

So it's stable.

mis_builder/views/mis_report_instance.xml
```diff
@@ -27,6 +27,7 @@
         <field name="model">mis.report.instance</field>
         <field name="arch" type="xml">
             <tree>
+             <field name="sequence" widget="handle" />
```
sbidoul:
```suggestion
               <field name="sequence" widget="handle" />
```

mis_builder/views/mis_report_instance.xml
```diff
@@ -104,6 +105,7 @@
                     <group>
                         <group>
                             <field name="report_id" string="Template" />
+                            <field name="sequence" />
```
sbidoul: Is that really useful to add it in the form? Most users won't know what to do with it.

u0f: In a scenario where the database has 50–60 reports and the user wants to move one of the ones at the top, it's useful to be able to change the sequence manually and assign a specific value, instead of having to drag it through all the others to find the right position. 

It's not really something we need, but it can be useful.

sbidoul: I would start without and add it later if the use case really arises.

sbidoul (approved):

EmilioPascual (approved):

yajo (approved):

sbidoul:
@u0f do you plan to forward port this to 18.0?

u0f:
> @u0f do you plan to forward port this to 18.0?

Hey, just forwarded on #731

PR #712, opened 2025-08-04
[FIX] mis_builder xlsxwriter compatibility

glitchov:
xlsxwriter has removed the internal aliases set_font() (set_font_name()), set_size() (set_font_size()) and set_color() (set_font_color()) from version 3.2.1 but reintroduced them in 3.2.3 to avoid code break. Since these aliases were there only for internal purpose they will be removed (https://xlsxwriter.readthedocs.io/changes.html).
So mis_builder is moving forward with the right names.

### DIFF
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -272,7 +272,7 @@ class MisReportKpiStyle(models.Model):
         xlsx_attributes = [
             ("italic", props.font_style == "italic"),
             ("bold", props.font_weight == "bold"),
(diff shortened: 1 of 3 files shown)


sbidoul (approved):
Code review and test on runboat (which has xlsxwriter 1.1.2)

sbidoul:
FW port to 18 in #713

sbidoul:
Since this works fine with xlsxwriter 1.1 (here) and 3.0 (in #713), this should be safe.

/ocabot merge patch

PR #716, opened 2025-09-04
[18.0][IMP] mis_builder: small usability improvements

alexis-via:
- Add sequence field to allow users to configure the order of the report instance.
- Add search view for mis.report.instance.
- Improve string for field target_move, to make it clear that canceled journal entries are not taken into account


sbidoul:
LGTM, but
- the sequence was added in #708 which should be forward ported independently
- can you do the other two changes in independent PRs to facilitate backports?

sbidoul:
The sequence field has been ported to 18 in https://github.com/OCA/mis-builder/pull/731

sbidoul:
@alexis-via can you do two separate PRs for the other improvements?

PR #717, opened 2025-09-15
[18.0][FIX] mis_builder: Fix JS error when adding custom filter

ThiagoMForgeFlow:
When a user attempts to add a custom filter while previewing a report, the system throws an error related to the dialog service.

@ForgeFlow


sbidoul:
Thanks! 

/ocabot merge patch

PR #724, opened 2025-09-16
[FIX] mis_builder: Allow to copy mis.report.instance models in case it has comparison (and/or sum?) periods

### LINKED ISSUE #714: [v18] Error when duplicating a report
## Error when duplicating a report
When we want to duplicate a report, it gives an error, as we can see in the video:

https://www.loom.com/share/16e30df74e674278b30ecd0716fb490d

## To Reproduce
**Affected versions**:

1. Go to mis reports
2. Select one
3. Try to duplicate it

**Expected behavior**
The report is expected to be successfully duplicated.

legalsylvain:
Fixes: #714

Rational: For the time being, copy_data is failing, if there are many periods in an mis.report.instance, because self.source only works with a singleton.

Bug introduced in version 13.0, in #343

CC : @LoisRForgeFlow

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -461,15 +461,14 @@ class MisReportInstancePeriod(models.Model):
                     )
 
     def copy_data(self, default=None):
-        if self.source == SRC_CMPCOL:
-            # While duplicating a MIS report instance, comparison columns are
-            # ignored because they would raise an error, as they keep the old
-            # `source_cmpcol_from_id` and `source_cmpcol_to_id` from the
-            # original record.
-            return [
-                False,
-            ]
-        return super().copy_data(default=default)
+        # While duplicating a MIS report instance, comparison columns are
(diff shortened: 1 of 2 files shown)


sbidoul (approved):

sbidoul:
Actually, I think this should also filter out `SRC_SUMCOL`, because both are references to other columns.
In ideal world, it should map the ids of the old periods to the ones in the copy, but that is a bit complicated, so removing the problematic columns from the copy is good enough solution.

LoisRForgeFlow (approved):
@legalsylvain Thanks, I've tested in a real instance :+1: 

Apart from @sbidoul comment, LGTM

sebastienbeau:
/ocabot merge patch
@legalsylvain thanks for the fix !

legalsylvain:
You're welcome. I have some pending PR to review. Feel free to take a look, if you have some time. 

https://github.com/OCA/mis-builder/pulls/legalsylvain

thanks !

PR #725, opened 2025-09-17
[18.0][MIG/IMP] mis_builder: analytic_distribution_search removed in 18

LoisRForgeFlow:
analytic_distribution_search has been removed in v18 and it was set on purpose on mib_builder migration scripts in 16.0. I think we need to handle it as well in 18.0 migration.

### DIFF
--- a/mis_builder/migrations/18.0.1.3.0/post-migration.py
+++ b/mis_builder/migrations/18.0.1.3.0/post-migration.py
@@ -1,0 +1,40 @@
+# Copyright 2025 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+from odoo.tools.safe_eval import safe_eval
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    """Update the value of the analytic_domain field."""
+    # Workaround to execute the migration script without errors
+    # see https://github.com/odoo/odoo/blob/2a839ef1ed09c36f27ce7536ca3052d9f65ceed9/odoo/modules/migration.py#L252-L256
+    env = cr
+    for record in env["mis.report.instance.period"].search(
+        [("analytic_domain", "!=", False)]
+    ):
+        new_domain = _update_domain(record)
+        record.write({"analytic_domain": new_domain})
+
+    for record in env["mis.report.instance"].search([("analytic_domain", "!=", False)]):
+        new_domain = _update_domain(record)
+        record.write({"analytic_domain": new_domain})
+
+
+def _update_domain(record):
+    # analytic_distribution_search has been removed in v18 and it was set on purpose
+    # on mis_builder migration scripts in 16.0.
+    domain = safe_eval(record.analytic_domain)
+    new_domain = []
+    for clause in domain:
+        if (
+            isinstance(clause, list | tuple)
+            and clause[0] == "analytic_distribution_search"
+        ):
+            operator = clause[1]
+            value = clause[2]
+            clause = ("distribution_analytic_account_ids", operator, value)
+        new_domain.append(clause)
+    return new_domain
--- a/mis_builder/README.rst
+++ b/mis_builder/README.rst
@@ -763,7 +763,7 @@ promote its widespread use.
 
 Current `maintainer <https://odoo-community.org/page/maintainer-role>`__:
 
-|maintainer-sbidoul| 
+|maintainer-sbidoul|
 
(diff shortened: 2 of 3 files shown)


mis_builder/migrations/18.0.1.0.0/post-migration.py
LoisRForgeFlow, line 1: @sbidoul not sure about the correct version to add the scripts, as it should have been applied from the very beginning but we are already at 18.0.1.2.1

legalsylvain (changes requested):
You should create a migration script in a folder with a name greater than the current version. (18.0.1.2.5 for now). Otherwise, it will not be executed on your instance.

Solution : 
- create a file named `mis_builder/migrations/18.0.2.0.0/post-migration.py`
- When, merging, ask @sbidoul to call `/ocabot merge major`

LoisRForgeFlow:
> You should create a migration script in a folder with a name greater than the current version. (18.0.1.2.5 for now). Otherwise, it will not be executed on your instance.
> 
> Solution :
> 
> * create a file named `mis_builder/migrations/18.0.2.0.0/post-migration.py`
> * When, merging, ask @sbidoul to call `/ocabot merge major`

I don't think this is any major change, if you prefer to go that path I would manually bump version and ask for a nobump merge. The problem for me appears now migrating to 18 from previous versions, so this script will be executed even if add it for 18.0.1.0.0. 

I was raising the questions because it can be that someone that already migrated to 18.0 manually fixed the error so I was not sure if they would like this script to be executed. But looking at it again, it the problem was manually fixed, this script would do nothing... so if you think it is fine, I will go with the manual bump + nobump-merge, or do you see enough reason to do a major bump here?

LoisRForgeFlow:
Manual bump done, to be merged with "nobump" option.

legalsylvain:
> Manual bump done, to be merged with "nobump" option.

Manual bump will conflict when the next PR on mis-builder will be merged. So it's ok, if the current PR is merged very quickly.

> do you see enough reason to do a major bump here?

 just to avoid that another minor bump done in another PR makes your script unavailable. If this PR is merged quickly it will not occures.


> But looking at it again, it the problem was manually fixed, this script would do nothing.

Writing idempotens migration scripts is a very good design. :+1:

LoisRForgeFlow:
It should be ready now, I had some problems with discrepancies between my local pre-commit and CI one.

legalsylvain (approved):

StefanRijnhart (approved):

mis_builder/migrations/18.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,37 @@
...
+        record.write({"analytic_domain": new_domain})
+
+    for record in env["mis.report.instance"].search([("analytic_domain", "!=", False)]):
+        new_domain = _update_domain(record)
+        record.write({"analytic_domain": new_domain})
+
+
+def _update_domain(record):
+    # analytic_distribution_search has been removed in v18 and it was set on purpose
+    # on mib_builder migration scripts in 16.0.
```
sbidoul:
```suggestion
    # on mis_builder migration scripts in 16.0.
```

mis_builder/migrations/18.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,37 @@
...
+    domain = safe_eval(record.analytic_domain)
+    new_domain = []
+    for clause in domain:
+        if clause[0] == "analytic_distribution_search":
```
sbidoul: Should this test that `isinstance(clause, Sequence)`. I'm thinking of cases where clause is `|` or `&`.

LoisRForgeFlow: @sbidoul I think that will not work as expected:

```python
isinstance("&", Sequence)
True
isinstance("|", Sequence)
True
# If you are concerned about the "out of index error", it will not raise
"|"[0]
'|'
```

sbidoul: Ah right, a string is a Sequence indeed. So `isinstance(clause, (list, tuple))`.

mis_builder/migrations/18.0.1.3.0/post-migration.py
LoisRForgeFlow: fixed!

mis_builder/migrations/18.0.1.3.0/post-migration.py
LoisRForgeFlow: Perfect, just added that as well.

sbidoul:
/ocabot merge nobump

Thanks!

Issue #726, opened 2025-09-18
[V18/V19] splitting module in many more little modules


legalsylvain:
Hi all,

First thanks for this great module !
I just discovered ``mis_builder`` during the last OCA days (in the training courses), and it is very powerful tool. It is super easy, super fast and super simple. As Odoo.

However I have a question, regarding the design. When I discovered the code, I was surprised to see that the base module ``mis_builder``, is depending on ``account`` : 
- In fact MIS mean "Management Information System", that is not related to account.
- after all, it's totally possible to want the MIS feature out of accounting context. (with sub queries, on any model). In fact, we could imagine some people wanting to use this tools in a framework project, without any ``product`` / ``account`` module installed.
- in very practical terms, I have an instance, where accounting data are not complete / correct. (account moves are exported in EBP, with other account moves that comes from other softwares). As a result, the "bal" features will not works, as Odoo doesn't have all the accounting data (or sometime, moves are changed in EBP, but not in Odoo). In that context, having a lot of useless code installed is not optimal.

As a result,
- the "main" ``mis_builder`` module is quite big.
- Users that don't have accounting access right can not see MIS reports.

I have so the idea to split modules into many ones.

**Structure changes** 

``mis_builder`` changes : split into :
- ``mis_builder_base``, depends = "base"
- ``mis_builder_board``, depends = ``["mis_builder_base", "board"]``, auto_install = True
- ``mis_builder_report_xlsx``, depends = ``["mis_builder_base", "report_xlsx"]``, auto_install = True
- ``mis_builder_date_range``, depends = ``["mis_builder_base", "date_range"]``, auto_install = True
- ``mis_builder_account``, depends = ``["mis_builder_base", "account"]``

``mis_builder_budget`` changes : 
- now depends on ``["mis_builder_account"]`` (before ["mis_builder", "account"]``)

**Optionally**

- keep an empty module ``mis_builder``, depends = ``["mis_builder_base", "mis_builder_board", "mis_builder_report_xlsx", "mis_builder_date_range", "mis_builder_account"]``, application=True, so for functional people, no changes, when making demo, etc.

**Main UX changes :**

- create a dedicated main "MIS" menu, with the existing icon : https://github.com/OCA/mis-builder/blob/18.0/mis_builder/static/description/icon.png and put all the existing MIS menu entries into the new dedicated menu. (reports & configuration)

**Important note :**

[...]

PR #728, opened 2025-09-19
[18.0][REF][ADD] mis_builder_report_xlsx

legalsylvain:
- depends on https://github.com/OCA/mis-builder/issues/726

### Creation of `mis_builder_report_xlsx`

- [x] Extract "xlsx" feature from `mis_builder`, creating `mis_builder_report_xlsx`
- [x] migration scripts : 
    - rename `mis_builder.xls_export` to  `mis_builder_report_xlsx.xls_export` 
    - install new module `mis_builder_report_xlsx` 
- [x] Create Readme section in the new module

### cloc of mis_builder

**Before :** 

```
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Python                          40            699            935           5667
XML                              9              2              6           1169
JavaScript                       1             24             15            164
```

**After :** 

```
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Python                          39            675            918           5395  (-5%)
XML                              8              2              6           1136  (-3%)
JavaScript                       1             23             15            155  (-5%)

```

### Description

This module improves the MIS Reports modules, allowing to download MIS Reports in XLSX format.

### Usage

- Go to MIS Reports list and form views.
- Click on 'Export' Button to download your XLSX file.

image

- In the same way, a button is present in the 'preview' form view of your MIS reports.

image

@sbidoul : what do you think ? 

Regards.


mis_builder/migrations/18.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+from openupgradelib import openupgrade
+
+xmlid_renames = [
+    ("mis_builder.xls_export", "mis_builder_report_xlsx.xls_export"),
+]
```
sbidoul, line 9: Is this not done automatically?

mis_builder_report_xlsx/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2017-2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "MIS Builder - Report XLSX",
+    "summary": "Glue module for MIS Builder and Report XLSX"
+    " that allow to export MIS report in XSLX format.",
```
sbidoul:
```suggestion
    "summary": "Export MIS Reports in XLSX format",
```

mis_builder_report_xlsx/models/mis_report_instance.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+SRC_ACTUALS = "actuals"
+SRC_ACTUALS_ALT = "actuals_alt"
+SRC_CMPCOL = "cmpcol"
+SRC_SUMCOL = "sumcol"
+
+MODE_NONE = "none"
+MODE_FIX = "fix"
+MODE_REL = "relative"
```
sbidoul: Why are these constants here?

mis_builder_report_xlsx/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV, GRAP, Odoo Community Association (OCA)",
```
sbidoul, line 9: Do feel it's important to add GRAP as author here. In a way it is not "new" as it is shuffling code around.

sbidoul (commented):
Thanks for this, Sylvain. I'm fine with extracting that part if you think it's useful. 

It looks good to me overall. I have not verified all the code was moved correctly but I found one suspicious thing. See the comments.

pedrobaeza (commented):
I think this should be done in 19, not in this version.

legalsylvain:
> I think this should be done in 19, not in this version.

Hi @pedrobaeza. Thanks for your answer ! In fact, this PR is, for the time being, just a PoC I did regarding the proposal present in https://github.com/OCA/mis-builder/issues/726.
As said in [#726, ](https://github.com/OCA/mis-builder/issues/726#issue-3429264017)

_As a result, my question here is ONLY "in an ideal world, Do you think it's better to split modules, or to keep a single big one ?".
Once this first question answered, and if refactor is desired by PSC, we could focus on the next question : "on which version we are applying this refactor ? (current V18 version, or new V19 release)._

Could you so give your PoV on the other main issue ? 
thanks !

pedrobaeza:
Although the excel export requires the extra dependency, we already have it also on account_financial_report, so for me it's not critical to split it, and I will always install both.

legalsylvain:
> Although the excel export requires the extra dependency, we already have it also on account_financial_report, so for me it's not critical to split it, and I will always install both.

Hi. I didn't make myself clear. I wanted you to comment on the other issue https://github.com/OCA/mis-builder/issues/726.

thanks.

mis_builder_report_xlsx/__manifest__.py
legalsylvain: Sorry, I don't understand. Do you want me to remove GRAP, or not?

Personally, I usually add my company name when I spend a significant amount of time on a module. (development, refactoring, etc.)
But it's a matter of taste, so tell me what to do!

mis_builder_report_xlsx/models/mis_report_instance.py
legalsylvain: Good catch. Totally useless. Removed.

mis_builder/migrations/18.0.2.0.0/pre-migration.py
legalsylvain: if the original database is : 


```
 id  |                    name                     |       type        
-----+---------------------------------------------+-------------------
 288 | {"en_US": "MIS report instance XLS report"} | ir.actions.report
```


**With that script:** 

- during update

2025-09-25 09:25:36,222 71906 DEBUG mis_builder_v1 OpenUpgrade: 1 rows affected after 0:00:00.000510 running UPDATE ir_model_data SET module = 'mis_builder_report_xlsx', name = 'xls_export' WHERE module = 'mis_builder' and name = 'xls_export' 

- in database, after the update

```
 id  |                    name                     |       type        
-----+---------------------------------------------+-------------------
 288 | {"en_US": "MIS report instance XLS report"} | ir.actions.report
```

**Without that script**

- during update

2025-09-25 09:30:18,601 72300 INFO mis_builder_v2 odoo.addons.base.models.ir_model: Deleting 288@ir.actions.report (mis_builder.xls_export) 

- in database, after the update

```

 id  |                    name                     |       type        
-----+---------------------------------------------+-------------------
 289 | {"en_US": "MIS report instance XLS report"} | ir.actions.report

```


So it can be usefull if this action was referenced in some ``action_id`` field.
(It's maybe quite theoritical, but it's an OpenUpgrade good practice)

sbidoul:
Thinking about this again I now have some doubts.

For instance, when the annotations feature was added it was added on the html, pdf and xlsx formats at once.
If the xlsx export had been in a separate module maybe the contributor could have argued that it was not necessary to implement it the Excel export and that would have been left with a PR doing only a part of the work. 

That is very hypothetical of course, but I worry a bit that the different parts would then evolve at a different speed and therefore lead to a less cohesive result.

legalsylvain:
Hi @sbidoul. Well, I think that each design has its limitations.

- Having a single module is harder to port, requires more dependencies, force users to install potentially non desired features.
- With many modules, some users could ignore some features, because the features are present in an non installed module. Also it allows to have features not fully implemented, as you describe.
To overcome this last concern, we can force contributors to implement each features in pdf / html / xlsx format. (whatever if there is one or more modules, it can be a prerequite to contribute).

Finally, It's up to you ! You are the chief architect of this repository !

Please let me know your decision.

legalsylvain:
Hi @sbidoul. could you finalize your review here : 
- merge this PR, if you agree.
- close this PR, if you disagree.

Migration to V19 begin to be done. This work is starting to get moldy ;-)

Thanks !

sbidoul:
I'm going to close this, as I think XLSX export is an integral part of the core MIS Builder app.

Thanks for your patience and sorry for the hesitation.

legalsylvain:
Thanks for your answer, @sbidoul.

Regarding my last PR on mis-builder, could you take a time to answer to that question when you have time ? thanks ! https://github.com/OCA/mis-builder/pull/732#issuecomment-3338646696

PR #729, opened 2025-09-20
Pseudo code for a P&L consistency check

sbidoul:
@alexis-via here is how I imagine a P&L consistency check.

For balance sheet it's a bit more complex due to nbal/pbal but the same kind of logic should work.

Feel free to pick it up from here and let me know if you have any questions.


sbidoul:
Continuing in #785

PR #711, opened 2025-07-31
[FIX][18.0] mis_builder: Error in multicompany report

Xino61122:
Description of the issue/feature this PR addresses:

Current behavior before PR:

When we have two companies selected, and we go to Invoicing > Reporting > MIS reports, select a report template, check ‘Multiple companies’ and click on preview, we get an error message.

Desired behavior after PR is merged:

The error no longer appears, and the preview is displayed correctly.

https://www.loom.com/share/cfeb4d5ee7a2431481d263ae6c2058ea?sid=d87d3d57-f3ab-49a3-8b07-77bb1cdd3d22

fix https://github.com/OCA/mis-builder/issues/710

cc: @Moduon MT-10671

PR #730, opened 2025-09-22
[18.0][FIX] mis_builder: account.account has company_ids in Odoo 18

StefanRijnhart:
Adapted from https://github.com/OCA/mis-builder/pull/711, but observing that in Odoo 18, there is no `company_id` field anymore, and the `company_ids` field is required, thus removing several test cases.

Limit display to 3 companies plus ellipsis.

Fixes
```
  File "/home/odoo/mis-builder/mis_builder/models/kpimatrix.py", line 478, in _get_account_name
    result = f"{result} [{account.company_id.name}]"
                          ^^^^^^^^^^^^^^^^^^
AttributeError: 'account.account' object has no attribute 'company_id'
```

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -469,7 +469,12 @@ class KpiMatrix:
     def _get_account_name(self, account):
         result = f"{account.code} {account.name}"
         if self._multi_company:
-            result = f"{result} [{account.company_id.name}]"
+            company_names = ", ".join(
+                [c.name for c in account.company_ids[:3]]
+            )  # Limit to first 3 companies
+            if len(account.company_ids) > 3:
+                company_names += ", ..."
+            result = f"{result} [{company_names}]"
         return result
 
     def get_account_name(self, account_id):
--- a/mis_builder/tests/__init__.py
+++ b/mis_builder/tests/__init__.py
@@ -7,6 +7,7 @@ from . import test_multi_company_aep
 from . import test_aggregate
 from . import test_data_sources
 from . import test_kpi_data
+from . import test_kpimatrix_account_names
 from . import test_mis_report_instance
 from . import test_mis_safe_eval
 from . import test_period_dates
--- a/mis_builder/tests/test_kpimatrix_account_names.py
+++ b/mis_builder/tests/test_kpimatrix_account_names.py
@@ -1,0 +1,126 @@
+# Copyright 2025 Geraldo Lopez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from unittest.mock import MagicMock, Mock
+
+from odoo.tests import common
+
+from ..models.kpimatrix import KpiMatrix
+
+
+class TestKPIMatrixAccountNames(common.TransactionCase):
+    """Unit tests for KpiMatrix._get_account_name() method enhancements"""
+
+    @classmethod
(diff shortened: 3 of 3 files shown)


mis_builder/models/kpimatrix.py
```diff
@@ -475,7 +475,12 @@ def _load_account_names(self):
     def _get_account_name(self, account):
         result = f"{account.code} {account.name}"
         if self._multi_company:
-            result = f"{result} [{account.company_id.name}]"
+            company_names = ", ".join(
+                [c.name for c in account.company_ids[:3]]
+            )  # Limit to first 3 companies
+            if len(account.company_ids) > 3:
+                company_names += ", ..."
+            result = f"{result} [{company_names}]"
```
sbidoul, line 477: Thanks @StefanRijnhart ! This should fix the bug indeed.

Functionally though I'm not sure.

I'd say, we at least need to show only companies that are in the report `query_company_ids`.
That should be relatively easy to do by passing this list of companies when constructing the KpiMatrix, next to or in place of the `multi_company` flag.

But, I wonder if it makes sense functionally to show, say, debit/credit by account with multiple companies on the same line.

Thinking even further I wonder if it is technically possible to have a KpiMatrixRow for an account detail that concern multiple companies, or if we rather always have one row per (account, company). In that case, the company must be a property of the Row.

I'll try to do a few tests to better (re)understand.

ThiagoMForgeFlow: Hi @sbidoul, have you been able to do the tests you mentioned?

sbidoul: Not yet. In the meantime any functional feedback of users testing this PR in various circumstances is welcome.

ThiagoMForgeFlow: Recently I have used this PR to compare customer data between v15 and v18 in multi-company MIS reports, and the results seem good.

sbidoul: Ok, thanks. I'll need to find answers to my questions above, though.

Xino61122:
Hello @StefanRijnhart, can do a rebase to test this change?
Thank you for your job :smiley:.

mis_builder/models/kpimatrix.py
Xino61122: Hello, thank you very much for your work.
I see a problem in how Odoo does things in this version, which is that the company to be represented must be selected in the context, otherwise account.code returns False and that is what is displayed on the screen, so it would never enter in _if len(account.company_ids) > 3:_

I upload screenshot with my local whit this changes:

image

StefanRijnhart:
@Xino61122 done.

mis_builder/models/kpimatrix.py
```diff
@@ -469,7 +469,12 @@ def _load_account_names(self):
     def _get_account_name(self, account):
         result = f"{account.code} {account.name}"
```
Xino61122, line 470: To ensure that False is not entered as the code, we can add the following to the code
```
if not account.code:
            account = account.with_company(account.company_ids[0])
```
This way, it takes the company code from the current account. The problem is that companies with the same accounting account still do not appear on the same line, but it is a possible solution.

This is how it would look with the change.

image

StefanRijnhart:
Closing in favour of https://github.com/OCA/mis-builder/pull/756

PR #732, opened 2025-09-26
[REF] mis_builder: create dedicated 'MIS Reports' Entry menu for mis_builder application

legalsylvain:
1. Create dedicated menu entry

2. change icon.

**Icon before**

image

**Icon after**

(put some similar style as in V18).

image


legalsylvain:
Hi @sbidoul

i have a question for the groups.

for the time being : 
- menu of "MIS Reporting" is accessible to ``account.group_account_manager``
- menu of "MIS Configuration" is accessible to ``account.group_account_manager``
- all models are readable by base.group_user
- all models are writable by account.group_account_manager

As a result, by API, any user can read all the data, but it is hidden in the UI, and only visible by accountant.

What I propose : 
- creation of a ``mis_builder.user`` group, that can read report. (but not templates).
- create of a ``mis_builder.manager`` group, that can edit all MIS models
In the migration script, all the users of account.group_account_manager are now member of account.group_account_manager to avoid loosing access during migration.

Is that what you had in mind?

Thanks !

legalsylvain:
Hi @OCA/mis-builder-maintainers ! friendly reminder here.

Thanks !

legalsylvain:
Hi @OCA/mis-builder-maintainers. any point of view, regarding my question ? 

https://github.com/OCA/mis-builder/pull/732#issuecomment-3338646696

thanks !

legalsylvain:
@sbidoul : any chance to have an answer on that topic ? Or I close this PR ? 

(if it's more simple, we can have a little talk, directly).

thanks !

sbidoul:
Hi Sylvain,

To progress in small steps, maybe we can start with the security groups first without changing the menus?

There was some preliminary discussion about this in https://github.com/OCA/mis-builder/issues/639#issuecomment-2466693705. So perhaps we could create a new issue to discuss a rework of the groups?

legalsylvain:
> To progress in small steps, maybe we can start with the security groups first without changing the menus?

Of course ! I'll take a look on the previous discussion and propose a PR if possible !

sbidoul:
I suggest making an issue with a proposal first, but up to you if you prefer expressing the proposal in code :)

PR #734, opened 2025-09-30
 [18.0][IMP] mis_builder: Add date_range_id field to mis_builder report preview

Andrii9090-tecnativa:
Date ranges can be selected in the mis builder report preview, and the information displayed is updated according to the selected date range. The date ranges can't be changed from the Dashboard.

## Description
When an MIS report is not based on a pivot date but instead on a date range, the system will display a date range selector in the report preview, provided that the “Show pivot date” option is enabled.

From the report preview, users can choose the desired date range, and the displayed information will automatically update based on the selected period.

## Test
https://www.loom.com/share/9bb0c8f938ff4cdc8b5f790f959e72e1?sid=45660243-d311-4718-8bd1-a75390663e1a

MT-10680 @moduon


sbidoul (commented):
By design the dates are on `mis.report.instance.period`, so I don't see how this could work. Except maybe if there is only one column.

PR #736, opened 2025-10-13
[16.0][FIX] mis_builder: fix rows all having None as row_id

huguesdk:
fix rows all having `None` as `row_id` (introduced in #676).

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -47,7 +47,7 @@ class KpiMatrixRow:
 
     @property
     def row_id(self):
-        self._matrix._make_row_id(self.kpi.id, self.account_id)
+        return self._matrix._make_row_id(self.kpi.id, self.account_id)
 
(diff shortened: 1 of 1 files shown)


sbidoul:
Oh. How could that work?

AnizR:
> Oh. How could that work?

The `row_id` doesn't seem to be used.

AnizR (approved):
Thanks for your contribution !

sbidoul:
If it's not used then let's remove it.

AnizR:
> If it's not used then let's remove it.

Well, it depends, @huguesdk, how did you realize that it wasn't set? 
Are you using it?

huguesdk:
we noticed it because we were using it in `l10n_be_mis_report_xml`, but i just changed that (in OCA/l10n-belgium#260).

AnizR:
Ok, I think it can be removed,

sbidoul:
I remove the property in #738

PR #737, opened 2025-10-27
[16.0][FIX] mis_builder: computation of currency conversion rates

AnizR:
## Description

In short, the problem is that Odoo doesn't corrtly cache the `rate` of a currency.

I realized that when changing the currency of a report, the conversion rate used was always the same (computed from the first period).
But, mis builder should use the conversion rate of each period ("end date of period"). It is actually what was intended.
But, there is a bug in Odoo v16.0 and fixed in later versions: https://github.com/odoo/odoo/pull/137609


sbidoul (approved):
Thanks! Would also make sense to test the expected values?

AnizR:
> Thanks! Would also make sense to test the expected values?

Ok, I'll do it.

I'am now writing the fix of "_get_company_rates"

mis_builder/models/aep.py
```diff
@@ -379,10 +379,14 @@ def get_aml_domain_for_dates(self, date_from, date_to, mode):
...
         # get exchange rates for each company with its rouding
         company_rates = {}
-        target_rate = self.currency.with_context(date=date).rate
         for company in self.companies:
             if company.currency_id != self.currency:
-                rate = target_rate / company.currency_id.with_context(date=date).rate
+                rate = self.env["res.currency"]._get_conversion_rate(
+                    from_currency=company.currency_id,
+                    to_currency=self.currency,
+                    company=self.env.company,
```
sbidoul, line 387: Here you use `self.env.company` while before it was `company.currency_id.with_context(date=date).rate`.   Is that intended?

AnizR: Yes, in fact, the company used when doing `.rate` was always the company from the environment.

Moreover, I think that we should take conversion rates on the current company and not any other company.

sbidoul (approved):

lmarion-source (approved):

PR #745, opened 2025-10-30
[18.0][IMP]mis_builder: improve rounding

ChristophAbenthungCibex:
The rounding of pythons built-in may lead to unexpected results. 
The following values are round with built-in python round to 1 digit:
9.15 -> 9.2
9.25 -> 9.2
9.35 -> 9.3
9.45 -> 9.4
9.55 -> 9.6
9.65 -> 9.7
9.75 -> 9.8
9.85 -> 9.8
9.95 -> 9.9
There is no consistency if the value is round up or down. The odoo float_round leads to consitent result. For that example always rounds up:
9.15 -> 9.2
9.25 -> 9.3
9.35 -> 9.4
9.45 -> 9.5
9.55 -> 9.6
9.65 -> 9.7
9.75 -> 9.8
9.85 -> 9.9
9.95 -> 10.0

### DIFF
--- a/mis_builder/models/mis_report_style.py
+++ b/mis_builder/models/mis_report_style.py
@@ -7,6 +7,7 @@ import sys
 
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tools import float_round
 
 from .accounting_none import AccountingNone
 from .data_error import DataError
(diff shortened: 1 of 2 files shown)


sbidoul:
Thanks for this contrib!

Could you add a test like this in `test_render.py`?

```python
    def test_render_ieee754(self):
        self.style.dp_inherit = False
        self.style.dp = 1
        self.assertEqual("9.5", self._render(9.45))
        self.assertEqual("9.6", self._render(9.55))
```

ChristophAbenthungCibex:
> Thanks for this contrib!
> 
> Could you add a test like this in `test_render.py`?
> 
> ```python
>     def test_render_ieee754(self):
>         self.style.dp_inherit = False
>         self.style.dp = 1
>         self.assertEqual("9.5", self._render(9.45))
>         self.assertEqual("9.6", self._render(9.55))
> ```

@sbidoul Thx for the feedback. I added the requested test.

sbidoul (approved):

OKT96 (approved):
LGTM 👍🏻

PR #749, opened 2025-11-01
[IMP] mis_builder: Add annotations column to reports

fsmw:
This PR addresses issue #159.\n\nIt adds a new "Annotations" column to the MIS reports (view, PDF, and XLSX).\n\nThe annotation for the first cell of each row is displayed in this new column.\n\nThe user can add or edit annotations by clicking the pencil icon in the "Annotations" column.

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -748,6 +748,9 @@ class MisReportInstance(models.Model):
             domain.extend(ast.literal_eval(self.analytic_domain))
         # contextual analytic domain filter
         domain.extend(self.env.context.get("mis_analytic_domain", []))
+        # company filter
+        if self.query_company_ids:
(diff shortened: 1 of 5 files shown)


sbidoul:
Thanks for this contrib. I'm afraid we had forgotten to close that issue because it has been implemented earlier this year in a more extensive way with cell-level annotations.

sbidoul:
For 18.0 it's in https://github.com/OCA/mis-builder/pull/689

sbidoul:
Hi, have you seen my comment above https://github.com/OCA/mis-builder/pull/749#issuecomment-3477508531? 

Why do you need this since we now have cell-level annotations?

fsmw:
Closing per @sbidoul feedback: cell-level annotations were already implemented in #689. This approach is superseded.

PR #751, opened 2025-11-19
[18.0][FIX] mis_builder: adapt account codes for reports

ThiagoMForgeFlow:
In version 18, Odoo displays the account codes using the `placeholder_code` field when it’s a company that is not the main active one. With this change, the code of each account is shown correctly in multi-company reports.

Screenshot of how it works without this new code: 

@ForgeFlow


ThiagoMForgeFlow:
Hi @sbidoul What do you think of this proposal?

DavidJForgeFlow:
@sbidoul What do you think about this?

sbidoul:
Hi, this should be resolved in #790. Thanks for your patience.

PR #756, opened 2025-11-26
[18.0] [FIX] mis_builder: account.account multi company support for Odoo 18

Shide:
Adapted from https://github.com/OCA/mis-builder/pull/711 and https://github.com/OCA/mis-builder/pull/730.

Limit is set con code and now query_companies it's used to show properly the companies that are being queried. 
Also fixed the account code that shows to False if the env.company is not the fetched company data.

Tests are mainly from @StefanRijnhart, but adapted a little bit.

Coauthored with @StefanRijnhart and @geraldo29

image

MT-11000 @moduon @rafaelbn @StefanRijnhart @geraldo29 @sbidoul @ThiagoMForgeFlow please review if you want 😄

### DIFF
--- a/mis_builder/i18n/es.po
+++ b/mis_builder/i18n/es.po
@@ -1,23 +1,21 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
-# * mis_builder
+# 	* mis_builder
 #
-# Translators:
-# OCA Transbot <transbot@odoo-community.org>, 2018
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 11.0\n"
+"Project-Id-Version: Odoo Server 18.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-01-13 15:37+0000\n"
-"PO-Revision-Date: 2023-11-03 21:38+0000\n"
-"Last-Translator: Ivorra78 <informatica@totmaterial.es>\n"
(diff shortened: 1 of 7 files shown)


yajo (approved):
Code review.

Shide:
Just added translations to "and <nbr_companies_exceeded> more" and translations for other missing terms in es_ES.

StefanRijnhart (approved):
Thanks!

sbidoul (changes requested):
Thanks!

The PR is fixes the issue but introduces a change, in the sense that before, there was one detail line per account and company.

I think I more or less know what to do with a first baby step towards configurable account detail expansion.

Marking as request change for now as I'm not convinced yet this is a good default behaviour.

I hope to come to this soon.

yajo:
IMHO the fix should be merged and, later, add that configurability. Otherwise the module explodes in users' hands right now.

rafaelbn:
@Shide could we apply this PR in PROD? The issue is affecting production clients enviroments. Thank you!

Shide:
@rafaelbn, unless @sbidoul are making another PR that makes this case better structured, I think we could.
When the final PR comes, we could remove this and adopt the standard way. This PR doesn't add new fields nor any fields that we need to revert in the future.

@sbidoul, could you ping us to know when the new PR will replace this one? Thanks

sbidoul:
@Shide @rafaelbn can you check #790?

PR #758, opened 2026-01-21
[18.0][FIX] mis_builder: Fix date rage multicompanies support

victoralmau:
Fix date rage multicompanies support

Related to https://github.com/OCA/server-ux/commit/f63cdf81c83ab482b7d70ce2f48ef19b0bba30ec

Please @pilarvargas-tecnativa and @christian-ramos-tecnativa can you review it?

@Tecnativa

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -144,9 +144,9 @@ class MisReportInstancePeriod(models.Model):
                         ("date_start", "<=", d),
                         ("date_end", ">=", d),
                         "|",
-                        ("company_id", "=", False),
+                        ("company_ids", "=", False),
                         (
-                            "company_id",
+                            "company_ids",
                             "in",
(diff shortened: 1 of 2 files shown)


sbidoul:
That's a big backward incompatible change in date range. Is there no way to preserve compatibility there?

pilarvargas-tecnativa (approved):

victoralmau:
> That's a big backward incompatible change in date range. Is there no way to preserve compatibility there?

There is currently no compatibility in date ranges with the `company_id` field, which is why this change is being done.

Similar to https://github.com/OCA/account-financial-reporting/pull/1437#pullrequestreview-3683171064

pedrobaeza:
Whatever the solution, it should be applied quickly. It can be this one, or have a computed searchable company_id field back in date_range.

etobella:
The change has been reverted, I think we can close it.

PR #759, opened 2026-02-20
[18.0][IMP] mis_builder: Change where the mock has to be loaded

victoralmau:
With the changes introduced in https://github.com/odoo/odoo/commit/de056cc784a3bbe2575fd3c9e81ca62e73c362d4#diff-b37c7fd5520c97a29ddc59495779e7be61a66e15e85603928e27b3affb9dc31f, an error was occurring because the change validation on the model was being performed before tearDownClass was executed. Therefore, it has been modified so that the mock changes are applied in each test and cleaned up after each test. This way, the error will no longer appear and the test will run normally.

@Tecnativa

### DIFF
--- a/mis_builder/tests/test_kpi_data.py
+++ b/mis_builder/tests/test_kpi_data.py
@@ -9,19 +9,18 @@ from ..models.mis_kpi_data import ACC_AVG, ACC_SUM
 
 
 class TestKpiData(TransactionCase):
-    @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
+    def setUp(self):
+        super().setUp()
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Let's fix it for having green CIs.

/ocabot merge patch

sbidoul:
Pedro, no so fast, especially on modules with declared maintainers.

pedrobaeza:
Do you see any problem in the patch?

This has been incubated during days in several places between a lot of people including @etobella @StefanRijnhart, @victoralmau and me, and it's breaking all the CIs, so it's fully justified.

I usually don't want to have nothing to do with this repo, but this is something critical.

PR #770, opened 2026-03-05
[16.0][FIX] mis_builder: Change migrate signature

Shide:
Use correct word in signature. 

- Enterprise migrations fail with previous signature.
- Does not affect to OpenUpgrade migrations.

MT-13062 @moduon @sbidoul @yajo please review if you want 😄


sbidoul:
Thanks!

/ocabot merge patch

PR #777, opened 2026-03-22
[18.0][FIX] mis_builder: guard against falsy drilldown action

### LINKED ISSUE #776: [18.0][BUG] mis_builder: JS error on drilldown for computed KPI lines
## Description

Clicking a computed KPI cell (e.g. `profit - loss`) in the MIS report widget triggers:

```
UncaughtPromiseError: No view found for act_window action undefined
```

## Root cause

`mis_report_instance.drilldown()` returns `False` when the expression has no account variable (`AEP.has_account_var()` is False for computed expressions like `profit - loss`). The JS handler calls `this.action.doAction(action)` without checking for a falsy return:

```js
// mis_report_widget.esm.js line ~137
async drilldown(event) {
    const drilldown = JSON.parse(event.target.dataset.drilldown);
    const action = await this.orm.call(
        "mis.report.instance", "drilldown",
        [this._instanceId(), drilldown],
        {context: this.context}
    );
    this.action.doAction(action);  // ← action is False
}
```

## Expected behavior

Clicking a computed KPI cell should either:
1. Do nothing (silently ignore), or
2. Show a notification that drill-down is not available for computed lines

## Steps to reproduce

1. Create a MIS report with a computed KPI (e.g. `profit - loss`)
2. Create a report instance with at least one period
3. Click on the computed KPI cell value
4. Observe the JS error

## Environment

- Odoo 18.0 CE
- mis_builder 18.0.1.8.1

dnplkndll:
## Description

Clicking a computed KPI cell (e.g. `profit - loss`) in the MIS report
widget triggers a JS error:

```
No view found for act_window action undefined
```

## Root cause

`mis_report_instance.drilldown()` returns `False` when the expression
has no account variable (`AEP.has_account_var()` is `False` for computed
expressions). The JS handler passes this falsy value directly to
`this.action.doAction()`.

## Fix

Add a truthy check before calling `doAction()`.

Closes #776

### DIFF
--- a/mis_builder/static/src/components/mis_report_widget.esm.js
+++ b/mis_builder/static/src/components/mis_report_widget.esm.js
@@ -137,7 +137,9 @@ export class MisReportWidget extends Component {
             [this._instanceId(), drilldown],
             {context: this.context}
         );
-        this.action.doAction(action);
+        if (action) {
+            this.action.doAction(action);
+        }
     }
 
     async refresh() {


sbidoul:
Hi, this change looks good but I really wonder why such a cell would be clickable in the first place.

For instance with the demo report you can see on runboat, the Total row is not clickable.

dnplkndll:
Closing this — the root cause turned out to be a migration issue, not a missing JS guard.

After migrating from 17.0 to 18.0, `source_aml_model_id` (a stored computed field introduced in 18.0) is NULL on all existing `mis.report.instance.period` records. This causes `drilldown()` to return an action with `res_model: False`, which the JS then fails on with "No view found for act_window action undefined".

On a clean 18.0 install, `source_aml_model_id` is computed correctly and drilldown works fine — no JS guard needed.

The proper fix is a migration script to recompute the field: #780 / ledoent/mis-builder#3

PR #781, opened 2026-03-24
[18.0][IMP] - css sticky table header added on mis-builder reports

hitrosol:


dora-jurcevic:
Hello @hitrosol ! 
 I tested on firefox and seems to be working well with the runbot demo data. 
Just one comment,  maybe it would be better to use the bootstrap variables for colors , like this : 
background-color: var(--gray-300);

hitrosol:
Thank you for your review and suggestions, @dora-jurcevic. I’ve implemented the changes.
I’m also planning in the next PR to freeze the left column in reports with many columns to help users maintain context.

PR #785, opened 2026-04-03
[18.0] mis_builder: add checks for P&L and balance sheet

alexis-via:
This PR adds an optional selection field on mis.report with 2 possible values:
1) P&L
2) Balance Sheet

If this optional selection field is set to P&L, it will check that only balp is used and that each income & expense account is used exactly once and that no other account is used
If the optional selection field is set to Balance sheet, it will check  that only bale/pbale/nbale are used, that each account (unless off_balance and equity_unaffected) is used once for bale, or once in pbale AND once is nbale.

Questions on this PR:

Where should we plug the constraint checks on mis.report.instance ?
At the moment, it is called when you click on the buttons "Preview" or "Print" or "Export". But it is not called when the report is in the dashboard. Which method should we inherit to perform checks in all circumstances ?

In v18, the display_name of an account depends on the company of the context. I use the display_name of accounts in error messages, but a MIS report template can apply to several companies. At the moment, I use the display_name of the first company of the MIS report template.

We could adds checks for the signs in the formula (cf TODO in comments).

### DIFF
--- a/mis_builder/models/aep.py
+++ b/mis_builder/models/aep.py
@@ -314,17 +314,23 @@ class AccountingExpressionProcessor:
         return account_ids
 
     def get_accounting_variables_for_expr(self, expr):
-        """Iterate accounting variables in an expression.
+        """Return the details of an expression. Used for consistency checks.
 
         Prerequisite: done_parsing() must have been invoked.
 
-        Returns a set of (field, mode) used in the expression.
+        Returns a list of (field, mode, account_ids, expression_item)
+        used in the expression. expression_item is useful to have
+        accurate error messages.
         """
-        vars = set()
+        res = []
         for mo in self._ACC_RE.finditer(expr):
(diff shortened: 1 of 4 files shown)


alexis-via:
In some countries (BE for example), there are some informative lines ("Of which ...") in P&L or balance sheet that need to be excluded from the constraint. We should add a boolean field such as "constraint_exclude" on mis.report.kpi that would be enabled for the lines "Of which ..." in order to exclude them from the constraint.

Example:
https://github.com/OCA/l10n-belgium/blob/18.0/l10n_be_mis_reports/data/mis_report_pl_m01-f.xml#L231

alexis-via:
@sbidoul as discussed, I added the boolean option "ignore_constraint" on KPIs.

sbidoul:
@alexis-via  Would it make sense to accept and ignore `balu` and the special account for unallocated result for Balance Sheet constraints?

PR #782, opened 2026-03-24
[18.0][FIX] mis_builder: fix multi-company account display

dnplkndll:
## Summary

Fixes multi-company MIS report issues in Odoo 18:

1. **`company_id` → `company_ids`**: `account.account.company_id` was replaced by `company_ids` (Many2many) in Odoo 18. `_get_account_name` crashed on multi-company reports.

2. **False account codes**: `account.code` is now company-dependent (stored in `code_store` jsonb). When viewing a report from a different company context, codes displayed as "False". Fix uses `with_company()` to read codes from the account's owning company.

3. **`query_companies` threading**: Passes company context through `KpiMatrix` and `prepare_kpi_matrix()` for proper multi-company resolution.

### Multi-Company Report — Account Codes with Company Labels

Report pulls data from **both companies** — account codes show correctly with company name in brackets. Total of **$37,000** consolidates Branch Office ($22,000) + YourCompany ($15,000):

Multi-Company Report

### Report Configuration

**Comparison Mode** enabled, **Year to Date** period with actuals from Journal Items:

Config

### Single Company Reference (Demo Data)

Standard single-company view for reference — Demo Expenses report with budget columns:

Single Company

## Files changed

- `mis_builder/models/kpimatrix.py` — accept `query_companies` in KpiMatrix
- `mis_builder/models/mis_report.py` — thread `query_companies` through `prepare_kpi_matrix()`
- `mis_builder/models/mis_report_instance.py` — use `with_company()` for account name resolution
- `mis_builder/tests/test_mis_report_instance.py` — multi-company test cases

## Notes

- Migration fix separated into its own PR as requested: ledoent/mis-builder#3
- Pre-commit: all checks pass
- Tests: pass on both Odoo and OCB

Closes #779

PR #788, opened 2026-05-13
[18.0][ADD] mis_builder: add multi-company demo data

dnplkndll:
## Summary

Adds demo data exercising multi-company MIS reports:

- A second company with expense accounts and journal entries
- A `mis.report.instance` configured with `multi_company=True` and `query_company_ids` set across both companies
- A `_post_demo_moves()` helper on `mis.report.instance` that posts the seeded draft journal entries so the demo report renders non-zero values

Only loaded when demo data is enabled (`--without-demo=False`).

## Why split from #782?

Per @sbidoul's review on #782, splitting the demo data out so the runtime fix can move independently. The fix in #782 stands on its own; this PR adds the fixture that makes the multi-company rendering visible/testable from the demo install.

## Test plan

- [ ] `-i mis_builder --without-demo=False`
- [ ] Open "Multi-Company Demo Report" — confirm both companies appear with non-zero totals and account codes render correctly
- [ ] Without `#782` applied: rendering should expose the `account.company_id` AttributeError this demo is designed to surface

Companion of: OCA/mis-builder#782

### DIFF
--- a/mis_builder/__manifest__.py
+++ b/mis_builder/__manifest__.py
@@ -28,6 +28,9 @@
         "report/mis_report_instance_qweb.xml",
         "report/mis_report_instance_xlsx.xml",
     ],
+    "demo": [
+        "demo/mis_multicompany_demo.xml",
+    ],
     "assets": {
         "web.assets_backend": [
             "mis_builder/static/src/components/mis_report_widget.esm.js",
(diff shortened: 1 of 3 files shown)


sbidoul:
Thanks. I'll review this topic and work on the v19 migration by end of May, likely during the sprint in Spain .

PR #775, opened 2026-03-12
[18.0][mis_builder] FIX multi companies issue

ecino:
Fixes #774

PR #790, opened 2026-05-26
[FIX] mis_builder: fix multi-company account display

### LINKED ISSUE #774: [18.0][mis_builder] AttributeError: account.account has no attribute company_id when computing multi-company report
## Module
mis_builder

## Describe the bug
When a MIS report instance is configured with multiple companies (multi_company=True) and "Expand accounts" is enabled on a KPI, computing the report raises an AttributeError because kpimatrix.py accesses account.company_id on account.account records, but in Odoo 18 that field was replaced by company_ids (Many2many).

## To Reproduce
**Affected versions**: 18.0.1.8.1, Odoo 18.0

1. Set up a multi-company environment (at least 2 companies with their own chart of accounts).
2. Create a MIS report with at least one KPI that has "Auto-expand accounts" enabled.
3. Create a MIS report instance that queries multiple companies (i.e. multi_company=True is set internally via query_company_ids).
4. Open or compute the report instance.

**Expected behavior**
The report renders correctly, showing account names with the company name appended in brackets, e.g. 101000 Cash [Company A].

**Additional context**
In Odoo 18, account.account.company_id (Many2one) was removed and replaced by account.account.company_ids (Many2many) to support accounts shared across multiple companies (see odoo/addons/account/models/account_account.py).

### LINKED ISSUE #779: [18.0][BUG] Account codes show as 'False' in auto_expand_accounts for multi-company reports
## Problem

When viewing a MIS report instance that belongs to a different company than the user's current company, account codes display as `False` in auto-expanded account rows. For example, `15050 Loan Fees` shows as `False Loan Fees`.

## Root cause

In `kpimatrix.py`, `_load_account_names()` reads `account.code` without the report instance's company context:

```python
def _load_account_names(self):
    account_ids = set()
    for detail_rows in self._detail_rows.values():
        account_ids.update(detail_rows.keys())
    accounts = self._account_model.search([("id", "in", list(account_ids))])
    self._account_names = {a.id: self._get_account_name(a) for a in accounts}

def _get_account_name(self, account):
    result = f"{account.code} {account.name}"
    ...
```

In Odoo 18, `account.account.code` is a company-dependent field backed by `code_store` (jsonb). When the `env` doesn't have the correct company context, `account.code` returns `False` for accounts belonging to other companies.

`KpiMatrix` is instantiated at `mis_report.py:544` with `self.env` which uses the current user's company, not the report instance's company:

```python
kpi_matrix = KpiMatrix(self.env, multi_company, self.account_model)
```

## Steps to reproduce

1. Create companies A and B, each with their own chart of accounts (with codes)
2. Create a MIS report instance for company B with `auto_expand_accounts` enabled
3. Log in as a user whose current company is A
4. View the MIS report instance for company B → account codes show as `False`

## Expected behavior

Account codes should display correctly regardless of the user's current company.

## Environment

- Odoo 18.0 CE
- mis_builder 18.0
- Multi-company setup with separate charts of accounts

sbidoul:
fixes #779 
fixes #774 

supersedes and closes #751
supersedes and closes #756
supersedes and closes #775
supersedes and closes #782
supersedes and closes #789

### DIFF
--- a/mis_builder/models/kpimatrix.py
+++ b/mis_builder/models/kpimatrix.py
@@ -139,12 +139,18 @@ class KpiMatrixCell:  # noqa: B903 (immutable data class)
 
 
 class KpiMatrix:
-    def __init__(self, env, multi_company=False, account_model="account.account"):
+    def __init__(
+        self,
+        env,
+        companies=None,
+        account_model="account.account",
+    ):
         # cache language id for faster rendering
         lang_model = env["res.lang"]
         self.lang = lang_model._lang_get(env.user.lang)
         self._style_model = env["mis.report.style"]
         self._account_model = env[account_model]
+        self._companies = companies
         # data structures
         # { kpi: KpiMatrixRow }
         self._kpi_rows = OrderedDict()
@@ -158,7 +164,6 @@ class KpiMatrix:
         self._sum_todo = {}
         # { account_id: account_name }
         self._account_names = {}
-        self._multi_company = multi_company
 
     def declare_kpi(self, kpi):
         """Declare a new kpi (row) in the matrix.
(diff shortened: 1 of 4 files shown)


ecino (approved):
👍 LGTM !

PR #793, opened 2026-05-27
[19.0] mig mis_builder_budget & mis_builder_demo

sbidoul:


AnizR (approved):
Code LGTM, thanks!

A refactor that could be great: use `Domain` objects instead of sticking with lists when playing with domains/filters. I think that it would need a lot of changes therefore, it could be done later in another PR.

sbidoul:
> A refactor that could be great: use Domain objects instead of sticking with lists when playing with domains/filters. I think that it would need a lot of changes therefore, it could be done later in another PR.

Yes I would like to do this. We need to keep backward compatibility on methods that are meant to be overridden so we must accept both a list and a Domain. There is `DomainType` expressing that.

mis_builder_budget/models/mis_report_instance.py
```diff
@@ -30,9 +30,9 @@ def _get_kpi_for_expressions(self, expressions):
             if kpi is None:
                 kpi = expression.kpi_id
             else:
-                assert (
-                    kpi == expression.kpi_id
-                ), "expressions must belong to the same kpi"
+                assert kpi == expression.kpi_id, (
+                    "expressions must belong to the same kpi"
+                )
```
petrus-v, line 35: Why not raise an exception here ? assert may not by run using python -O

non blocking point

petrus-v (approved):

mis_builder_budget/models/mis_report_instance.py
sbidoul: I *think* the assert is here to document what the `else` means, but this constraint is actually validated elswhere.

PR #794, opened 2026-05-28
[19.0][PORT] 790 from 18.0

sbidoul:
Port #790 from 18.0 to 19.0.


sbidoul:
Trivial forward port, tested it works the same as in 18

/ocabot merge patch

PR #806, opened 2026-06-12
[19.0][IMP] mis_builder: mis report sticky header 

petrus-v:
## Description

Sticky header (search view + pivot date + table header)

image

### DIFF
--- a/mis_builder/static/src/components/mis_report_widget.scss
+++ b/mis_builder/static/src/components/mis_report_widget.scss
@@ -12,6 +12,9 @@
     }
 
     .mis_builder {
+        border-collapse: separate !important;
+        border-spacing: 0 !important;
+
         a {
             // we don't want the link color, to respect user styles
             color: inherit;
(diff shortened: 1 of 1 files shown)


AnizR (approved):
Code LGTM + tested on runboat: works correctly

mis_builder/static/src/components/mis_report_widget.scss
```diff
@@ -21,6 +24,23 @@
...
         }
+
+        // Make the header cells sticky and ensure they render on top of scrolling content
+        thead {
+            top: -10px;
+            position: sticky;
+            z-index: var(--sticky-header-zindex) !important;
+
+            tr {
+                th {
```
petrus-v, line 35: It seems the first column header is transparent and let display row labels in some case....

Image

petrus-v: this was a side effect of the stacking context create with sticky position + z-index, I overwrite it with higher value to work in different stack z-index context....

petrus-v:
I just rebase and  add the search view and pivot date in the sticky header

samirGuesmi (approved):
LGTM

PR #817, opened 2026-07-20
[19.0][IMP] mis_builder: generic KPI detail expansion by field

max3903:
## Summary
- Add KPI `detail_by` selection (`none` / `account` / `partner`) so report rows can expand one line per partner, not only per account.
- Keep `auto_expand_accounts` as a stored compute/inverse for backward compatibility, with a post-migration mapping existing flags to `detail_by=account`.
- Extend AEP/evaluator/matrix/drilldown for partner-level queries (sentinel `0` = empty partner).

## Test plan
- [ ] Install/upgrade `mis_builder` to `19.0.1.1.0` and confirm KPIs with `auto_expand_accounts` still expand by account
- [ ] Create a KPI with `detail_by=partner` on receivable/payable expressions and verify one row per partner (+ “(No partner)” when empty)
- [ ] Drilldown from a partner detail cell and confirm domain includes `partner_id`
- [ ] Run `mis_builder.tests.test_partner_detail` and existing auto-expand tests (`test_data_sources`)

Made with [Cursor](https://cursor.com)


max3903:
@OCA/mis-builder-maintainers ping

sbidoul:
@max3903 From a first quick look, it does not seem likely that we can merge this. This seems too specific to grouping by partner and when we'll solve this it must be with a mechanism that can group by any field. If not with a configurable UI, at least with a correct abstraction in the code.

max3903:
@sbidoul thanks for the feedback. I reworked the PR around a generic abstraction:

- KPI field `detail_groupby`: technical name of any field on the move line source (`account_id`, `partner_id`, `journal_id`, …)
- `account_id` still uses the optimized AEP path
- other fields share `do_queries_by_groupby` / `replace_exprs_by_groupby` / drilldown via `detail_groupby` + `detail_id`
- `auto_expand_accounts` remains as compatibility alias for `detail_groupby=account_id`

No partner-specific branching left in the public API. Happy to adjust naming/UI further if you prefer a Many2one on `ir.model.fields` instead of a Char.

hitrosol:
I tested this on Runboat, and it now supports the general feature, which means we can expand almost everything from account.move.line.

I was hoping we could also expand the Analytic Accounts section, perhaps by using account.analytic.line as the source for the move lines.

max3903:
@hitrosol thanks for testing on Runboat.

Yes — with `detail_groupby` you can expand on any “simple” field of the configured move-line source (`partner_id`, `journal_id`, `account_id`, etc.).

Analytic accounts are a different case on recent Odoo: analytics live mainly in `analytic_distribution` on `account.move.line`, which is not a plain Many2one, so it cannot be used as `detail_groupby` as-is.

Using `account.analytic.line` as an alternative move-line source is possible in principle (same idea as other AML-like sources: needs `debit`/`credit`/`account_id`/`date`/`company_id`, or a SQL view that exposes them), and then you could set `detail_groupby` to the analytic account field. That would be a separate addon/customization though, not something this PR adds out of the box.

Happy to look at a follow-up if you have a concrete analytic source model/view in mind.

hitrosol:
You can see the module for version 17.0, https://github.com/OCA/mis-builder-contrib/tree/17.0/mis_builder_analytic

max3903:
@hitrosol thanks — that module is exactly the right approach.

[`mis_builder_analytic`](https://github.com/OCA/mis-builder-contrib/tree/17.0/mis_builder_analytic) already exposes an AML-like SQL view (`mis.account.analytic.line`) with `debit` / `credit` / `account_id` / `date` / `company_id`, plus `analytic_account_id`.

With this PR, the intended setup would be:

1. Period source: **Actuals (alternative)** → `MIS Account Analytic Line`
2. KPI **Expand details by**: `analytic_account_id`

So analytic expansion should work on top of that contrib module, without baking analytics into `mis_builder` core.

I don’t see `mis_builder_analytic` on 19.0 yet (only 17.0 in that repo). Migrating it to 19.0 would be a good follow-up in [mis-builder-contrib](https://github.com/OCA/mis-builder-contrib); happy to help with that separately once this PR is in.

sbidoul:
I think I'll close this PR on basis of our AI Policy, irrespective of it's quality (which I don't know).

Only 12 minutes elapsed between my last comment and the updated PR, which means you don't materially have had the time to review and test the (big) change. If I want to talk to a LLM I can do it myself.

Sorry for the direct tone. LLMs increases the asymmetry between code production and code review by orders of magnitude. As a maintainer I feel aggressed when I see this.

max3903:
@sbidoul No need to be sorry. Point taken.

rvalyi:
> I think I'll close this PR on basis of our AI Policy, irrespective of it's quality (which I don't know).
> 
> Only 12 minutes elapsed between my last comment and the updated PR, which means you don't materially have had the time to review and test the (big) change. If I want to talk to a LLM I can do it myself.
> 
> Sorry for the direct tone. LLMs increases the asymmetry between code production and code review by orders of magnitude. As a maintainer I feel aggressed when I see this.
> 
> 

hello @sbidoul, this is a huge problem and it is probably harming good ideas too among the AI slope... 

Catching your attention here to tell you we are having the same issues with OCA/l10n-brazil with usual fake troll (mi..o) who didn't help shit over the past 4 years (like hard account v13 or v16 refactors and who were promoting absolutely deadly stuff before) now suddenly opening 30 PRs per day and coordinating pressure to get us merge any irrelevant thing...

I know we have the AI policy, but being co-author of the largest OCA module (l10n_br_fiscal) and largest OCA repo (l10n-brazil), I feel this is something we need to address further because if the trolls (not including you @max3903 cause I know you are generally a dedicated contributor) make the real maintainers run away, then OCA is screwed because the AI is not going to hold all the bits together in the next 2-3 years at least. That is really the biggest threat for the OCA right now...

cc @renatonlima, @sebastienbeau @alexis-via @marcelsavegnago, @antoniospneto, @pedrobaeza @etobella @dreispt

max3903:
@sbidoul We have made some tests and examples:
image

Please consider reviewing this pull request.

jelenapoblet:
I understand the concern raised regarding AI. Producing code with AI and then asking a human maintainer to review it is probably not a good practice.

I'm shure @max3903 will take this into account for future contributions. 

That said, some time has passed since the initial discussion, and IMO we cloud give the current PR another opportunity to go through a review process.

We use MIS Builder ourselves for our reporting, so from our side the concern is backward compatibility. We want to make sure  that when updating the module, our existing reports continue to work as before.

We are planning to test this with some of our existing reports in the next few days and can provide feedback here.

If the existing behavior is preserved and the migration works correctly, I think we can leave the initial LLM issue behind and evaluate the PR based on its current implementation, tests and impact on existing users.

I also think the original concern around the partner approach has been addressed with the move to the generic detail_groupby solution, so I would suggest giving the PR another review when possible.

PR #819, opened 2026-07-20
[IMP] remove redundant attributes string

sbidoul:


sbidoul:
Fixing linter issues revealed in #812

PR #820, opened 2026-07-20
[15.0] Dotfiles update

sbidoul:


sbidoul:
Done manually to unblock renovate #818

PR #821, opened 2026-07-24
[15.0][ADD] mis_report_account_coverage_check: New module

carlosdauden:
Wizard that, given a MIS report instance and an account-code range, detects accounts posted to during the report's period(s) that are not referenced by any KPI expression of the report -- e.g. to find why a balance sheet built with mis_builder doesn't balance.

@Tecnativa

ping @sergio-teruel @carlos-lopez-tecnativa @victoralmau


mis_report_account_coverage_check/tests/test_mis_report_account_coverage_check.py
```diff
@@ -0,0 +1,162 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import datetime
+
+from odoo import fields
+from odoo.tests import common
+
+
+class TestMisReportAccountCoverageCheck(common.TransactionCase):
+    def setUp(self):
+        super().setUp()
```
carlos-lopez-tecnativa: Why not in the setUpClass method?

carlos-lopez-tecnativa (commented):
Interesting feature. Please check the CI.

mis_report_account_coverage_check/tests/test_mis_report_account_coverage_check.py
carlosdauden: Changes done.

The reason was to ensure consistency, given that even in version 19.0 ‘setUp’ is used:
https://github.com/OCA/mis-builder/blob/19.0/mis_builder/tests/test_aep.py#L23-L24
but if it doesn't bother anyone, I'm replacing it with our test creation method, which yields better results:

sbidoul:
Can you move this PR to oca/mis-builder-contrib?

You may also want to have a look at https://github.com/OCA/mis-builder/pull/785 which is almost ready to merge, and seem to address a similar use case.

carlosdauden:
https://github.com/OCA/mis-builder-contrib/pull/67

Issue #619, opened 2024-07-31
Support the Accounting Read-only group


sbidoul:
Provide read-only access to MIS Report instances to members of `account.group_account_readonly`.

PR #822, opened 2026-07-29
[18.0][FIX] mis_builder: extend the MIS Builder menu in accounting to the Accounting Read Only group

victoralmau:
Define the appropriate group in the menu

A user with the "Show Accounting Features - Read-only" group (`account.group_account_readonly`) should be able to view the menu

@Tecnativa TT63708

### DIFF
--- a/mis_builder/views/mis_report_instance.xml
+++ b/mis_builder/views/mis_report_instance.xml
@@ -225,7 +225,7 @@
         parent="account.menu_finance_reports"
         name="MIS Reporting"
         sequence="101"
-        groups="account.group_account_manager"
+        groups="account.group_account_readonly"
     />
     <menuitem
         id="mis_report_instance_view_menu"


pedrobaeza (approved):

sbidoul:
We have an open issue to support that group: https://github.com/OCA/mis-builder/issues/619

But why changing it for the menu only when the child menu entries will not work?

pedrobaeza:
@victoralmau please check that, and you should also check `ir.model.access`

victoralmau:
> But why changing it for the menu only when the child menu entries will not work?

The menu options do work because, at https://github.com/OCA/mis-builder/blob/18.0/mis_builder/security/ir.model.access.csv, `base.group_user` has read access.

sbidoul:
Ah yes. So that's all we need to do to close #619 ?

pedrobaeza:
@sbidoul indeed it's working:

imagen

and able to create MIS reports (and obviously read):

imagen

pedrobaeza:
> Ah yes. So that's all we need to do to close #619 ?

I would say yes according the testing I have just done.

sbidoul (approved):
Thanks!

pedrobaeza:
@sbidoul should I trigger the merge command?

sbidoul:
Thanks!

PR #826, opened 2026-08-20
[19.0][IMP] mis_builder: Preserve filters and pivot date on report drilldown

petrus-v:
When navigating away from a MIS report via drilldown links and returning using the breadcrumb, active search filters and pivot date were lost.

This commit preserves the report widget state across view navigation:
- Register `useSetupAction` in `MisReportWidget` to export search model state and pivot date when leaving the controller.
- Forward `props.state` to `env.config.state` in `FormController` so that restored controller state is accessible during component `willStart`.
- Hydrate `searchModel` and `pivot_date` upon view restoration.
- Add JS tour `mis_report_drilldown_tour` and HttpCase test `TestMisReportWidgetTour` to validate filter preservation end-to-end.

### DIFF
--- a/mis_builder/static/src/components/mis_report_widget.esm.js
+++ b/mis_builder/static/src/components/mis_report_widget.esm.js
@@ -1,36 +1,64 @@
 import {Component, onMounted, onWillStart, useState, useSubEnv} from "@odoo/owl";
+import {parseDate, serializeDate} from "@web/core/l10n/dates";
 import {useBus, useService} from "@web/core/utils/hooks";
+import {AnnotationDialog} from "../annotation_dialog/annotation_dialog.esm";
 import {DateTimeInput} from "@web/core/datetime/datetime_input";
+import {FormController} from "@web/views/form/form_controller";
 import {SearchBar} from "@web/search/search_bar/search_bar";
 import {SearchModel} from "@web/search/search_model";
-import {parseDate} from "@web/core/l10n/dates";
-import {registry} from "@web/core/registry";
-import {AnnotationDialog} from "../annotation_dialog/annotation_dialog.esm";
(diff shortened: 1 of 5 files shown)


mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -1,44 +1,80 @@
...
+// Patch FormController to forward restored controller state to env.config.
+// When returning from drilldown via breadcrumbs, ActionService restores FormController
+// and passes saved exportedState via props.state. FormController does not propagate
+// props.state to env.config.state by default, so patching it here ensures child components
+// (like MisReportWidget) can access searchState in willStart() during view restoration.
+patch(FormController.prototype, {
+    setup() {
+        super.setup();
+        if (this.props.state) {
+            this.env.config.state = this.props.state;
```
AnizR: Does this will have impact on other model's forms?

AnizR (approved):
Code LGTM
Tested on runboat, seems to work fine.

I realised that using browser's back arrow doesn't work when being on mis report instances but this is already the case on 19.0's branch.

mis_builder/static/src/components/mis_report_widget.esm.js
petrus-v: I miss your comment yesterday, good point, you are right this will apply on any forms 🤔... we should restricted it to mis.report.instance or extend the FormController and use the new form view. I'd like to try the second approach

petrus-v: I've give a try by extending FormController but it seems odoo do not take care of js_class while coming back from drilldown instanciate a FormController... so I rollback for a simplest version

petrus-v:
I'm closing this PR as changed is included in #827

PR #827, opened 2026-08-20
[19.0][IMP] mis_builder: Improve drilldown navigation with cache report support

petrus-v:
When navigating away from a MIS report via drilldown links and returning using the breadcrumb, active search filters and pivot date were lost.

- First commit preserves the report widget state across view navigation (only using breadcrumb)
- second commit Allow caching computed report data in browser memory during drilldown
navigation to prevent recomputing the report when returning via breadcrumbs (this feature is opt out by default for consistency with current version).
- third commit fix navigation while using browser navigation back button (before this commit we get the config view instead of the preview form)

https://github.com/user-attachments/assets/d2ba095d-5434-4902-8b62-28377b19d128

* commit 8882b31 is to fix an issue when search bar widget is not displayed (internal issue ref #98869)

### DIFF
--- a/mis_builder/static/src/components/mis_report_widget.esm.js
+++ b/mis_builder/static/src/components/mis_report_widget.esm.js
@@ -86,10 +86,14 @@ export class MisReportWidget extends Component {
 
     exportState() {
         return {
-            searchState: this.searchModel ? this.searchModel.exportState() : null,
-            pivotDate: this.state.pivot_date
-                ? serializeDate(this.state.pivot_date)
-                : null,
+            searchState:
+                this.showSearchBar && this.searchModel && this.searchModel.sections
+                    ? this.searchModel.exportState()
+                    : null,
+            pivotDate:
+                this.showPivotDate && this.state.pivot_date
+                    ? serializeDate(this.state.pivot_date)
+                    : null,
             cachedReportData:
                 this.widget_cache_report_on_drill_down && this.state.mis_report_data
                     ? this.state.mis_report_data
--- a/mis_builder/static/src/components/mis_report_widget.esm.js
+++ b/mis_builder/static/src/components/mis_report_widget.esm.js
@@ -11,17 +11,17 @@ import {patch} from "@web/core/utils/patch";
 import {registry} from "@web/core/registry";
 import {useSetupAction} from "@web/search/action_hook";
 
-// Patch FormController to forward restored controller state to env.config
+// Patch FormController to forward restored globalState to env.config
 // for mis.report.instance form views.
-// When returning from drilldown via breadcrumbs, ActionService restores FormController
-// and passes saved exportedState via props.state. FormController does not propagate
-// props.state to env.config.state by default, so patching it here ensures child components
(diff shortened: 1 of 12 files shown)


petrus-v:
I've force/push a refactor to expose `exportState` and `_restoreState` that a future module could improve the search view (such as the pivot date) and save/restore its state easily !

AnizR (approved):
Code lgtm
Tested on runboat

sbidoul:
Not looked at the code yet but a question: what is the scope of the cache, how is memory managed, is there a risk of memory leak?

petrus-v:
> Not looked at the code yet but a question: what is the scope of the cache, how is memory managed, is there a risk of memory leak?

Cache is scoped to the Form Controller so as soon user will click on a new menu, the allocated RAM should be available to the GC.

petrus-v:
I've close #826 because this PR was based on it. Also I've add the third commit that fix coming back from drill-down using browser navigation buttons

vdewulf:
hello !
@petrus-v I was wondering if the improvement you propose would also solve this case:
- I view the report which is very long. I need to scroll down to check a specific line. 
- I click on the value for the specific line. This opens the drilldown view.
- I want to go back to the report, I click on the breadcrumb
- Result: the report is displayed from the top. I need to scroll again to continue the control on the subsequent lines

When we do control on a lot of clients databases, this takes a lot of time and makes us scroll a lot. Any idea if this fix would help or if this is something different?
Thanks!

petrus-v:
> hello ! @petrus-v I was wondering if the improvement you propose would also solve this case:
> 
>     * I view the report which is very long. I need to scroll down to check a specific line.
> 
>     * I click on the value for the specific line. This opens the drilldown view.
> 
>     * I want to go back to the report, I click on the breadcrumb
> 
>     * Result: the report is displayed from the top. I need to scroll again to continue the control on the subsequent lines
> 
> 
> When we do control on a lot of clients databases, this takes a lot of time and makes us scroll a lot. Any idea if this fix would help or if this is something different? Thanks!

Hi @vdewulf :) !

Thanks for reporting this, I've test on a customer instance where this PR is installed, unfortunately I've reproduce the behavior so at the moment this PR doesn't seems to scroll down to the clicked cell while coming back from a drilldown  ! 

(on that instance we have project module that play with sticky column and other css stuff that I'm not 100% sure if that could be different without those project modules).

Anyway I think this PR is large enough for review and if getting stable, I would suggested to report the issue if not already present or do it in a new PR on top of this one !

Definitively a good idea, I already imagine a css animation to highlight the clicked cell few seconds :star2:

mis_builder/static/src/components/mis_report_widget.esm.js
```diff
@@ -232,6 +325,9 @@ export class MisReportWidget extends Component {
 
     onDateTimeChanged(ev) {
         this.state.pivot_date = ev;
+        if (this.props.record) {
+            this.props.record._misReportPivotDate = serializeDate(ev);
+        }
```
sbidoul: Can we do this in _export? Or comment why this needs to be done here?

PR #829, opened 2026-09-02
[FIX] mis_builder: multi company reports

petrus-v:
This PR fix an issue to preview multi-company report:

To reproduce/test:
- create at least two companies
- setup multi-company mis report that belongs to those two company
- create a user that belongs one of them
- connect with the freshly created user and open the preview of the document

before this commit we get an access error on res.company model because the user is not allow to read the second company 
after this PR user can open the preview without issue

### DIFF
--- a/mis_builder/models/mis_report_instance.py
+++ b/mis_builder/models/mis_report_instance.py
@@ -644,10 +644,8 @@ class MisReportInstance(models.Model):
     def _compute_query_company_ids(self):
         for rec in self:
             if rec.multi_company:
-                if not rec.company_ids:
-                    rec.query_company_ids = self.env.companies
-                else:
-                    rec.query_company_ids = rec.company_ids & self.env.companies
+                company_ids = rec.sudo().company_ids & self.env.companies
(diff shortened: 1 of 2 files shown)


AnizR (approved):
Code LGTM

mis_builder/models/mis_report_instance.py
```diff
@@ -644,10 +644,8 @@ def _onchange_company(self):
     def _compute_query_company_ids(self):
         for rec in self:
             if rec.multi_company:
-                if not rec.company_ids:
-                    rec.query_company_ids = self.env.companies
-                else:
-                    rec.query_company_ids = rec.company_ids & self.env.companies
+                company_ids = rec.sudo().company_ids & self.env.companies
+                rec.query_company_ids = company_ids or self.env.companies
```
sbidoul: @pniederlag I find this version more difficult to read and I'm not entirely sure it does exactly the same as before. Would it make sense to keep the same if/else structure? Or does this address another problem than the missing sudo?

PR #830, opened 2026-09-10
[19.0][FIX] mis_builder: don't report unresolvable styles on intermediate passes

MiquelRForgeFlow:
## Description

`KpiMatrix.set_values_detail_account` is called on every pass of the computation loop, including the passes where a KPI is requeued because its expression referenced a KPI that was not evaluated yet. On those passes `locals_dict` is incomplete, so a `style_expression` referring to the KPI's own value fails and `mis_safe_eval` returns a `DataError`.

`DataError` subclasses `Exception`, so it is truthy and renders as an empty string. It was therefore passed to the style search and logged as `Style '' not found.` once per intermediate pass. That is noise on a perfectly correct report, and it hides the real problem when a style expression is actually broken:

```yml
    # a report whose KPI style expression references its own value, before
    ERROR ... kpimatrix: Style '' not found.
    ERROR ... kpimatrix: Style '' not found.
    ERROR ... kpimatrix: Style '' not found.

    # after
    (nothing)

    # a genuinely broken style expression, before
    ERROR ... kpimatrix: Style '' not found.

    # after
    ERROR ... kpimatrix: Error evaluating style expression
      <'Big' if unknown_kpi > 10 else 'Small'>: NameError: name 'unknown_kpi' is not defined
```

The style evaluation is now skipped when the cell value is itself in error, since that error is already reported in the cell, and a `DataError` coming from the style expression is reported with its actual message. `style_name` is also initialised, as it was left unbound if `mis_safe_eval` ever raised.

No preliminary issue was opened, as the change is small and self-contained.

## Test

`mis_builder/tests/test_style_expression.py` is added in a separate commit (e6dd7818) so the bug can be reproduced before the fix. On that commit the three tests fail:

```yml
    FAIL: TestStyleExpression.test_style_expression_of_requeued_kpi
    AssertionError: Unexpected logs found: ["ERROR:...kpimatrix:Style '' not found."]
    FAIL: TestStyleExpression.test_style_expression_error
    AssertionError: 2 != 1
    FAIL: TestStyleExpression.test_style_expression_not_found
    AssertionError: 2 != 1
```

With the fix applied the full module suite is green (122 tests, 0 failed).

## Target branch

19.0, where the bug is. The "Target branch" section of the PR template is out of date: it still refers to versions 9 to 12 and asks to target 10.0.

## CLA

Signed.

## Changelog entry

`mis_builder/readme/newsfragments/830.bugfix`, included in the fix commit.


mis_builder/models/kpimatrix.py
```diff
@@ -262,7 +267,13 @@ def set_values_detail_account(
                         row.kpi.style_expression,
                         exc_info=True,
                     )
-                if style_name:
+                if isinstance(style_name, DataError):
```
petrus-v: *question*: `mis_safe_eval` can return `NameDataError` instance as well, should we managed it or it don't make sense ?

petrus-v (approved):

mis_builder/models/kpimatrix.py
MiquelRForgeFlow: `NameDataError` is a subclass of `DataError`, so it's already handled by the `isinstance(style_name, DataError) `branch and logged with its message (covered by `test_style_expression_error`, where `unknown_kpi` yields a `NameDataError`).

I don't think it deserves a special treatment: the style is only evaluated when the kpi value is not in error, i.e. when the kpi is not requeued and this is its final evaluation. A `#NAME` at that point won't be resolved by a later pass, so it's a genuine configuration error and should be reported. The expected `#NAME` of intermediate passes is already filtered out by the `not isinstance(val, DataError)` guard.

PR #832, opened 2026-09-21
[19.0][FIX] mis_builder: fetch name of archived accounts

AnizR:
## Description

If an account is archived but should be displayed in a mis report, an error is raised.
The error is due to the fact that fetch of account's name is done without the context key "active_test".


pierre-halleux (approved):

FrancoMaxime (approved):
LGTM: code review

RaphaelVicini:
LGTM: code review

sbidoul:
makes sense. It makes sense to do it there as in general mis_builder must treat archived accounts like any other account, as long as they are used in a report.

sbidoul:
Did this lead to invalid data in reports, or just missing names in the left column?

AnizR:
> Did this lead to invalid data in reports, or just missing names in the left column?

Well, neither, it gave an AccessError

